PR #2, opened 2022-04-01
Add an example policy related to attestation data

simonbaird:
I'm aiming to create a set of useful and tested examples that can
serve as a reference when we do write the real policies.

This example shows how to use some mocked data and write some test
coverage for a more realistic rule.

### DIFF
--- a/policies/examples/attestation_checks.rego
+++ b/policies/examples/attestation_checks.rego
@@ -1,0 +1,21 @@
+package examples.attestation_checks
+
+accepted_attestation_types := ["https://in-toto.io/Statement/v0.1"]
+
+attestation_type_ok(attestation_type) {
+	accepted_attestation_types[_] = attestation_type
+}
+
+deny = {"msg": msg} {
+	not attestation_type_ok(data.attestation._type)
+
+	quoted_list := [quoted_type |
+		t := accepted_attestation_types[_]
+		quoted_type := sprintf("'%s'", [t])
+	]
+
+	msg := sprintf(
+		"Invalid value in attestation _type field: '%s'. Expecting one of the following: %s",
+		[data.attestation._type, concat(", ", quoted_list)],
+	)
+}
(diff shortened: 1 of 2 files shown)


policies/examples/attestation_checks.rego
```diff
@@ -0,0 +1,21 @@
...
+accepted_attestation_types := ["https://in-toto.io/Statement/v0.1"]
+
+attestation_type_ok(attestation_type) {
+	accepted_attestation_types[_] = attestation_type
+}
+
+deny = {"msg": msg} {
+	not attestation_type_ok(data.attestation._type)
+
+	quoted_list := [quoted_type |
```
lcarva, line 12: What a wild syntax!

policies/examples/attestation_checks.rego
```diff
@@ -0,0 +1,21 @@
+package examples.attestation_checks
+
+accepted_attestation_types := ["https://in-toto.io/Statement/v0.1"]
+
+attestation_type_ok(attestation_type) {
+	accepted_attestation_types[_] = attestation_type
+}
+
+deny = {"msg": msg} {
```
lcarva, line 9: I wonder if we should agree on the usage of `allow` vs `deny`. It matters when we query for the value. `allow` has the benefit of being possible to evaluate to a single boolean value, `true`. If it's undefined, or a different type, then it's not allowed. With `deny`, I guess we'd have to be a bit smarter around undefined values?

lcarva (approved):

policies/examples/attestation_checks.rego
simonbaird: Not sure, but I agree we should be thinking about this.

Some considerations:
* I expect you can derive allow from deny without much trouble, e.g. `allow { count(deny) == 0 }`.
* I think our focus is on providing useful reason messages, so whatever convention we come up with should provide a clear way to do that.

PR #1, opened 2022-03-31
First unit test and Github action

simonbaird:

PR #3, opened 2022-04-02
Check the sha when installing opa

simonbaird:
As suggested by @lcarva in #1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,12 +27,17 @@ OPA_VER=v0.39.0
 OPA_FILE=opa_linux_amd64_static
 OPA_URL=https://openpolicyagent.org/downloads/$(OPA_VER)/$(OPA_FILE)
 OPA_SHA=19a24f51d954190c02aafeac5867c9add286c6ab12ea85b3d8d348c98d633319
-OPA_DEST=/usr/bin/opa
+ifndef OPA_BIN
+	OPA_BIN=$(HOME)/bin
+endif
+OPA_DEST=$(OPA_BIN)/opa
 
 install-opa:
 	curl -s -L -O $(OPA_URL)
 	echo "$(OPA_SHA) $(OPA_FILE)" | sha256sum --check
-	sudo cp $(OPA_FILE) $(OPA_DEST)
-	sudo chmod 755 $(OPA_DEST)
+	mkdir -p $(OPA_BIN)
+	cp $(OPA_FILE) $(OPA_DEST)
+	chmod 755 $(OPA_DEST)
+	rm $(OPA_FILE)
 
 .PHONY: help test fmt fmt-check ci install-opa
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -24,11 +24,17 @@ fmt-check:
 ci: fmt-check test
 
 OPA_VER=v0.39.0
-OPA_URL=https://openpolicyagent.org/downloads/$(OPA_VER)/opa_linux_amd64_static
+OPA_FILE=opa_linux_amd64_static
+OPA_URL=https://openpolicyagent.org/downloads/$(OPA_VER)/$(OPA_FILE)
+OPA_SHA_FILE=$(OPA_FILE).sha256
+OPA_SHA_URL=$(OPA_URL).sha256
```
lcarva, line 30: I was thinking more of including the checkusm in git so we protect ourselves against `openpolicyagent.org` being compromised. The current version wouldn't detect if someone replaced the opa binary with some malicious binary and also replace the digest. It would help in detecting a corrupt file due to download issues, though.

simonbaird: Ah yeah ok. Let me try again.

Makefile
```diff
@@ -24,11 +24,17 @@ fmt-check:
...
 OPA_DEST=/usr/bin/opa
 
 install-opa:
-	sudo curl -s -L -o $(OPA_DEST) $(OPA_URL)
+	curl -s -L -O $(OPA_URL)
+	curl -s -L -O $(OPA_SHA_URL)
+	sha256sum -c $(OPA_SHA_FILE)
```
zregvart, line 36: Something like this could work:
```suggestion
	echo "73ab60fab6c28131b4fa8f2e7e44973165c2be0c46e3e6feabce68d7eec9335a  opa_linux_amd64_static" | sha256sum --check
```

Makefile
```diff
@@ -24,11 +24,17 @@ fmt-check:
 ci: fmt-check test
 
 OPA_VER=v0.39.0
-OPA_URL=https://openpolicyagent.org/downloads/$(OPA_VER)/opa_linux_amd64_static
+OPA_FILE=opa_linux_amd64_static
+OPA_URL=https://openpolicyagent.org/downloads/$(OPA_VER)/$(OPA_FILE)
+OPA_SHA_FILE=$(OPA_FILE).sha256
+OPA_SHA_URL=$(OPA_URL).sha256
 OPA_DEST=/usr/bin/opa
```
zregvart, line 31: Perhaps like this so we don't need to `sudo`?
```suggestion
OPA_DEST=$HOME/opa
```

zregvart (commented):
Some ideas...

Makefile
simonbaird: Thanks! Included in new revision.

zregvart (approved):

PR #5, opened 2022-04-05
WIP: Adding policy checks

bcaton85:
Work in progress, will be updated regularly. 

Rego scripts added from a separate PR created by @simonbaird :
https://github.com/redhat-appstudio/infra-deployments/pull/199

### DIFF
--- a/policies/transparency_logs.rego
+++ b/policies/transparency_logs.rego
@@ -1,0 +1,130 @@
+package hacbs.contract.transparency_urls
+
+import future.keywords.every
+
+#
+# Sanity check the configuration of tekton-chains in the cluster
+#
+# Actually this is a weak test since the configuration might have changed
+# since the pipeline ran.
+#
+# Todo:
+# - This doesn't fail if the key is entirely absent, which is bad.
+# - Maybe check that rest the chains configuration is what we require.
+#
+deny[{"msg": msg}] {
+	data.k8s.ConfigMap["chains-config"].data["transparency.enabled"] != "true"
+
+	msg := "Chains configuration has transparency disabled"
+}
+
+#
+# Sanity check that each taskrun has been marked as signed by chains
+#
+# Also not a strong test, but if it says "failed" then we should be failing.
+#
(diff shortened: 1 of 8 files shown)


simonbaird:
Some initial notes:
* Let's think of a different directory for scripts other than "hack". Maybe just "scripts" is a good enough name for it.
* Perhaps mention in the commit message that these files are being copied from an unmerged POC in a different repo.
* It should be easy to run `make fmt` and fix all the formatting so the CI will be happy.

policies/transparency_logs.rego
```diff
@@ -0,0 +1,130 @@
...
+# Sanity check the configuration of tekton-chains in the cluster
+#
+# Actually this is a weak test since the configuration might have changed
+# since the pipeline ran.
+#
+# Todo:
+# - This doesn't fail if the key is entirely absent, which is bad.
+# - Maybe check that rest the chains configuration is what we require.
+#
+deny[{"msg": msg}] {
```
caugello, line 15: I think this should be in a `config` policy, we will probably validate other settings in the future.

simonbaird: You mean move it to its own separate rego file, right? Agreed.

simonbaird: But it could be done a a later commit. This commit is largely just copying stuff from the original PR in the other repo.

bcaton85: Yeah this PR's goal is to just have the policies available for the enterprise task.

simonbaird:
I did some testing, see details in https://github.com/hacbs-contract/ec-tasks/pull/1 .

As per slack discussion, let's merge early and iterate.

I would like to see a jira mentioned in the commit message, and a sentence to explain where this code originated, i.e. the abandoned POC PR in infra-deployments, (since we only get one chance to write a good commit message..)

simonbaird:
Replaced in part by #6 , but note that the scripts are in build-definitions repo now, not here.

PR #9, opened 2022-04-13
chore(HACBS-293): test coverage threshold

caugello:


zregvart (approved):
Let's try, see if we can reach the 100% test coverage

PR #8, opened 2022-04-12
Realistic test data

simonbaird:

PR #10, opened 2022-04-13
Embed "real" data into a rego file for testing

simonbaird:
Also:
- Check in some generated data
- Add some example rego tests showing how we can use it

For testing purposes it will be useful to have realistic test data.
I don't think it's possible to access external data when running
`opa test` so this is a creative way to provide access a useful data
set that can be used to write tests against.

Possibly not the "normal" way that people do testing for opa, but I
think it can work well for us, so let's see how it goes.

The data checked in now comes from a nodejs builder pipeline run in
my local cluster.

### DIFF
--- a/policies/test_data.rego
+++ b/policies/test_data.rego
@@ -1,0 +1,2108 @@
+package test_data
+
+# Generated automatically with `make data-to-rego`
+data := {
+	"cluster": {
+		"ConfigMap": {"chains-config": {
+			"apiVersion": "v1",
+			"data": {
+				"artifacts.oci.storage": "oci",
+				"artifacts.taskrun.format": "in-toto",
+				"artifacts.taskrun.storage": "oci",
+				"transparency.enabled": "true",
+			},
+			"kind": "ConfigMap",
+			"metadata": {
+				"annotations": {
+					"argocd.argoproj.io/sync-options": "SkipDryRunOnMissingResource=true",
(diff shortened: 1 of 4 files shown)


simonbaird:
PTAL, but I'm pretty happy with this, so it's likely to merge soon unless there are objections.

policies/test_data.rego
```diff
@@ -0,0 +1,2108 @@
+package test_data
+
+# Generated automatically with `make data-to-rego`
+data := {
```
simonbaird, line 4: It's pretty large - I guess we could prune out parts we know we don't care about.

zregvart:
Fails on my end:

```shell
$ make data-to-rego
env: ‘./show-ec-data.sh’: No such file or directory
make[1]: *** [Makefile:51: show-json] Error 127
failed to parse Rego source file: 2 errors occurred:
policies/test_data.rego:3: rego_parse_error: unexpected eof token
	data :=
	      ^
policies/test_data.rego:3: rego_parse_error: unexpected eof token: expected rule value term (e.g., data := <VALUE> { ... })
	data :=
	      ^
make[1]: *** [Makefile:25: fmt] Error 2
make: *** [Makefile:74: data-to-rego] Error 2
```

I can't figure out where `show-ec-data.sh` comes from or where is `show-json` defined. Something missing (in my head or here)?

zregvart:
Also, duplicate of #8?

simonbaird:
It expects you have build-definitions repo checked out in ../build-definitions. That repo has the bash scripts in it.

You also need to be signed in to a cluster with a pipeline run.

simonbaird:
In #11 I got the test coverage to 100% without too much trouble, so maybe let's hold off on this one.

Also it's better not to check in a bunch of cluster data that we're about to remove.

simonbaird:
It's number 1 on https://lcarva.github.io/review-rot/ but I want to keep it open since I think maybe there's a useful idea here.

policies/test_data.rego
lcarva: It would be nice to have a PipelineRun attestation here, and maybe we don't need the ConfigMap nor the PipelineRun objects?

Would it be possible to write up how to recreate this test data? I like the idea of embedding it here but this will eventually get stale and we'll struggle to figure out how to refresh it. Wdyt?

simonbaird: To generate it you just do this `make data-to-rego`. It depends on code in the build-definitions repo, which made sense to me at the time.

simonbaird:
Close it anyhow to tidy up.

PR #11, opened 2022-04-20
Rego cleanup

simonbaird:
* Remove a bunch of demo and cluster data based rules
* Organize rego files and packages, introduce shared rules and utilities
* Add test coverage for all the remaining rules


zregvart (approved):
LGTM

policies/lib/all_rekor_attestations.rego
```diff
@@ -0,0 +1,26 @@
...
+# "rekor_host", "log_index", and "data".
+#
+all_rekor_attestations = attestations {
+	attestations := [attestation |
+		rekor_hosts := data.rekor[rekor_host]
+		log_entry := rekor_hosts.index[log_index].entry
+		attestation := {
+			"rekor_host": rekor_host,
+			"log_index": log_index,
+			"data": decode_attestation(log_entry.Attestation),
```
lcarva, line 16: Oh nice!

policies/attestation_type.rego
```diff
@@ -0,0 +1,24 @@
+package hacbs.contract.attestation_type
+
+import data.lib
+
+#
+# Using attestations extracted from rekor transparency log entries,
+# confirm they all have the expected attestation type
+#
+deny[{"msg": msg}] {
+	att := lib.all_rekor_attestations[_]
```
lcarva, line 10: I wonder if we should just have a single directory called `data/attestation`. We can dump all the attestations we care about in the directory, and just let rego verify those.

I don't think rego should care about where those attestations come from. That's the responsibility of cosign. The rego polices should care about the content of the attestation instead.

simonbaird: Yeah, I think you're right. This means the idea that the keys in the data reflect where it came from is not useful.

Q: In your new task definition there's just one attestation. Should we adapt the rego to this model, or do you think it's worthwhile supporting the idea that we validate a list of attestations, (with the list being of size one to begin with).

I have heard others mention we might EC validate a list of images, not just one, but not sure if the task should support that, or if it should run once per image.

simonbaird: Actually it's just dumping an attestation predicate currently, not the full attestation, right?

simonbaird:
As per slack discussion just now, let's merge early and often.

There are some changes needed here so make it compatible with https://github.com/redhat-appstudio/build-definitions/pull/56 but let's address those in the near future in different PRs.

policies/lib/all_rekor_attestations.rego
simonbaird: It is, but I suppose we won't need it soon.. :)

policies/attestation_type.rego
lcarva: It's the payload of the attestation which includes pretty much everything we care about:
```
{
  "payloadType": "application/vnd.in-toto+json",
  "payload": "<this is the predicate>",
  "signatures": [
    {
      "keyid": "SHA256:Ox+5CRzA3UhgdVFNCN4SuUSKL7NwEB5iLnjX4V+Vzs0",
      "sig": "MEQCIDnJ82cgXil5dotGvHqHovhsoISNC1H9KWw1YGIEPTdbAiB3VkbHbWwGJjbupoHTpvzYlQQQfFXi6KkK98zcZ6dcMw=="
    }
  ]
}
```

We probably don't care about the payloadType and signature at the rego level. But I can also change the task to save the whole thing.

lcarva: Thinking about it more. I think it would be better to have rego do the decoding so it has access to the whole thing. I can't tell the future and this additional data doesn't hurt anyone, so let's include it.

lcarva: Well.. the predicate from rekor is wrapped differently from the predicate from oci.... The snippet above is from oci which include `payloadType` and `signatures`.

This is different for Rekor:
```
$ rekor-cli get --uuid 5256689f64f45b3447240ce4fe738fb4f298ffd8c289f542c7cabe0591d18110 --format json | jq '.'

{
  "Attestation": "<this is the prediate>",
  "AttestationType": "",
  "Body": {
    "IntotoObj": {
      "content": {
        "hash": {
          "algorithm": "sha256",
          "value": "de0cae8d3917b76e6e20a6dfaa83040d40ac4148453037162870f451d4f1691b"
        }
      },
      "publicKey": "LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KTUZrd0V3WUhLb1pJemowQ0FRWUlLb1pJemowREFRY0RRZ0FFMUtIRDhkR293QUw5NG03cHFWOW82dThUcHF5WgplUEgzQlFLVmlScFk0ZTlNdkN3UG14bWJhYWcxenVOYkg1MGxnbXlaMzFLVEN3aU10OHlIYW1XWFlnPT0KLS0tLS1FTkQgUFVCTElDIEtFWS0tLS0tCg=="
    }
  },
  "LogIndex": 2031715,
  "IntegratedTime": 1650298256,
  "UUID": "5256689f64f45b3447240ce4fe738fb4f298ffd8c289f542c7cabe0591d18110",
  "LogID": "c0d23d6ad406973f9559f3ba2d1ca01f84147d8ffc5b8445c224f98b9591801d"
}

```

The predicate is the common denominator.

simonbaird: From slack discussion: I think it's all clear. The "payload" value in cosign output matches the "Attestation" value in rekor output. (And if that's wrong then don't worry we'll figure it out later, along with what exactly the predicate is.)

PR #12, opened 2022-04-21
Use image attestation instead of task runs

lcarva:


simonbaird:
LGTM.

simonbaird (approved):

policies/lib/all_attestations.rego
```diff
@@ -0,0 +1,10 @@
+package lib
+
+#
+# A list of all attestations
+#
+all_attestations = attestations {
+	attestations := [attestation |
+		attestation := {"data": data.attestations}
```
simonbaird, line 8: The plural "attestations" makes it look like an array, but I think maybe it's just one attestation now.

policies/lib/all_attestations_test.rego
```diff
@@ -0,0 +1,6 @@
+package lib
+
+test_all_attestations {
+	expected := [{"data": {"foo": 10}}]
+	expected == all_attestations with data.attestations as {"foo": 10}
```
simonbaird, line 5: And here, it indicates data.attestations is singular. So maybe it should be called "attestation"..? Which then makes me unsure if we want to support a list of them, and maybe in that case all_attestations is no longer needed.

(I might be missing something.)

simonbaird: Sounds good to me.

simonbaird: A cloudy memory returns to me: https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/chains/kaniko-cosign-verify.sh#L109 .

So I think the jq magic would need to deal with "{..att1..}\n{..att2.}\n". maybe that's JSONL format?

policies/lib/all_attestations.rego
```diff
@@ -0,0 +1,10 @@
+package lib
+
+#
+# A list of all attestations
+#
+all_attestations = attestations {
+	attestations := [attestation |
+		attestation := {"data": data.attestations[_]}
+	]
+}
```
simonbaird, line 10: There's still the possibility that we should just scrap this file entirely and just use data.attestations instead of lib.all_attestations, and replace the `att.data` with just `att` (for example).

There might be some value in keep it in case we someday decide we want fields in the hash other than "data", (like before we had rekor_host and log_index), and maybe someday we might lift up a sha or something, but right now I think this isn't bringing much value.

Is there another reason to convert ["..."] into `[{data: "..."}]`?

lcarva: Nope. It was just as a way to provide additional metadata for each attestation. Let's get rid of it.

lcarva: Done.

simonbaird:
That's the one. Smash the merge.

simonbaird (approved):

PR #13, opened 2022-04-22
create policy for step annotations

joejstuart:


simonbaird:
One piece missing, if you want to actually have this policy used in the contract, is to add the import and the deny rule in `main.rego`.

policies/step_annotations.rego
```diff
@@ -0,0 +1,25 @@
...
+import data.lib
+
+deny[{"msg": msg}] {
+	att := data.attestations[_]
+
+	some step_index
+
+	# null annotations show up when there are none defined
+	# can there be null annotations?
+	att.predicate.buildConfig.steps[step_index].annotations != null
```
simonbaird, line 12: Not 100% sure it would work, but maybe you could instead put `null` in the list of allowed_annotations, to express more clearly the idea that the policy allows it. Then it's tidier and you don't need this code to handle it specially.

(IIUC this is different from the key being missing entirely. I think then the rule would quietly quit and we won't see a violation.)

policies/step_annotations.rego
```diff
@@ -0,0 +1,25 @@
...
+	annotations := [x | x := att.predicate.buildConfig.steps[step_index].annotations[_]; not valid_annotation(x)]
+	annotations != []
+	msg := sprintf(
+		"step %d has invalid annotations(s): %s",
+		[step_index, annotations],
+	)
+}
+
+valid_annotation(annotation) {
+	lib.config.allowed_annotations[_] = annotation
```
lcarva, line 24: I see that this works, but why is it `=` instead of `==` ?

policies/step_annotations.rego
```diff
@@ -0,0 +1,25 @@
...
+deny[{"msg": msg}] {
+	att := data.attestations[_]
+
+	some step_index
+
+	# null annotations show up when there are none defined
+	# can there be null annotations?
+	att.predicate.buildConfig.steps[step_index].annotations != null
+
+	# if an annotations is defined, it shows up in a list
```
lcarva, line 14: I'm surprised that it is a list of strings!

lcarva (approved):

policies/step_annotations.rego
simonbaird: Because rego..?

(I was trying to explain it, but I realized I don't fully get it myself. Do the tests pass with == ?

joejstuart: Because rego is all I got too. I originally had ==, but had some issues and tried the single = and never changed it back.

lcarva: I think I get it:
> [Unification](https://www.openpolicyagent.org/docs/latest/policy-language/#unification-)  (=) combines assignment and comparison. Rego will assign variables to values that make the comparison true. Unification lets you ask for values for variables that make an expression true.

Because in this case, the variable is already defined on both sides, this is just a comparison. Replacing it with `==` should have the same result.

> [Best practice](https://www.openpolicyagent.org/docs/latest/policy-language/#best-practices-for-equality) is to use assignment := and comparison == wherever possible. The additional compiler checks help avoid errors when writing policy, and the additional syntax helps make the intent clearer when reading policy.

> Under the hood := and == are syntactic sugar for =, local variable creation, and additional compiler checks.

And to prove this, try this version of the function (it works the same way!):
```rego
valid_annotation(annotation) {
	lib.config.allowed_annotations[_] = spam
	spam == annotation
}
```

I think we should go through the code base and change to `==` whenever possible. Wdyt?

policies/step_annotations.rego
lcarva: Digging through the Chains code, I don't see how `annotations` will ever not be `null` :smile: 
It's defined [here](https://github.com/tektoncd/chains/blob/647e29892e2facd8897286fbc782bed74b2264c6/pkg/chains/formats/intotoite6/buildconfig.go#L37) but I can't find any usage of it. @bcaton85 wdyt?

joejstuart: odd, Annotations doesn't seem to be referenced anywhere.

lcarva:
@joejstuart, is this still relevant?

joejstuart:
> @joejstuart, is this still relevant?

It is not. I'll close.

PR #14, opened 2022-04-22
Makefile action for fetching registry attestations

simonbaird:
So we can conveniently test these rego rules for hacking purposes.

It uses cosign download to avoid needing to provide the public signing
key, which would be needed if we did a cosign verify-attestation.

Also:
- Throw in the dummy policy config so we skip the not useful test,
  (which is also what happens in the related task in the
  build-definitions repo)
- Pretty up the makefile a little and mention the new tasks in the
  help
- Add quay.io/buildah just to make the default attestation pass
  the policy check
- Some todo commentary about the repo matching since I realized it
  would generally match on organization (in quay.io at least) rather
  than a repo. That could be a feature, but it's not described well.

Note:
- The file path data/attestations.json would work the same as
  data/attestations/data.json from opa's perspective, but this way
  is more consistent with how it's done elsewhere.
- I chose to keep the older fetch-data scripts intact rather than
  clean them up, even though we're considering them deprecated.
  This follows the approach from the build-definitions repo, and
  also there might be some use for the them as we start working on
  other kinds of policies such as validating pipeline definitions
  or ApplicationSnapshot CRs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,14 +4,24 @@ help:
 	@echo "  make test         # Run all tests"
 	@echo "  make fmt          # Apply default formatting to all rego files"
 	@echo "  make ci           # Check formatting and run all tests"
+	@echo "  make coverage     # Show which lines of rego are not covered by tests"
+	@echo
 	@echo "  make install-opa  # Install opa if you don't have it already (Linux only)"
+	@echo
+	@echo "  make fetch-att    # Fetch an attestation for an image"
+	@echo "                    # Add \`IMAGE=<someimage>\` to fetch a specific attestation"
+	@echo "                    # Note: This is compatible with the 'verify-enterprise-contract' task"
+	@echo
 	@echo "  make fetch-data   # Fetch data for the most recent pipeline run"
-	@echo "                    # Add \`PR=prname\` to fetch a specific pipeline run"
+	@echo "                    # Add \`PR=<prname>\` to fetch a specific pipeline run"
+	@echo "                    # Note: This is compatible with the deprecated 'enterprise-contract' task"
+	@echo "                    # and requires the build-definitions repo checked out in ../build-definitions"
+	@echo
 	@echo "  make show-files   # List data files"
 	@echo "  make show-data    # Show all data visible to opa in one big object"
 	@echo "  make show-keys    # List all the keys in the data"
+	@echo
 	@echo "  make check        # Check rego policies against the fetched data"
-	@echo "  make coverage     # Show which lines of code are not covered by tests"
 
 test:
 	@opa test . -v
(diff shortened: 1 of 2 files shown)


simonbaird:
Example usage:
```
$ make fetch-att                                                  
cosign download attestation "quay.io/lucarval/tekton-test:latest" | \                                                 
  jq -s '[.[].payload | @base64d | fromjson]' > \                                                                     
    /home/sbaird/code/ec-policies/data/attestations/data.json       

$ make show-data                                                  
attestations:                                                                                                         
  - _type: https://in-toto.io/Statement/v0.1                                                                          
    predicate:                                                                                                        
    ...
```

Makefile
```diff
@@ -39,6 +49,37 @@ fmt-check:
...
+#  make fetch-attestation IMAGE=...
+#
+# The format and file path is intended to match what is used in the
+# verify-attestation-with-policy script in the build-definitions repo
+# so you can test your rules as they would be applied by the
+# verify-enterprise-contract task.
+#
+
+ifndef IMAGE
+  IMAGE="quay.io/lucarval/tekton-test:latest"
```
joejstuart, line 72: Should this be here? https://quay.io/organization/hacbs-contract

simonbaird: We could use anything with at attestation, but I think this is what Luiz was using for testing the task, so I figured it's as good as any. Though it might be wise to specify a sha in case a future push doesn't come with an attestation.

simonbaird: I'm expecting user would specify their own as required. The default is just for convenience/laziness.

lcarva: Should the public key be added to this repo?

joejstuart: OK, makes sense

Makefile
```diff
@@ -70,12 +112,14 @@ check:
...
 
+#--------------------------------------------------------------------
+
 OPA_VER=v0.39.0
 OPA_FILE=opa_linux_amd64_static
 OPA_URL=https://openpolicyagent.org/downloads/$(OPA_VER)/$(OPA_FILE)
 OPA_SHA=19a24f51d954190c02aafeac5867c9add286c6ab12ea85b3d8d348c98d633319
 ifndef OPA_BIN
-	OPA_BIN=$(HOME)/bin
+  OPA_BIN=$(HOME)/bin
```
simonbaird, line 122: Intentional. Line 72 didn't work with a tab. Tabs are weird in make I guess? So I switched to spaces here to be consistent with that.

policies/step_image_registries.rego
```diff
@@ -11,6 +11,10 @@ deny[{"msg": msg}] {
 
 	some step_index
 	step = att.predicate.buildConfig.steps[step_index]
+
+	# TODO: This is more likely to extract an organization rather than a repo.
```
lcarva, line 15: I think it depends on the registry. It's perfectly valid to have an image reference without an "organization". BTW, sometimes the "organization" bit is called "namespace". Sometimes, namespace/repo is called repo... I don't think there's a consistent way of calling these things - it's very much registry dependent.

Makefile
```diff
@@ -39,6 +49,36 @@ fmt-check:
...
+# Set IMAGE as required like this:
+#   make fetch-attestation IMAGE=<someimage>
+#
+# The format and file path is intended to match what is used in the
+# verify-attestation-with-policy script in the build-definitions repo
+# so you can test your rules as they would be applied by the
+# verify-enterprise-contract task.
+#
+ifndef IMAGE
+  IMAGE="quay.io/lucarval/tekton-test:latest"
```
lcarva, line 71: Should the public key be added to this repo?

policies/step_image_registries.rego
simonbaird: Ack. Maybe I'll revise the comment.

Makefile
simonbaird: No, I don't care about checking the sig. Just doing a `cosign download`.

lcarva: Oh my bad. I thought `cosign download` would check signature too. This is great.

policies/step_image_registries.rego
simonbaird: Actually I'll take it out since it doesn't belong in this PR..

Makefile
simonbaird: I added some comments and switched to a sha.

lcarva: I tagged the image with an additional tag so it's not garbage collected when the `latest` tag moves to a different image.

simonbaird:
Good enough, let's merge.

Makefile
simonbaird: Oh, I didn't think of that. Thanks.

simonbaird: Now I think maybe just latest was better, but never mind. If we think of a better default later we'll change it.

PR #15, opened 2022-04-25
Policy for asserting test output

zregvart:
This is the simplest policy for asserting the test stream's output. Two
rules are provided: that there is some test data available and that
all[1] data indicates successful outcome.

Ref. https://issues.redhat.com/browse/HACBS-340

[1]there are more than one test data JSON files, currently at two for
`sanity-inspect-image` and `sanity-label-check` tasks.

### DIFF
--- a/policies/test.rego
+++ b/policies/test.rego
@@ -1,0 +1,28 @@
+package hacbs.contract.test
+
+# Check if we have any test data is present
+deny[{"msg": msg}] {
+	not data.test
+	msg := "No test data provided"
+}
+
+# Check if we have any test data provided
+deny[{"msg": msg}] {
+	count(data.test) == 0
+	msg := "Empty test data provided"
+}
+
+deny[{"msg": msg}] {
+	with_results := [result | result := data.test[_].result]
+	count(with_results) != count(data.test)
+
+	msg := "Found tests without results"
+}
+
+# Check if all tests succeeded 
+deny[{"msg": msg}] {
+	all_failed := [failure | data.test[_].result != "SUCCESS"; failure := 1]
+	count(all_failed) > 0
+
+	msg := "All tests did not end with SUCCESS"
+}
--- a/policies/test_test.rego
+++ b/policies/test_test.rego
@@ -1,0 +1,33 @@
+package hacbs.contract.test
+
+test_needs_to_have_data {
+	deny == {{"msg": "No test data provided"}}
+}
+
+test_needs_non_empty_data {
+	deny == {{"msg": "Empty test data provided"}} with data.test as []
+}
+
+test_needs_tests_with_results {
+	deny == {{"msg": "Found tests without results"}} with data.test as [{}]
+}
+
+test_needs_tests_with_results_mixed {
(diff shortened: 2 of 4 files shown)


policies/test.rego
```diff
@@ -0,0 +1,15 @@
+package hacbs.contract.test
```
lcarva, line 1: Should we enable this in `main.rego` ?

policies/test_test.rego
```diff
@@ -0,0 +1,21 @@
...
+test_pass_with_some_data {
+	not deny with data.test as [{}]
+}
+
+test_success_data {
+	not deny with data.test as [{"result": "SUCCESS"}]
+}
+
+test_failure_data {
+	deny == {"msg": "Tests did not end with SUCCESS"} with data.test as [{"result": "FAILURE"}]
```
lcarva, line 16: Can we add a test that mixes results? e.g. `[{"result": "FAILURE"},{"result": "SUCCESS"}]`

policies/test.rego
```diff
@@ -0,0 +1,15 @@
...
+
+# Check if we have any test data provided
+deny = {"msg": msg} {
+	not count(data.test) > 0
+	msg := "No test data provided"
+}
+
+# Check if all tests succeeded 
+deny = {"msg": msg} {
+	results := [r | r := data.test[_].result == "SUCCESS"]
```
lcarva, line 11: I'm reading the [Universal Quantification](https://www.openpolicyagent.org/docs/latest/policy-language/#universal-quantification-for-all) doc, and I'm wondering if the current form is "idiomatic" rego.

Actually, I don't see the doc mentioning the `all` function at all in the language spec! :exploding_head:

zregvart: I think it can be, options are either using negation, comprehensions as we do here, or `every`. I've renamed the `results` to make it a bit more readable.

zregvart: I've must have been reading older version of the reference where `all` builtin function was still documented. It was in fact [deprecated](https://github.com/open-policy-agent/opa/issues/2437), so I used `count` instead here.

zregvart:
Switched to using `deny[{"msg": msg}]` instead of `deny = {"msg": msg}` to be in line with the existing policies -- otherwise output would be incorrect. Found an additional edge case where `"results"` key was not provided in the data. This should be good for merge I've tested against https://github.com/redhat-appstudio/build-definitions/pull/72 and it looks good.

policies/test.rego
```diff
@@ -0,0 +1,28 @@
+package hacbs.contract.test
+
+# Check if we have any test data is present
+deny[{"msg": msg}] {
+	not data.test
+	msg := "No test data provided"
+}
```
simonbaird, line 7: Cool, that's a useful pattern since we otherwise might pass if there's no attestation for example.

simonbaird:
Great stuff!

simonbaird:
Merge when you like as far as I'm concerned.

lcarva (approved):

policies/test.rego
zregvart: Yeah, I would not have thought of this if I didn't do it TDD-style. There are other things in here that a novice Rego writer, such as myself, would be tripped upon.

PR #19, opened 2022-04-27
Add make helper to run entr for live TDD

simonbaird:
Stolen from @lcarva 's suggestion on slack just now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,13 @@ quiet-test:
 	@opa test .
 	@opa test . --threshold 100 >/dev/null 2>&1
 
+# Do `dnf install entr` then run this a separate terminal or split window while hacking
+live-test:
+	@trap exit SIGINT; \
+	while true; do \
+	  git ls-files -c -o '*.rego' | entr -d -c $(MAKE) --no-print-directory quiet-test; \
+	done
+
 # Rewrite all rego files with the preferred format
 # Use before you commit
 fmt:
@@ -135,5 +142,5 @@ install-opa:
 
 #--------------------------------------------------------------------
 
-.PHONY: help test fmt fmt-check ci install-opa check \
-  clean-data dummy-config fetch-att fetch-data show-data
+.PHONY: help test coverage quiet-test live-test fmt fmt-check ci clean-data \
+  dummy-config fetch-att show-data fetch-data check install-opa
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -35,6 +35,11 @@ quiet-test:
 	@opa test .
 	@opa test . --threshold 100 >/dev/null 2>&1
 
+# Install https://github.com/eradman/entr then run this in a
+# separate terminal or split window while hacking
+live-test:
+	@git ls-files '*.rego' | entr -c $(MAKE) --no-print-directory quiet-test
```
lcarva, line 41: `git ls-files '*.rego'` will miss new files that have not yet been added to git. We probably want `find . -name '*.rego'`

simonbaird: good point.

zregvart (approved):
/me likes

Makefile
simonbaird: `git ls-files -c -o` should do it.

simonbaird: ..and a loop. `entr -d` means quit if any files appeared or disappeared.

simonbaird:
Ugh, now you can't quit with ctrl-c

simonbaird:
Trapped the sigint. I think now it's good.

Makefile
```diff
@@ -135,5 +142,5 @@ install-opa:
 
 #--------------------------------------------------------------------
 
-.PHONY: help test fmt fmt-check ci install-opa check \
-  clean-data dummy-config fetch-att fetch-data show-data
+.PHONY: help test coverage quiet-test live-test fmt fmt-check ci clean-data \
+  dummy-config fetch-att show-data fetch-data check install-opa
```
simonbaird, line 146: I just tidied them up and put them in order. FWIW, we could probably get away with dumping these since IIUC the only impact would be if we had a file or dir with the same name as one of them.

Makefile
```diff
@@ -35,6 +35,13 @@ quiet-test:
 	@opa test .
 	@opa test . --threshold 100 >/dev/null 2>&1
 
+# Do `dnf install entr` then run this a separate terminal or split window while hacking
+live-test:
+	@trap exit SIGINT; \
+	while true; do \
+	  git ls-files -c -o '*.rego' | entr -d -c $(MAKE) --no-print-directory quiet-test; \
```
lcarva, line 42: Sweet!

lcarva (approved):

PR #20, opened 2022-04-29
DNM experiment with error codes and messages

simonbaird:
This doesn't work, but it shows some ideas I'm thinking about.

I'm saying "code" but it's a short string not a number.

### DIFF
--- a/policies/lib/messages.rego
+++ b/policies/lib/messages.rego
@@ -1,0 +1,30 @@
+package lib.messages
+
+import data.lib
+
+#
+# If we extract the static parts this could become a big yaml file maybe...
+# Ps, wtf is opa fmt doing here.. :/
+messages := {"step_image_disallowed": {
+	"fail_message": "Step %d has disallowed registry '%s' for attestation.",
+	"pass_message": sprintf(
+		"All task steps run on images from the allowed registries: %s",
+		[lib.quoted_values_string(lib.config.allowed_registries)],
+	),
+}}
+
+# Imagine many other keys in this hash, one for each kind of failure,
+# and then imagine (somehow) producing a list of all the pass_messages
+# that are not already included in the failure output.
+
+fail_message(message_key, sprintf_values) = message {
+	# Could return a hash with a keys like "fail_code" and "reason" in future maybe instead
+	# of just a string by itself
+	full_message_template := concat("", ["[%s] ", messages[message_key].fail_message])
+	message := sprintf(full_message_template, array.concat([message_key], sprintf_values))
+}
+
+pass_message(message_key) = message {
+	full_message_template := concat("", ["[%s] ", messages[message_key].pass_message])
+	message := sprintf(full_message_template, [message_key])
+}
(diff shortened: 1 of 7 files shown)


policies/step_image_registries_test.rego
```diff
@@ -18,6 +18,6 @@ test_image_registry_valid {
 test_attestation_type_invalid {
 	deny_set := deny with data.attestations as prepare_mock_attestation_data(bad_image)
 	count(deny_set) == 1
-	expected_msg := "Step 0 has disallowed registry 'hackz.io/openshift-pipelines' for attestation."
+	expected_msg := "[step_image_disallowed] Step 0 has disallowed registry 'hackz.io/openshift-pipelines' for attestation."
```
simonbaird, line 21: This format isn't great, that's what make me think a hash would be better, e.g. `{ "reason_code": ..., "reason_message": ... }`

simonbaird:
FYI @hacbs-contract/devs

policies/lib/messages.rego
```diff
@@ -0,0 +1,30 @@
...
+		[lib.quoted_values_string(lib.config.allowed_registries)],
+	),
+}}
+
+# Imagine many other keys in this hash, one for each kind of failure,
+# and then imagine (somehow) producing a list of all the pass_messages
+# that are not already included in the failure output.
+
+fail_message(message_key, sprintf_values) = message {
+	# Could return a hash with a keys like "fail_code" and "reason" in future maybe instead
```
lcarva, line 21: I think a hash would be a much more natural representation of the data. I just thought of something... what if we have this sort of data structure for `data.hacbs.contract.main.results`:

```json
{
  "allow": false,
  "tests": {
    "<check1_key>": {
      "pass": true,
      "description": "<check1_description>"
    },
    "<check2_key>": {
      "pass": true,
      "description": "<check2_description>"
    },
    "<check3_key>": {
      "pass": false,
      "description": "<check3_description>"
      "message": "<something helpful>"
    }
  }
}
```

That way, we could first query for the value of `data.hacbs.contract.main.results` to provide a list of all the checks that were executed, then query for the value of `data.hacbs.contract.main.results.allow` to pass/fail the task.

This could be two calls to `opa`, or maybe a single call to `opa` followed by a call to `jq` magic to pass/fail the task.

wdyt?

lcarva: The data structure I suggested can definitely be improved. I guess the essence of my comment is about providing more information to users than just a boolean.

simonbaird: Yeah that could work.

We might need to worry about the overall size. If we have many descriptions, then it could start to approach the 4k limit. This makes me think we could have a url that links to a description instead of the inline description.

simonbaird: I see that as being doable in the `main.final_result`. It could arrange and collate the data it whatever way we like.

zregvart: > We might need to worry about the overall size.

We could have a smaller Tekton result and push the detailed output in an attestation image.

zregvart: I think there was a requirement to generate a report, so data for the report needs to be persisted somewhere.

simonbaird:
Superseded by #32 .

PR #21, opened 2022-05-02
Nicer coverage threshold not met message

zregvart:
We deliberately dropped the output of `opa test --threshold ...` as when the test coverage target is met `opa` will output a large JSON file with the coverage detail.

If the `make test` fails on the test coverage not being met it fails with this message:

```
...
PASS: 26/26
make: *** [Makefile:28: test] Error 2
```

This could lead to confusion about what actually failed, so with this in such cases the message is now:

```
...
PASS: 26/26
ERROR: Code coverage threshold not met: got 99.35 instead of 100.00
make: *** [Makefile:28: test] Error 2
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+SHELL := /bin/bash
+COVERAGE = @opa test . --threshold 100 2>&1 | sed -e '/^Code coverage/!d' -e 's/^/ERROR: /'; exit $${PIPESTATUS[0]}
 
 help:
 	@echo "Usage:"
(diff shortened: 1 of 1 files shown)


simonbaird (approved):
LGTM. One thought - perhaps quiet test should have the same output. Quiet isn't well named here perhaps. I consider it more like "quiet when passing but loud when failing".

lcarva (approved):

simonbaird:
I like it.

PR #22, opened 2022-05-02
Refactor to generalize

zregvart:
This moves the policy rules to `policies` package so we can generalize over all policies. That is, in `main` we no longer need to call out each policy by name.
The test for main is now a bit more fragile as changes to the policies could fail the test, this is to assert that the output from main hasn't changed in it's shape (say from a set to an array).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ fetch-data: fetch-
 
 POLICIES_DIR=$(THIS_DIR)/policies
 OPA_FORMAT=pretty
-OPA_QUERY=data.hacbs.contract.main.deny
+OPA_QUERY=data.main.deny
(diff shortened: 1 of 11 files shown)


simonbaird (approved):
Lgtm. Could be out of scope in this commit, but did you consider ditching the "hacbs.contract" name space? Not sure if it's bringing much value.

zregvart:
> Lgtm. Could be out of scope in this commit, but did you consider ditching the "hacbs.contract" name space? Not sure if it's bringing much value.

Need to remember to update the task as well...

lcarva (approved):

simonbaird:
Lgtm.

PR #23, opened 2022-05-02
Add support for future policies

lcarva:
https://issues.redhat.com/browse/HACBS-308

I consider this PR complete but I'm marking it as a draft because a discussion is needed to determine if this is the right direction we want to take.

### DIFF
--- a/policies/main.rego
+++ b/policies/main.rego
@@ -6,12 +6,22 @@ deny = {denial |
 	denial := data.policies[policy].deny[_]
 }
 
+future_deny = {denial |
+	skip_not_in_effect(policy)
+	data.policies[policy].deny[_]
+	denial := data.policies[policy].deny[_]
+}
+
 skip(test_name) {
 	data.config.policy.non_blocking_checks[_] == test_name
 }
 
-skip(policy_name) {
+skip_not_in_effect(policy_name) {
 	# Use the nanosecond epoch defined in the policy if present. Otherwise, use now.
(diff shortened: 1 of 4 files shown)


policies/main.rego
```diff
@@ -9,3 +9,9 @@ deny = {denial |
 skip(test_name) {
 	data.config.policy.non_blocking_checks[_] == test_name
 }
+
+skip(policy_name) {
+	# Use the nanosecond epoch defined in the policy if present. Otherwise, use now.
+	when_ns := object.get(data.config.policy, ["when_ns"], time.now_ns())
```
zregvart, line 15: TIL from the [docs](https://www.openpolicyagent.org/docs/latest/policy-reference/#time):

> Multiple calls to the `time.now_ns` built-in function within a single policy evaluation query will always return the same value.

I thought there might be a subtle race condition here but that is not the case.

policies/main.rego
```diff
@@ -9,3 +9,9 @@ deny = {denial |
 skip(test_name) {
 	data.config.policy.non_blocking_checks[_] == test_name
 }
+
+skip(policy_name) {
```
zregvart, line 13: TIL [Logical OR](https://www.openpolicyagent.org/docs/latest/#logical-or)

zregvart (approved):
Simple and effective, I thought it might be interesting to see as an output of the execution what policy checks will fail in the future so I created https://github.com/lcarva/ec-policies/pull/1

PR #24, opened 2022-05-02
Fix rego formatting

lcarva:


lcarva:
Hmm not quite sure why CI still reports a failure. `make fmt-check` does not return any errors with this change on my local system.

lcarva:
Ha! There was a fix on opa 39: https://github.com/open-policy-agent/opa/issues/4376

PR #26, opened 2022-05-04
Upgrade opa CLI to v0.40.0

zregvart:


simonbaird:
I didn't test it, but if it works in Github we should be good to go.

simonbaird (approved):

PR #27, opened 2022-05-09
Conftest demo

joejstuart:


zregvart:
> /retest

We use GitHub actions, there is a `Re-run all jobs` button on the [Checks](https://github.com/hacbs-contract/ec-policies/pull/27/checks) tab.

Currently the checks fail on the Rego files format check and `make fmt` should fix those. Make sure to use a recent `opa` CLI though, we use 0.40 for GitHub `make install-opa` should install the same version as used on GitHub Actions.

PR #28, opened 2022-05-11
Nicer `make help` output

zregvart:
This is stolen from the Makefile generated by kubebuilder. The `help` rule now runs `awk` over the Makefile and interprets `#@` and `##` comments to determine categories and descriptions of rules. This way we don't need to update the text in the `help` rule and the description of the rule is now inline with the rule itself.

Output now looks like:

```
make help

Usage:
  make <target>

General
  help             Display this help.

Development
  test             Run all tests in verbose mode and check coverage
  coverage         Show which lines of rego are not covered by tests
  quiet-test       Run all tests in quiet mode and check coverage
  live-test        Continuously run tests on changes to any `*.rego` files, `entr` needs to be installed
  fmt              Apply default formatting to all rego files. Use before you commit
  opa-check        Check Rego files with strict mode (https://www.openpolicyagent.org/docs/latest/strict/)

CI
  fmt-check        Check formatting of Rego files. Used in CI.
  ci               Runs all checks and tests. Used in CI.

Data helpers
  clean-data       Removes everything from the `./data` directory
  dummy-config     Changes the configuration to mark the `not_useful` check as non-blocking to avoid a "feels like a bad day.." violation
  fetch-att        Fetches attestation data for IMAGE, use `make fetch-att IMAGE=<ref>`. Note: This is compatible with the 'verify-enterprise-contract' task
  show-data        Dump available data in `./data` as YAML
  fetch-data       Fetch data for the most recent pipeline run. Add `PR=<prname>` to fetch a specific pipeline run. Note: This is compatible with the deprecated 'enterprise-contract' task and requires the build-definitions repo checked out in ../build-definitions

Running
  check            Run policy evaluation with currently fetched data in `./data` and policy rules in `./policies`

Utility
  install-opa      Install `opa` CLI from GitHub releases
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,77 +1,88 @@
 SHELL := /bin/bash
 COVERAGE = @opa test . --threshold 100 2>&1 | sed -e '/^Code coverage/!d' -e 's/^/ERROR: /'; exit $${PIPESTATUS[0]}
 
-help:
-	@echo "Usage:"
-	@echo "  make test         # Run all tests"
-	@echo "  make fmt          # Apply default formatting to all rego files"
-	@echo "  make ci           # Check formatting and run all tests"
-	@echo "  make coverage     # Show which lines of rego are not covered by tests"
-	@echo
-	@echo "  make install-opa  # Install opa if you don't have it already (Linux only)"
-	@echo
-	@echo "  make fetch-att    # Fetch an attestation for an image"
-	@echo "                    # Add \`IMAGE=<someimage>\` to fetch a specific attestation"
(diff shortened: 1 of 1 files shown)


simonbaird (approved):

lcarva (approved):
Love it!

zregvart:
Thanks for the feedback. I've added word wrap and instead of printing the `<goal>` `make <goal>` is printed.

simonbaird:
Beautiful.

lcarva:
This is fantastic

PR #29, opened 2022-05-11
rewriting to use conftest

joejstuart:


simonbaird:
I'm getting a `Not found` on the curl command to download conftest.

simonbaird:
Will debug some more tomorrow.

joejstuart:
> ~Looks like there's no .tar.gz file for amd64~
> 
> Actually I fixed it by moving this so it applies for Linux too:
> 
> ```
> +ifeq ($(CONFTEST_GOARCH),amd64)
> +    CONFTEST_GOARCH=x86_64
> +endif
> ```
> 
> Not sure if the arch difference matters, but seems to work so far.

I'm not sure either. For some reason I thought it was only a Mac thing. I moved the `ifeq`

simonbaird:
Re the coverage support for conftest, I found this: https://github.com/open-policy-agent/conftest/issues/518 .

So maybe we need to keep opa binary around for the coverage check.

simonbaird:
There's some pieces here that aren't in #30 so I think this PR is still good. Needs a rebase and some careful selection of hunks.

simonbaird:
Or perhaps it's better to start fresh and add the `make fmt` and make `install-conftest` change in a new PR.

lcarva:
Is this still relevant?

PR #31, opened 2022-05-19
Use annotations as the source of effective_on date

zregvart:
Demonstrates how annotations can be placed in different scopes and how to fetch the effective value of the annotation based on a scoped precedence.

Also includes:

[Different scopes of annotations](https://github.com/hacbs-contract/ec-policies/commit/98af5b0243b07b25f24b0e78c33eb37db4147dd8)

Demonstrates how annotations can be placed in different scopes and how to fetch the effective value of the annotation based on a scoped precedence.

### DIFF
--- a/policies/examples/time_base_test.rego
+++ b/policies/examples/time_base_test.rego
@@ -3,19 +3,27 @@
 # package under test by name in these particular tests.
 package examples.time_based_test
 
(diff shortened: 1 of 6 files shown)


lcarva (approved):

joejstuart (approved):
Nice work!

PR #32, opened 2022-05-24
Annotation code and failure msg

simonbaird:
The interesting part is probably `lib.result_helper` but I'm pretty happy with the assertion rules too.


policies/attestation_type.rego
```diff
@@ -2,22 +2,22 @@ package policies.attestation_type
 
 import data.lib
 
+# Currently this is the only type we know about
+known_types := ["https://in-toto.io/Statement/v0.1"]
+
+# METADATA
+# custom:
+#   short_name: unknown_att_type
+#   name: An unknown attestation type was found
```
zregvart, line 11: Perhaps it would make sense to take advantage of the [builtin annotations](https://www.openpolicyagent.org/docs/latest/annotations/#annotations) (`title` and `description`)

zregvart (approved):
LGTM

PR #35, opened 2022-05-26
Generate policy docs from rego annotations

simonbaird:
JIRA: HACBS-522

### DIFF
--- a/.github/workflows/pre-merge-ci.yaml
+++ b/.github/workflows/pre-merge-ci.yaml
@@ -22,5 +22,8 @@ jobs:
     - name: Install opa
       run: make install-opa OPA_BIN=$HOME/.local/bin
 
+    - name: Install gomplate
+      run: make install-gomplate GOMPLATE_BIN=$HOME/.local/bin
+
     - name: Run checks
       run: make ci
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,16 @@ fmt: ## Apply default formatting to all rego files. Use before you commit
 opa-check: ## Check Rego files with strict mode (https://www.openpolicyagent.org/docs/latest/strict/)
 	@opa check . --strict
 
+DOCS_BUILD_DIR=$(THIS_DIR)/docs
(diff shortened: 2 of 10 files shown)


simonbaird:
It looks like this: https://gist.github.com/simonbaird/8e0b9cc44e7dfec03b8efade89525569 .

.github/workflows/pre-merge-ci.yaml
```diff
@@ -24,3 +24,17 @@ jobs:
...
     - name: Run checks
       run: make ci
+
+    - name: Install gomplate
+      run: make install-gomplate GOMPLATE_BIN=$HOME/.local/bin
+
+    - name: Build docs
+      run: make docs
+
+    # FIXME: This produces a zip file so it's not much use.
```
lcarva, line 34: GitHub pages I think is the way to go. This is what I did for https://lcarva.github.io/review-rot/

Relevant commits:
* https://github.com/lcarva/review-rot/commit/56702d02af1e07c5c272b57c8b017b89bb043c70
* https://github.com/lcarva/review-rot/commit/c03948b019afdd6b052f7286645faf350a6d346c
* https://github.com/lcarva/review-rot/commit/0a09019bf3895d1fe770759e11af249d7cd500e8

lcarva (commented):
This is neat!

simonbaird:
Notes about this current revision:
* Ditch the artifact and instead check in the generated docs file, but add a CI check to make it less annoying to keep it updated.
* It's also visble in pages, here is a preview: https://simonbaird.github.io/ec-policies/

.github/workflows/pre-merge-ci.yaml
simonbaird: Looks like it's working, see https://simonbaird.github.io/ec-policies/ .

Actually didn't need to do much, other than add a docs/_config.yml file turn it on in settings.

lcarva (approved):
:rocket:

simonbaird:
I'm putting more work on top of this, which would make the PRs hard to manage. Let's merge.

PR #37, opened 2022-05-27
check for existence and correct bundle image

joejstuart:


policies/attestation_task_bundle.rego
```diff
@@ -0,0 +1,36 @@
+package policies.attestation_task_bundle
+
+import data.lib
+
+allowed_bundles := ["quay.io/appstudio-utils:latest"]
```
lcarva, line 5: These seem to be the valid ones:

* quay.io/redhat-appstudio/build-templates-bundle
* quay.io/redhat-appstudio/hacbs-templates-bundle

Maybe just a prefix for now as there's no `latest` tag in those repos, only git commit tags which change very very frequently.

policies/attestation_task_bundle.rego
```diff
@@ -0,0 +1,36 @@
...
+
+# METADATA
+# title: Task bundle was not used or is not defined
+# description: |-
+#   Check for existence of a task bundle. Enforcing this rule will
+#   fail the contract if the task is not called from a bundle.
```
lcarva, line 11: It would be useful to see which bundles are allowed in the docs. I like the approach that Simon took [here](https://github.com/hacbs-contract/ec-policies/blob/d71af3c03431fc3bbc39aed3c9a2d295efffe6d2/policies/step_image_registries.rego#L14-L21). Basically, define the allowed bundles as a metadata custom value, then read from it in the rule definition.

policies/attestation_task_bundle.rego
```diff
@@ -0,0 +1,36 @@
...
+# custom:
+#   short_name: disallowed_task_reference
+#   failure_msg: Task '%s' does not contain a bundle reference
+deny[result] {
+	task := lib.tasks[_]
+	name := task.name
+	not task.ref.bundle
```
lcarva, line 18: Will this catch an empty string as well? Maybe a unit tests is helpful here.

policies/attestation_task_bundle.rego
joejstuart: Good idea. I'll update.

policies/attestation_task_bundle.rego
joejstuart: Now that is cool.

policies/attestation_task_bundle.rego
joejstuart: This rule doesn't catch it, but the second rule does. Is that good enough or is there value having this rule catch it also? I did add a test for it just now.

lcarva: Nah. All good.

policies/attestation_task_bundle.rego
```diff
@@ -0,0 +1,38 @@
...
+#   failure_msg: Task '%s' has disallowed bundle image '%s'
+#   allowed_bundles:
+#   - quay.io/redhat-appstudio/build-templates-bundle
+#   - quay.io/redhat-appstudio/hacbs-templates-bundle
+
+deny[result] {
+	task := lib.tasks[_]
+	name := task.name
+	bundle := task.ref.bundle
+	not lib.item_in_list(bundle, rego.metadata.rule().custom.allowed_bundles)
```
lcarva, line 36: I think we want to do prefix check here instead, right?

lcarva (approved):

PR #39, opened 2022-05-27
Short cuts for committing docs and fmt changes

simonbaird:
You pretty much always want to add the fmt and docs changes to your
commit, let's make that easier.

I amended the wrong commit a few times, to decided to add a
confirmation so you have a chance to bail out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,13 @@ conftest-test: ## Run unit tests with conftest
 fmt: ## Apply default formatting to all rego files. Use before you commit
 	@opa fmt . --write
 
+amend-fmt: fmt ## Apply default formatting to all rego files then amend the current commit
+	@git diff $$(git ls-files '*.rego')
+	@echo "Amend commit '$$(git log -n1 --oneline)' with the above diff?"
+	@read -p "Hit enter to continue, Ctrl-C to abort."
+	git add $$(git ls-files '*.rego')
+	git commit --amend --no-edit
+
 opa-check: ## Check Rego files with strict mode (https://www.openpolicyagent.org/docs/latest/strict/)
 	@opa check . --strict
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -63,6 +63,13 @@ conftest-test: ## Run unit tests with conftest
 fmt: ## Apply default formatting to all rego files. Use before you commit
 	@opa fmt . --write
 
+amend-fmt: fmt ## Apply default formatting to all rego files then amend the current commit
```
lcarva, line 66: Do we need an update to `.PHONY`?

Makefile
```diff
@@ -63,6 +63,13 @@ conftest-test: ## Run unit tests with conftest
...
+	@git diff $$(git ls-files '*.rego')
+	@echo "Amend commit '$$(git log -n1 --oneline)' with the above diff?"
+	@read -p "Hit enter to continue, Ctrl-C to abort."
+	git add $$(git ls-files '*.rego')
```
lcarva, line 70: `-u` might be useful here to avoid accidentally adding untracked files.

Makefile
```diff
@@ -63,6 +63,13 @@ conftest-test: ## Run unit tests with conftest
 fmt: ## Apply default formatting to all rego files. Use before you commit
 	@opa fmt . --write
 
+amend-fmt: fmt ## Apply default formatting to all rego files then amend the current commit
+	@git diff $$(git ls-files '*.rego')
+	@echo "Amend commit '$$(git log -n1 --oneline)' with the above diff?"
```
lcarva, line 68: I wonder if this prompt would be better displayed after running `git add` so we can also display which files are being added.

Makefile
simonbaird: Yep. Thanks.

Makefile
simonbaird: I think it's fine since we're doing the `git ls-files` instead of `find .`

Makefile
simonbaird: The git diff does show that info. Also I'd rather not leave things staged if user wants to abort. It's a little clunky if there's no diff, since the "above diff" is blank, so it might be nice to skip everything unless there's a diff, but maybe it's good enough as is.

zregvart (approved):
LGTM, merge?

Makefile
lcarva: You are correct

Makefile
lcarva: Good point!

lcarva (approved):

PR #40, opened 2022-05-31
Support more than one namespace for policies

simonbaird:
This grew a little from the original version, but the main intent is the same, i.e. support running different policies when verifying different things, and get ready to do that for verifying a pipeline definition.


Makefile
```diff
@@ -172,9 +172,12 @@ CONFIG_DATA_FILE=$(DATA_DIR)/config.json
 INPUT_DIR=./input
 INPUT_FILE=$(INPUT_DIR)/input.json
 
+RELEASE_NAMESPACE=release.main
+
 POLICIES_DIR=./policies
 OPA_FORMAT=pretty
-OPA_QUERY=data.main.deny
+OPA_QUERY=data.$(RELEASE_NAMESPACE).deny
```
joejstuart, line 179: With conftest you can just specify the namespace. This might only select the deny rules, I'm not sure though.

robnester-rh:
Wasn't there discussion around renaming the `policies` folder to `policy` as it's the "standard" name expected?

robnester-rh (approved):

policies/pipeline/basic.rego
```diff
@@ -0,0 +1,22 @@
...
+import data.lib
+
+expected_kind := "Pipeline"
+
+# (Not sure if we need this, but I'm using it to test the docs build.)
+
+# Fixme: It doesn't fail if the kind key is entirely missing..
+
+# METADATA
+# title: Check the kind is "PipelineRun"
```
zregvart, line 12:
```suggestion
# title: Check the kind is "Pipeline"
```

zregvart (approved):
Looks straightforward

docs/index.md
```diff
@@ -13,8 +13,21 @@ Those policies are defined using
...
 described here.
 
-Policy Rules
-------------
+Pipeline Policy
+----------------
+
+### Basic Rules
+
+#### `[unexpected_kind]` Check the kind is "PipelineRun"
```
robnester-rh, line 21: Should this be `"Pipeline"` instead of `"PipelineRun"`?

policies/pipeline/basic.rego
robnester-rh: +1 on this suggested change, see associated comment in `index.md`

robnester-rh (commented):
Pending until all commits are finalized. Notes left.

docs/index.md
simonbaird: Oh yeah, thanks!

joejstuart (approved):
Looks good. Do we have to coordinate merging with an update to `--policy` in build-definitions? https://github.com/redhat-appstudio/build-definitions/blob/main/appstudio-utils/util-scripts/check-ec-policy.sh#L20

simonbaird:
I want to merge, but first we need to:
* Pin the sha in e2e
* Get a MR ready for the EC task (as @joejstuart suggests above)

simonbaird:
Since https://github.com/redhat-appstudio/e2e-tests/pull/100 is merged, I think we can merge this without breaking e2e. So here goes...

PR #41, opened 2022-06-01
create source image task verification

joejstuart:


policies/source_image_task.rego
```diff
@@ -0,0 +1,18 @@
...
+# title: Verify the source-image-verify task accepts an image
+# description: |-
+#    Verify the source-image-verify task accepts a particular image
+#    as an input param.
+# custom:
+#   short_name: disallowed_input_image
+#   failure_msg: Task '%s' does not contain '%s' as a param
+warn[result] {
+	image_sha = data.source_image_verify.input_image
+	task_name = data.source_image_verify.task_name
```
simonbaird, line 15: I don't understand where data.source_image_verify comes from here.

simonbaird: What's a typical value for input_image and task_name?

joejstuart: The `input_image` is coming from the release pipeline, so my initial thought was the release pipeline had to pass this information down. Zoran brought up pulling it from the attestation instead. I need to test it out, but it looks like that's doable and a better solution. So, in short, this code most likely is changing.

simonbaird: Ack, ok.

simonbaird: If it is a manually set value, maybe it can live in the EnterpriseContractPolicy .

PR #43, opened 2022-06-01
Display future failures as conftest warnings

lcarva:
Issue: [HACBS-520](https://issues.redhat.com/browse/HACBS-520)

## Add default `effective_on` date to all rules

Every existing rule has a default `effective_on` date set to the beginning
of the year 2022. This is an arbitrary date to capture rules that have
already been effective before the introduction of time-based rules.

As per previous design decision, every policy rule should have an
explicit `effective_on` date to better assist the policy rule authors when
introducing a new policy rule.

## Expose `effective_on` in rules result

In addition to the error message, `msg`, and `metadata.code`, a rule
result will also include `metadata.effective_on`. This is meant to
assist users in determining when a blocking issue has been introduced.

## Rename `future_deny` to warn

So it is automatically reported as warnings by conftest.

Since `metadata.effective_on` is already present in the rules result,
this also aids users in determining when a warning will be treated as a
hard failure.

### DIFF
--- a/checks/annotations.rego
+++ b/checks/annotations.rego
@@ -57,6 +57,9 @@ violation[msg] {
 	# just examine Rego files that declare policies
 	annotation.location.file == file
 
+	# ... and ignore non-rule annotations, e.g. package, document.
+	annotation.annotations.scope == "rule"
+
 	# gather all annotations in a dotted format (e.g. "custom.short_name")
 	declared_annotations := union({a |
 		annotation.annotations[x]
--- a/policies/lib/string_utils.rego
+++ b/policies/lib/string_utils.rego
@@ -1,5 +1,7 @@
 package lib
 
+import data.lib.time
+
 import future.keywords.in
 
 item_in_list(item, list_or_set) {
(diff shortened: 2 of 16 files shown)


checks/annotations.rego
```diff
@@ -57,6 +57,9 @@ violation[msg] {
 	# just examine Rego files that declare policies
 	annotation.location.file == file
 
+	# ... and ignore non-rule annotations, e.g. package, document.
+	annotation.annotations.scope == "rule"
```
zregvart, line 61: I'll improve the checks in a followup PR.

policies/lib/string_utils.rego
```diff
@@ -21,9 +23,14 @@ quoted_values_string(value_list) = result {
 	result := concat(", ", quoted_list)
 }
 
-result_helper(rule_metadata, failure_sprintf_params) = result {
+result_helper(chain, failure_sprintf_params) = result {
```
zregvart, line 26: We might need a test for this

zregvart (approved):
Thanks, nothing really of note, nitpickery, we should merge this

policies/lib/string_utils.rego
lcarva: Ack. I think we should also move it out of `string_utils.rego` because it doesn't really feel like a string utility.

lcarva:
> Thanks, nothing really of note, nitpickery, we should merge this

Right on. Merging now.

PR #55, opened 2022-06-09
Support doc inclusion in the HACBS antora docs build

simonbaird:
The signficant change is in the first commit. The other three commits are related tweaks and improvements.

### DIFF
--- a/antora-docs/modules/ROOT/pages/index.adoc
+++ b/antora-docs/modules/ROOT/pages/index.adoc
@@ -7,13 +7,13 @@
 
 == About
 
-+++<p>The HACBS Enterprise Contract is a Tekton task that can be used to verify the
++++The HACBS Enterprise Contract is a Tekton task that can be used to verify the
 provenence of container images built in HACBS and validate them against a set of
-policies.</p>+++
+policies.+++
 
-+++<p>Those policies are defined using the
++++Those policies are defined using the
 <a href="https://www.openpolicyagent.org/docs/latest/policy-language/">rego policy language</a>
-and are described here.</p>+++
+and are described here.+++
 
 == Pipeline Policy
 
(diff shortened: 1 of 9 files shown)


simonbaird:
Okay, I think it works.

simonbaird:
FYI @gtrivedi88 .

docsrc/index.adoc.tmpl
```diff
@@ -1,12 +1,19 @@
...
 
+:toc: left
+:icons: font
+:numbered:
+
 == About
 
 {{ range $static.about }}
-+++<p>{{ . }}</p>+++
++++{{ . }}+++
```
zregvart, line 16: We probably don't need the pasthrough block any more here since `about` has no HTML in it
```suggestion
{{ . }}
```

simonbaird: There's one link to rego, so that's the reason for it. (The problem goes away if we stop having different formats...)

antora-docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,4 +1,10 @@
 
+////
+This content is automatically generated from a template, see
+https://github.com/hacbs-contract/ec-policies/tree/main/docsrc
+Do not edit it manually.
+////
+
 = HACBS Enterprise Contract Policies
```
zregvart, line 8: I think this will disable the "Edit this page" links, see [default-ui docs](https://docs.antora.org/antora-ui-default/templates/#template-variables)
```suggest
:page-origin-private: true
```

simonbaird: Nice. I think there's some value of having a link to the source repo, so maybe we can let it stay as it is for now.

simonbaird:
Got a lgtm via slack from @gtrivedi88. Let's merge.

simonbaird:
https://red-hat-hybrid-application-cloud-build-services-documentation.pages.redhat.com/hacbs-documentation/ec-policies/index.html is the result.

PR #56, opened 2022-06-13
Remove markdown docs

simonbaird:
Because the asciidoc is now properly integrated in the official
docs, let's stop maintaining the old markdown.

I kept a placeholder version of index.md just so the pages url keeps
working for a little while. Could remove it later.


zregvart (approved):
LGTM, one nitpick...

PR #57, opened 2022-06-14
Readme improvements

simonbaird:
Add more detail about how to fetch pipeline definition data for
hacking. Also move the "running with real data" sections
down a level since they don't seem to warrant being level 2
headings.


simonbaird:
It's Friday, let's merge.

PR #58, opened 2022-06-17
Add appstudio-tasks to permitted task bundles list

simonbaird:
This should get the e2e tests passing again.


simonbaird:
This is low risk and might unblock https://github.com/redhat-appstudio/e2e-tests/pull/101 so I'm going to insta-merge it.

PR #61, opened 2022-07-04
Support "SKIPPED" test result status

zregvart:
With this we support the "SKIPPED" test result status. The definition of `deny` for test results is changed to not being a "FAILURE" or "ERROR", instead of being "SUCCESS"; and a new `warn` is added for tests with the status of "SKIPPED".

Ref. https://issues.redhat.com/browse/HACBS-740

### DIFF
--- a/policy/release/test.rego
+++ b/policy/release/test.rego
@@ -1,6 +1,7 @@
 package policy.release.test
 
 import data.lib
+import future.keywords.in
 
 # METADATA
 # title: No test data found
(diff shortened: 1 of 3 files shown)


policy/release/test.rego
```diff
@@ -35,28 +36,23 @@ deny[result] {
...
 #   failure_msg: "The following tests did not complete successfully: %s"
 #
 deny[result] {
-	# Collect all failed tests and convert their name to "test:<name>" format
-	all_failed := {failure |
-		result := lib.results_from_tests[_]
-		result.result != "SUCCESS"
-		failure := sprintf("test:%s", [result.__task_name])
-	}
+	all_failed = results_with_status({"FAILURE", "ERROR"})
```
lcarva, line 48: If there is a typo on the status, say "EROR", I think the test be marked as being successful. Previously, a typo would assume a failure which I think is a safer  default. Maybe we need to add a check to ensure the state is one of the known states?

zregvart: Perhaps as a warning and in another pull request.

simonbaird (approved):
Lgtm. One grammar nitpick.

The idea of failing on unknown status vs warning on unknown status is worth talking about.

I think if I was voting I would have picked fail on unknown status since it's the more conservative approach, and less likely to release something that should not have been released. Also it's kind of what we have now before this patch. That said I don't feel strongly about it and I don't think it's a blocker for merging.

zregvart:
Fixed grammar issue and added a `deny` rule to check for known/supported test results.

zregvart:
Also took the opportunity to revisit some of the naming...

lcarva (approved):
:rocket:

PR #62, opened 2022-07-05
Validate required tasks present in attestation

zregvart:
This adds two policy checks: `tasks_missing` if there are no tasks in the PipelineRun attestation; and `tasks_required` if the required tasks are not present in the PipelineRun attestation.

In addition to custom annotation for `rule_data` in the rule scope, this also allows custom `rule_data` annotation at the package scope that is nested within the rule's short name. In this case this was necessary so that the list of required tasks is not repeated between the test and the rule implementation. This allows us to define `rule_data` for the `tasks_required` rule and the tests in the single location.

Ref. https://issues.redhat.com/browse/HACBS-676

### DIFF
--- a/policy/release/tasks.rego
+++ b/policy/release/tasks.rego
@@ -1,0 +1,65 @@
+# METADATA
+# custom:
+#   tasks_required:
+#     rule_data:
+#       required_task_refs:
+#       - add-sbom-and-push
+#       - clamav-scan
+#       - deprecated-image-check
+#       - get-clair-scan
+#       - sanity-inspect-image
+#       - sanity-label-check
+#       - sanity-optional-label-check
+#       - sast-go
+package policy.release.tasks
+
+import data.lib
+import future.keywords.in
+
+# This generates all errors that can be omitted from the `tasks_required`
+# rule. Since required tasks can change over time, we need this so we
+# don't need to repeat the list of tasks in the test where this list of
+# errors is also used. It needs to be placed here to be able to access
+# the package level metadata/annotations above.
+all_required_tasks := {t | t := rego.metadata.chain()[_].annotations.custom.tasks_required.rule_data.required_task_refs[_]}
+
+# METADATA
+# title: No tasks run
+# description: |-
+#   This policy enforces that at least one Task is present in the PipelineRun
+#   attestation.
+# custom:
+#   short_name: tasks_missing
+#   failure_msg: No tasks found in PipelineRun attestation
+deny[result] {
+	att := lib.pipelinerun_attestations[_]
+
+	count(att.predicate.buildConfig.tasks) == 0
+
+	result := lib.result_helper(rego.metadata.chain(), [])
+}
+
+# METADATA
+# title: Required tasks not run
+# description: |-
+#   This policy enforces that the required set of tasks is run in a
+#   PipelineRun.
+# custom:
+#   short_name: tasks_required
+#   failure_msg: Required task(s) '%s' not found in the PipelineRun attestation
+deny[result] {
+	att := lib.pipelinerun_attestations[_]
+
+	# reported by tasks_missing above
+	count(att.predicate.buildConfig.tasks) > 0
+
+	# collects names of tasks that are present in the attestation
+	attested_tasks := {t | att.predicate.buildConfig.tasks[_].ref.kind == "Task"; t := att.predicate.buildConfig.tasks[_].ref.name}
+
+	# if all attested_tasks equal all_required_tasks this set is empty,
+	# otherwise it contains the tasks that are required but are not
+	# present in the attestation
+	all_missing := all_required_tasks - attested_tasks
+
+	result := lib.result_helper(rego.metadata.chain(), [concat("', '", all_missing)])
+}
--- a/policy/release/tasks_test.rego
+++ b/policy/release/tasks_test.rego
@@ -1,0 +1,76 @@
+package policy.release.tasks
+
+import data.lib
+
+test_no_tasks_present {
+	expected := {{
+		"code": "tasks_missing",
+		"msg": "No tasks found in PipelineRun attestation",
+		"effective_on": "2022-01-01T00:00:00Z",
+	}}
+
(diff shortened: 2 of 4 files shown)


policy/release/tasks.rego
```diff
@@ -0,0 +1,54 @@
...
+#       - sanity-inspect-image
+#       - sanity-label-check
+#       - sanity-optional-label-check
+#       - sast-go
+package policy.release.tasks
+
+import data.lib
+import future.keywords.in
+
+all_tasks_missing_errors = {e | task := rego.metadata.chain()[_].annotations.custom.tasks_required.rule_data.required_task_refs[_]; e := {
```
lcarva, line 19: ~Should this go in the `tasks_test.rego` file since it's only used there?~
No, because this needs access to the package annotations. I guess we could use `document` annotations but that seems overkill just for this use case. Maybe we can add a comment here explaining why `all_tasks_missing_errors` exists in this file to help us remember in the future?

policy/release/tasks.rego
```diff
@@ -0,0 +1,54 @@
...
+#   PipelineRun.
+# custom:
+#   short_name: tasks_required
+#   failure_msg: Required task '%s' not found in the PipelineRun attestation
+deny[result] {
+	att := lib.pipelinerun_attestations[_]
+	attested_tasks := {t | t := att.predicate.buildConfig.tasks[_].name}
+	required_tasks := {t | t := rego.metadata.chain()[_].annotations.custom.tasks_required.rule_data.required_task_refs[_]}
+	all_missing := required_tasks - attested_tasks
+	missing := all_missing[_]
```
lcarva, line 52: Will this cause it to emit multiple results? Can we add a test to illustrate what happens when more than one tasks are missing? There's `test_no_tasks_present` but that's asserting against a dynamic value, `all_tasks_missing_errors`. Can we add a test that asserts against a literal so the intended behavior is more obvious?

zregvart: In `test_all_but_one_required_task_not_present` all but one task is missing, added `test_several_tasks_not_present` where all but three tasks are missing.

policy/release/tasks.rego
```diff
@@ -0,0 +1,59 @@
...
+	result := lib.result_helper(rego.metadata.chain(), [missing])
```
simonbaird, line 58: In other places, e.g. policy/pipeline/required_tasks.rego we've done a single deny that lists all the missing items, rather than one per item (which IIUC) we have here.

E.g. it could be like this:

    result := lib.result_helper(rego.metadata.chain(), [concat(",", all_missing)])

Not saying that's better or worse, but just noting it's different. We might want to consider the pros and cons for each approach.

policy/release/tasks.rego
simonbaird: FYI in policy/pipeline/required_tasks_test.rego we do have the duplication that is being avoided here, but it's in a different form. Consider refactoring that so we use the same approach in both places. (Not a blocker for this PR, just something to consider.)

policy/release/tasks.rego
```diff
@@ -0,0 +1,59 @@
...
+# description: |-
+#   This policy enforces that the required set of tasks is run in a
+#   PipelineRun.
+# custom:
+#   short_name: tasks_required
+#   failure_msg: Required task '%s' not found in the PipelineRun attestation
+deny[result] {
+	att := lib.pipelinerun_attestations[_]
+	attested_tasks := {t | t := att.predicate.buildConfig.tasks[_].name}
+	required_tasks := {t | t := rego.metadata.chain()[_].annotations.custom.tasks_required.rule_data.required_task_refs[_]}
```
simonbaird, line 55: Maybe it's worth adding a comment here to explain the intentions since it's not 100% explicit, e.g. "# We're expecting this to find the required_task_refs list in the package level annotations, see above" or something along those lines.

policy/release/tasks_test.rego
```diff
@@ -0,0 +1,59 @@
...
+		{
+			"code": "tasks_required",
+			"msg": "Required task 'clamav-scan' not found in the PipelineRun attestation",
+			"effective_on": "2022-01-01T00:00:00Z",
+		},
+		{
+			"code": "tasks_required",
+			"msg": "Required task 'add-sbom-and-push' not found in the PipelineRun attestation",
+			"effective_on": "2022-01-01T00:00:00Z",
+		},
```
simonbaird, line 52: Could maybe write a rule and make it tidier, e.g.:

    present := {
       mock_deny('sanity-inspect-image'),
       mock-deny('clamav-scan'),
       ...
    }

Or even something that takes a list, e.g.:

    present := mock_task_required_denies(['sanity-inspect-image','clamav-scan','add-sbom-and-push')

policy/release/tasks.rego
zregvart: Let's keep it consistent

policy/release/tasks.rego
zregvart: Found good use of `all_tasks_missing_errors` to simplify this and make it easier to grasp

zregvart:
@simonbaird please have another look

policy/release/tasks.rego
```diff
@@ -0,0 +1,65 @@
...
+#   short_name: tasks_required
+#   failure_msg: Required task(s) '%s' not found in the PipelineRun attestation
+deny[result] {
+	att := lib.pipelinerun_attestations[_]
+
+	# reported by tasks_missing above
+	count(att.predicate.buildConfig.tasks) > 0
+
+	# collects names of tasks that are present in the attestation
+	attested_tasks := {t | t := att.predicate.buildConfig.tasks[_].name}
```
lcarva, line 57: I wonder if this should be the TaskRun name (what you have now), or the Task name. A TaskRun name would make it easier for someone to fake having run a required task.

lcarva (commented):
Sorry for forgetting about this PR.

zregvart:
Includes #74, so review/merge that first.

lcarva (approved):
LGTM! It probably needs a rebase since the other commit was already merged.

PR #63, opened 2022-07-06
Add resource to track acceptable tekton bundles

lcarva:
https://issues.redhat.com/browse/HACBS-748


zregvart:
Would `/data` be a better place for these? With affordances in the Makefile not to remove this in `clean-data`?

lcarva:
> Would `/data` be a better place for these? With affordances in the Makefile not to remove this in `clean-data`?

I'm not sure. The name `/data` seems like the right thing. I do want to easily differentiate between the different sources of data. We can create subdirectories under `/data`. My concern is that it might break things I'm not aware of.

simonbaird:
I think it's okay to put it in a different place.

This PR could include something like this in Makefile to make it clear what the plan is:


    # For additional data that we're not expecting to be generated at run time
    RESOURCE_DIR=./resources

    ...
    --data $(RESOURCE_DIR) \  # ...in a few places
    ...

zregvart:
My inclination would be to keep it simple, in Rego terms there is "data", "policy" and "inputs", we should not invent a fourth thing here.

But, it seems that we're building a set of conventions here, i.e. we need a few key data points in order to evaluate, like what the Rego namespace is, where are the policy files, what to include as data in the evaluation. A separate but somehow related discussion would be to introduce a descriptor by which we could tell what the Rego namespace is, what policy files and what data files should be included when evaluating.

So for now, we can go with either `data` or `resources`, but in the future I think we need to consider if our conventions will suit everyone and either enforce them or make them prescribed using a descriptor file.

zregvart (commented):
I think we need a change to the Makefile in the `check-release` target to include this as data.

lcarva:
I moved the file from `resources` to `data` since that's already a well established convention in conftest/rego. I also modified the `Makefile` accordingly.

We talked today about committing a dummy (or empty file) under `data/config.json`. However, I'm not 100% sure if that's a good idea because we do expect the config file to be created by clients.

There is one case that might need some revisiting. @robnester-rh mentioned that some parts of the ec-cli will check for the presence of the `data` directory. If missing, it creates the dir and and empty file at `data/config.json`. It sounds like that check would have to check for the presence of the file instead. (Or maybe we need to fix the ec-cli so it doesn't have this requirement anymore.)

(I think calling conftest or opa indirectly via the `Makefile` targets doesn't actually require the config file to be present - just that the data dir exists)

We also talked about a little more about @zregvart's [comment](https://github.com/hacbs-contract/ec-policies/pull/63#issuecomment-1177267798). The overall consensus was to hold off on introducing the mentioned descriptor for now. The reception was positive but we need more usage experience (especially when other policy repos are being used) to determine if it's something we really need.

zregvart (approved):
I think we have everything here. Merge!

PR #66, opened 2022-07-14
Add parser for image references

lcarva:


lcarva:
This will be used later one when evaluating a policy with `data/acceptable_tekton_bundles.yml`.

robnester-rh (approved):

PR #67, opened 2022-07-14
git commit policy

joejstuart:
warn on non-conforming commit message

### DIFF
--- a/policy/release/commit_signature.rego
+++ b/policy/release/commit_signature.rego
@@ -1,0 +1,40 @@
+package policy.release.commit_signature
+
+import data.lib
+
+
+# METADATA
+# title: Signatures is a valid email address
+# description: |-
+#   Enterprise Contract verifies if a valid email address was used to sign a commit 
+# custom:
+#   short_name: disallowed_commit_signature_email
+#   failure_msg: Signature %s in commit %s is not a valid email address
+warn[result] {
+    signature := input.signatures[_]
+    email := split(signature, "@")
+    count(email) != 2
+    result := lib.result_helper(rego.metadata.chain(), [signature, input.body.Hash])
+}
+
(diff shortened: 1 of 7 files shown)


policy/release/commit_signature_test.rego
```diff
@@ -0,0 +1,33 @@
+package policy.release.commit_signature
+
+import data.lib
+
+mock_data(signatures) = d {
+    d := {"body": {"Hash": ["nm32n5235"]},"signatures": signatures}
```
zregvart, line 6: (nit) the structure could be simpler, something like:

```json
{
  "<commit-id1>": ["email1", "email2", ...],
  "<commit-id2>": ["email3", "email4", ...],
}
```

zregvart (approved):
Pragmatic enough for a proof of concept. Run `make fmt` to pass the formatting lint check.

policy/release/commit_signature.rego
```diff
@@ -0,0 +1,40 @@
...
+#     - redhat.com
+warn[result] {
+    signature := input.signatures[_]
+    domain := split(signature, "@")
+    not known_domains(domain[1], rego.metadata.rule().custom.rule_data.allowed_email_domains)
+    result := lib.result_helper(rego.metadata.chain(), [signature, input.body.Hash])
+}
+
+known_domains(domain, allowed_email_domains) {
+    lib.item_in_list(domain, allowed_email_domains)
```
lcarva, line 39: Might be worth making this check case insensitive.

policy/release/commit_signature_test.rego
lcarva: Just for my own understanding, why is `Hash` an array instead of a single string?

lcarva (approved):

PR #71, opened 2022-07-19
commit policy restructure

joejstuart:
package name rename; update to reflect new commit structure


policy/commit/signature.rego
```diff
@@ -1,4 +1,4 @@
-package policy.release.commit_signature
+package policy.commit.signature
```
lcarva, line 1: Why not just move this into `policy/release`? I can see an argument for a commit being a property of a release (or image).

lcarva (approved):

PR #85, opened 2022-07-20
Fix acceptable_tekton_bundles data

lcarva:
- Remove latest task bundle for demo purposes
- Change yml to json for data file
- Add top-level key on data file


lcarva:
The conversion from yaml to json, it turns out, was not required. The important thing here was adding the top-level key to the data file.

PR #88, opened 2022-07-28
Ignore hidden files in `opa test`

zregvart:
We can't have `opa test` load files from `.github` directory, the YAML
files in there will not be of the same schema, so we'll end up with:

```
$ opa test .
1 error occurred during loading: .github/workflows/pre-merge-ci.yaml: merge error
```

This adds `--ignore` to ignore all hidden directories.


simonbaird:
Would `opa test ./policy` be a solution? There is some rego in ./checks though..

simonbaird:
Either way, lgtm.

simonbaird (approved):

PR #101, opened 2022-08-03
Docs reorg including partial templates

simonbaird:
Includes:
* Separating release and pipeline rules docs
* Reusable templates for docs content
* Appearance tweaks and improvements


simonbaird:
Let's call it a draft. I'll add more commits soon.

simonbaird:
Updated preview here: https://simonbaird.github.io/ec-policies/ec-policies/index.html

docsrc/templates/warn_deny_indicator.adoc.tmpl
```diff
@@ -0,0 +1,11 @@
+{{/* Todo: Use custom classes instead of this hackery */}}
+
+{{- define "color-span" -}}
+  +++<span style="font-weight:bold; color:{{ .color }}; opacity:70%;">{{ .text }}</span>+++
```
zregvart, line 4: Rather than inline styles we should use CSS class names, like:

```suggestion
  [policy {{ .text }}]#{{ .text }}#
}}</span>+++
```

This would give a CSS selector like `.doc .policy .WARNING`.

Then this can be styled/themed as needed. We should not opacity here, also we should check for WCAG for contrast. The easiest would be to use the existing variables from the default UI, e.g. `--important-color` and `--warning-color` to keep with the existing styling and to ease downstream styling/customization. Though that might be a bit difficult as the optimized CSS already built in the UI doesn't have those variables -- so forking the UI to get that in will be required.

simonbaird: I agree, but I have zero interest in forking the UI now. Is there a way to inject some css without doing that? Maybe we could ask the docs team what their plans are.

Ps, I did have a version before that use some syntax highlighting classes, e.g. hljs-string hljs-addition to add some color. It worked, but not sure if it's better or worse than the line styles.

Anyway, I'll have another look at this and see if there's a better way to do it, or maybe I'll just take it out for now.

zregvart: We can add a `.css` file via `suplemental-ui` and in addition to that add [partials/head-styles.hbs](https://gitlab.com/antora/antora-ui-default/-/blob/master/src/partials/head-styles.hbs) with link to that css (copy-paste change filename)

zregvart: Though ^, doesn't get the variables, unless some CSS magic pulls those in...

simonbaird: Much better, thanks!

PR #106, opened 2022-08-09
Rule to assert Java dependency provenance

zregvart:
Adds a new rule to assert that the `SBOM_JAVA_COMPONENTS_COUNT` result
contains only `redhat` and `rebuilt` counts.

Ref. https://issues.redhat.com/browse/HACBS-931

### DIFF
--- a/policy/release/java.rego
+++ b/policy/release/java.rego
@@ -1,0 +1,35 @@
+package policy.release.java
+
+import data.lib
+import future.keywords.in
+
+# METADATA
+# title: Prevent Java builds from depending on foreign dependencies
+# description: |-
+#   The SBOM_JAVA_COMPONENTS_COUNT TaskResult finds dependencies that have
+#   originated from foreign repositories, i.e. ones that are not rebuilt or
+#   redhat.
+# custom:
+#   short_name: java_foreign_dependencies
+#   failure_msg: Found Java dependencies from '%s', expecting to find only from '%s'
+#   rule_data:
+#     allowed_component_sources:
+#       - redhat
+#       - rebuilt
+deny[result] {
+	results := lib.results_named(lib.java_sbom_component_count_result_name)
+
+	# convert to set
+	allowed := {a | a := rego.metadata.rule().custom.rule_data.allowed_component_sources[_]}
+
+	# contains names of dependency sources that are foreign, i.e. not one of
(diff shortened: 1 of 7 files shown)


policy/release/java.rego
```diff
@@ -0,0 +1,28 @@
...
+#   redhat.
+# custom:
+#   short_name: java_foreign_dependencies
+#   failure_msg: Found Java dependencies from '%s', expecting to find only from '%s'
+#
+deny[result] {
+	results := lib.results_named(lib.java_sbom_component_count_result_name)
+	foreign := [name |
+		results[i][name]
+		not name in (_allowed_component_sources | {lib.task_name})
```
lcarva, line 22: Should we make use of the `lib.task_name` in this rule? Or is that something to be done later on which relates to the issue you brought up in chat?

lcarva (approved):

policy/release/java.rego
```diff
@@ -0,0 +1,28 @@
+package policy.release.java
+
+import data.lib
+import future.keywords.in
+
+_allowed_component_sources := {"redhat", "rebuilt"}
```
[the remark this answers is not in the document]
zregvart: Yeah much better

policy/release/java.rego
zregvart: The `lib.results_named` adds `__task_name` key to the results, we need to add it here so we don't trip on it and consider it one of the dependency sources...

lcarva: Yes yes yes. My question was about adding some sort of assertion to the value of that attribute to ensure it came from the task we want. It's beyond the scope of this PR, I think.

lcarva (approved):

policy/release/java.rego
zregvart: We could definitely use it and I guess adding `__task_name` was a bit of forward-thinking to that regard.
Though I think we should think about this for a bit, that whole thing is a bit iffy to me (requires a spike I think). For example: do we know when the result is expected, i.e. could it be that a rouge pipeline did a `oc delete tr ...` to delete the results and this policy check passes (I think that's possible given enough permissions). Perhaps for that we could assume that such permissions would not be granted (should we have a policy to check that?); still I think we should model this a bit and see what else could throw a wrench in our assumptions.

PR #113, opened 2022-08-16
Avoid dependabot alert on glob-parent node module

simonbaird:
Not sure how, why or it this works, just copying it from here:
https://github.com/hacbs-contract/ec-cli/blob/8d7c6ec9e0b640e5a1f75aeae8488037d49c81f5/website/package.json#L14-L16

I also ran npm update to refresh the lock file.


lcarva (approved):

simonbaird:
Hopefully it doesn't introduce subtle bugs in Antora...

PR #118, opened 2022-08-25
Use opa-inspect npm package for docs build

simonbaird:
JIRA: HACBS-999


simonbaird:
To try it out:

```
npm i && npm run inspect-js-test
```

simonbaird:
As mentioned in Jira, I need some help with waiting for async code to be done. I tried a bunch of times to make it work how I want but I think I'm stuck.

inspect-js-test.js
```diff
@@ -0,0 +1,40 @@
...
+    result = JSON.parse(`{"annotations":[ ${allJson.join(",")} ]}`)
+
+    console.log("Real output is now ready...")
+
+    // This eventually shows the data we really want...
+    console.log(JSON.stringify(result.annotations.slice(0,2), null, 2))
+  })
+
+  // ..but this returns immediately so we always return null
+  // Need to await somehow but I've made so many different attempts
```
lcarva, line 32: With the existing state of `opaInspect`, if you want to create a wrapper that does something with the results, I believe you'll have to use some sort of callback:

```diff
diff --git a/inspect-js-test.js b/inspect-js-test.js
index 9fdc2ae..19977ec 100644
--- a/inspect-js-test.js
+++ b/inspect-js-test.js
@@ -11,7 +11,7 @@ function readFile(fileName) {
   return fs.readFileSync(fileName).toString()
 }
 
-function readRegoAnnotations() {
+function readRegoAnnotations(cb) {
   var result = null
 
   opaInspect(inspect => {
@@ -26,6 +26,8 @@ function readRegoAnnotations() {
 
     // This eventually shows the data we really want...
     console.log(JSON.stringify(result.annotations.slice(0,2), null, 2))
+
+    cb(result);
   })
 
   // ..but this returns immediately so we always return null
@@ -37,4 +39,6 @@ function readRegoAnnotations() {
   return result
 }
 
-console.log(`Result: ${readRegoAnnotations()}`)
+readRegoAnnotations(result => {
+    console.log(`Result: ${result}`)
+})
```

Or, maybe you can just call `opaInspect` directly with the function that you want. Any reason not to do that?

I think a better solution is probably to make `opaInspect` return a promise that you could chain on, e.g.:
```
opaInspect().then(...).then(...);
```

I don't think we should care too much that the line after `opaInspect` is executed before the data is actually read. If there's anything that relies on the result from opaInspect, use one of the approaches above.

zregvart: I've reworked the [opa-inspect-js](https://github.com/zregvart/opa-inspect-js), it now returns a Promise, so chaining should be possible.

simonbaird:
With a few tweaks, see https://github.com/zregvart/opa-inspect-js/pull/1 , I think it's working. Will leave this in draft pending review of that PR.

Makefile
```diff
@@ -103,19 +103,17 @@ ready: fmt-amend ## Amend current commit with fmt changes
...
-.PHONY: clean-annotations
-clean-annotations:
-	@rm -f $(ANNOTATIONS_JSON)
+# These two should produce identical output
+# (For debugging only; we don't need these for anything important.)
+#
+.PHONY: annotations-js
+# opa-inspect-js produces some duplicate records hence the unique_by
+annotations-js:
+	@node antora/hack/inspect-annotations.js | jq 'unique_by(.) | sort_by(.location.file, .location.row)'
```
simonbaird, line 112: FYI @zregvart re the `unique_by` here. There are dupes coming from the module, I think - maybe due to each rule including the package scoped annotations in its chain.

Makefile
```diff
@@ -103,19 +103,17 @@ ready: fmt-amend ## Amend current commit with fmt changes
...
-ANNOTATIONS_JSON=$(ANTORA_DOCS)/rule_annotations.json
-
-$(ANNOTATIONS_JSON):
-# Use jq sort_by here because otherwise the order is different every time
-	@opa inspect --annotations --format json $(POLICY_DIR) | jq -S '.annotations |= sort_by(.location.file, .location.row)' > $@
-
-.PHONY: clean-annotations
-clean-annotations:
-	@rm -f $(ANNOTATIONS_JSON)
+# These two should produce identical output
```
lcarva, line 106: Can we say "annotations-js and annotations-opa" instead of "These two" ?

simonbaird:
https://simonbaird.github.io/ec-policies/ec-policies/release_policy.html was built on this sha.

lcarva (approved):

Makefile
simonbaird: Sure.

simonbaird:
Let's merge.

PR #119, opened 2022-08-29
Include exclude rules & rego org experiments

simonbaird:
A few experimental changes. I'm trying things out.

http://file.bos.redhat.com/~sbaird/ec-policies-docs/public/ec-policies/release_policy.html is a demo of the docs.

### DIFF
--- a/policy/lib/include_exclude_rules.rego
+++ b/policy/lib/include_exclude_rules.rego
@@ -1,0 +1,50 @@
+package lib
+
+#-----------------------------------------------------------------------------
+# Used in main_denies.rego to determine if a package should be skipped.
+# (Somewhat duplicates the non_blocking_checks behavior and hopefully will
+# replace that entirely in future.)
+#
+
+package_included(package_name) {
+	any_included_in(with_wildcard(package_name), include_rules)
+	none_included_in(with_wildcard(package_name), exclude_rules)
+}
+
+package_excluded(package_name) {
+	any_included_in(with_wildcard(package_name), exclude_rules)
+}
+
+package_excluded(package_name) {
+	none_included_in(with_wildcard(package_name), include_rules)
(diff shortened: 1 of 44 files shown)


policy/lib/include_exclude_rules.rego
```diff
@@ -0,0 +1,56 @@
...
+
+package_excluded(package_name) {
+	any_included_in(with_wildcard(package_name), exclude_rules)
+}
+
+package_excluded(package_name) {
+	none_included_in(with_wildcard(package_name), include_rules)
+}
+
+with_wildcard(package_name) := {package_name, "*"}
```
lcarva, line 22: I see that this works as intended but I don't quite understand how. If we are always adding a wildcard to the set, wouldn't the matching always be true?

policy/lib/include_exclude_rules.rego
```diff
@@ -0,0 +1,56 @@
...
+include_rules := result {
+	data.config
+	result := object.get(data.config, ["policy", "include_rules"], default_include_rules)
+}
+
+# Find the exclude_rules configuration or set the default if doesn't exist
+exclude_rules := result {
+	data.config
+	result := object.get(data.config, ["policy", "exclude_rules"], default_exclude_rules)
+}
```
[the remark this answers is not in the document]
simonbaird: Thanks, I'll try it.

policy/lib/include_exclude_rules.rego
lcarva: Ah right. Looking at this with fresh eyes makes sense. I don't think it needs to be separated for clarity sake.

simonbaird:
Need to rebase this on #124 now and maybe split it up. Likely to be abandoned eventually.

PR #122, opened 2022-09-01
Add task bundle rules for pipeline package

lcarva:
```
Add task bundle rules for pipeline package

Refactored the existing task bundle rules for release package as to
minimize code duplication.
```

```
Add warning when unpinned tekton bundle is used
```

https://issues.redhat.com/browse/HACBS-984


simonbaird:
Looks good to me. Nice PR.

PR #124, opened 2022-09-02
Package docs and misc cleanup

simonbaird:
I pulled the less experimental parts out of #119 and added them here.


policy/pipeline/task_bundle.rego
```diff
@@ -1,3 +1,11 @@
+#
+# METADATA
+# description: |-
+#   HACBS expects that certain tests were run during the build.
```
lcarva, line 4: This should probably be very similar to the package description in policy/release/attestation_task_bundle.rego. Maybe something like this:

```
#   To be able to reproduce and audit builds accurately it's important
#   to know exactly what happens during the build. To do this
#   Enterprise Contract requires that all tasks are defined in a set of
#   known and trusted task bundles. This package includes rules to
#   confirm that the tasks in a Pipeline definition are defined in task
#   bundles, and that the task bundles are from the list of known
#   and trusted bundles.
```

policy/release/tasks.rego
```diff
@@ -1,4 +1,10 @@
+#
 # METADATA
+# description: |-
+#   Enterprise Contract expects that a set of tasks were included
+#   in the pipeline build for each image to be release.
```
lcarva, line 5: *released.

lcarva (approved):

PR #128, opened 2022-09-08
Version number bump for antora extension

simonbaird:
Since it's a probably a breaking change let's make it easier to see if we have the right version.


simonbaird:
I want to update the package-lock.json in the main docs repo, so I plan to merge this right away.

PR #129, opened 2022-09-08
Fixes for downstream antora build with opa-inspect

simonbaird:
The assumption that all the rego files are visible on disk when building the docs is not correct for the downstream docs build.

(It would have been wise to test out the downstream build before merging.)

- When working with a git cloned source the abspath value for each file isn't set, which makes sense because it's actually a bare git repo. Avoid throwing an error in that situation.

- Add a symlink so that the rego files become visible to antora. We define antora/docs as the start path, so anything below that isn't visible.

- Can't use file glob any more, so there's a new way to collect the rego annotations, plus some related refactoring and cleanup.


antora/ec-policies-antora-extension/index.js
```diff
@@ -229,41 +219,55 @@ const helpers = {
...
+    const content = contentAggregate.find(c => c.name == 'ec-policies')
 
-      // Find and load the acceptable bundle data
-      const rawBundlesData = helpers.dataFromAntoraFile(yaml, content, 'acceptable_tekton_bundles.yml')
-      const acceptableBundles = helpers.processBundlesData(rawBundlesData)
+    //------------------------------------------------------------------
+    // Extract annotations from the rego files
+    //
+    const regoFiles = helpers.filesMatching(content, /^policy\/.*(?!_test)\.rego$/)
+    const opaInspectPromises = regoFiles.map(f => opa.inspect(f.path, f._contents.toString()))
```
zregvart, line 229: This should work with the latest opa-inspect-js, I think:

```suggestion
    const regoFiles = helpers.filesMatching(content, /^policy\/.*(?!_test)\.rego$/).map(f => f.path)
    const opaInspectPromises = opa.inspect(regoFiles)
```

simonbaird:
Ookay let's merge.

PR #130, opened 2022-09-09
Remove obsolete hello world example rego

simonbaird:


simonbaird:
It's trivial and should be zero risk, hence I'm gonna instamerge.

PR #131, opened 2022-09-12
Provide "raw" data file in docs

simonbaird:
The idea is that the UI team could use that to expose the rules and their metadata to users via the web UI.

Do some extra work so that the package information is available in a more convenient place.


lcarva (approved):
Nice!

simonbaird:
Let's merge.

PR #135, opened 2022-09-19
Add some docs for the new rule inclusion config

simonbaird:
Preview at https://simonbaird.github.io/ec-policies/ec-policies/policy_configuration.html .

### DIFF
--- a/antora/docs/modules/ROOT/templates/policy_configuration.hbs
+++ b/antora/docs/modules/ROOT/templates/policy_configuration.hbs
@@ -1,0 +1,213 @@
+= Policy Configuration
+
+:numbered:
+
+It's possible to configure which rules should be included and which should be
+skipped when the Enterprise Contract check runs. The way this is configured
+is using fields in the policy configuration as described here.
+
+Note that it's expected that the policy configuration will be created
+automatically by the Enterprise Contract Task, based on data from the
+applicable Enterprise Contract Policy CRD, so it's likely you won't need to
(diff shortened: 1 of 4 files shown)


zregvart (approved):

simonbaird:
I forgot some parts, in particular the special syntax for tests, and the deprecation of the old field.

simonbaird:
Let's merge.

antora/docs/modules/ROOT/templates/policy_configuration.hbs
```diff
@@ -0,0 +1,213 @@
...
+create or manage this configuration manually.
+
+The Enterprise Contract Task and the Enterprise Contract Policy CRD are
+documented elsewhere. _TODO: Where?_
+
+== Configuration Source
+
+When the Enterprise Contract runs it will look for policy configuration in
+`data.config.policy`. The easiest way to populate that data is by adding a JSON
+formatted file in the `data` directory such as `data/config.json`.
```
lcarva, line 21: The name of the file is irrelevant. The `config` from `data.config.policy` comes from the top-level key in the `config.json` file, called `config`, not the file name.

Just sharing what I learned. I think it is a good idea to stick with a known file name for the sake of convention though.

simonbaird: Yeah, you're right. I could maybe explain that to avoid potential confusion.

PR #136, opened 2022-09-22
Recognize the new buildType values

lcarva:
https://issues.redhat.com/browse/HACBS-1071

### DIFF
--- a/policy/release/attestation_task_bundle_test.rego
+++ b/policy/release/attestation_task_bundle_test.rego
@@ -5,7 +5,7 @@ import data.lib
 mock_data(task) = d {
 	d := [{"predicate": {
 		"buildConfig": {"tasks": [task]},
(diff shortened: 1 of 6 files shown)


lcarva:
Marking it as a draft until we actually update Chains to use the new buildType value for the PipelineRun attestations.

zregvart:
Seeing this now makes me think if we should match via regex to disregard the version (like. `"tekton.dev/\w+/PipelineRun"` or would that be risky approach to take?

lcarva:
> Seeing this now makes me think if we should match via regex to disregard the version (like. `"tekton.dev/\w+/PipelineRun"` or would that be risky approach to take?

I think there should be a conscious decision about which build types we accept. I wouldn't bother simplifying with a regex until we have 3+ items in the list that match the pattern. Right now there's one.

zregvart (approved):

PR #137, opened 2022-09-26
Remove time-based policies and flatten release namespace

lcarva:
Two changes are introduced:

1. Logic for handling time-based rules is removed. Instead, it's expected that the client, i.e. ec-cli, performs the expected comparisons to determine if a rule is effective now.
2. All `release.*` rules are flattened into a single namespace: `release`. Because there is no longer a main.rego to aggregate all the results from the `release.*` packages, a single namespace is needed. Otherwise, the client, i.e. ec-cli, would have to know about all the sub-namespaces, and invoke the policy evaluation for each one.

There are some notable differences (awkwardness) with this approach:
1. By using a single namespace, we have to be careful about name collisions. For example, `mock_data` is a common helper function in the unit tests. For the sake of this patch, I simply renamed them to `one_mock_data`, `two_mock_data`, and so on. Obviously, this is not ideal. If going further with this approach, I suggest either better names, or a more generic helper function that can be used by different unit tests.
2. The newly added include/exclude configuration options breaks and needs to be redone. 
3. Each rule is uniquely named as to leverage conftest's [exceptions](https://www.conftest.dev/exceptions/) mechanism.
4. Docs generation breaks and needs to be adjusted.
5. `short_name`, (code), doesn't have much of a purpose. We may be able to dynamically determine this based on the rule name when using `result_helper`, which may be useful to better identify violations/warnings from the output. Oddly enough, conftest does not show the actual rule name that was violated in the output.


lcarva:
This PR has served its usefulness. We decided to go forward with moving the logic for time-based policies from ec-policies to ec-cli. However, we're not quite ready to drop main.rego right now. Instead, we'll just focus on the time-based policies for now. This work will be done in [HACBS-1220](https://issues.redhat.com/browse/HACBS-1220). The changes coming out of that story should be significantly smaller and less intrusive than what was done in this PR.

PR #139, opened 2022-09-30
Check provenance of test and java TaskRun results

zregvart:
With this the `TaskRun` results considered by the `test` and `java` policy rules need to be from a `TaskRun` loaded from via a Tekton Bundle, i.e. not defined on the cluster or inline in a `PipelineRun`, and the image reference from which the `Task` was loaded needs to be one from the list of accepted bundles (`acceptable_tekton_bundles.yml`).

The caveat with the `java` policy rules is that there is no policy denial resulting from omission of the result. This is relevant here because the omission of the result could be due to the `TaskRun` result resulting from an inline, cluster or unacceptable bundle.

Ref. https://issues.redhat.com/browse/HACBS-995

### DIFF
--- a/policy/lib/bundles_test.rego
+++ b/policy/lib/bundles_test.rego
@@ -2,6 +2,16 @@ package lib.bundles
 
 import data.lib
 
+# used as reference bundle data in tests
+bundle_data := {"registry.img/acceptable": [{
+	"digest": "sha256:digest",
+	"tag": "",
+	"effective_on": "2000-01-01T00:00:00Z",
+}]}
+
+# used as reference bundle data in tests
+acceptable_bundle_ref := "registry.img/acceptable@sha256:digest"
+
 test_disallowed_task_reference {
 	tasks := [
 		{"name": "my-task-1", "taskRef": {}},
@@ -236,3 +246,11 @@ task_bundles = {"reg.com/repo": [
 		"effective_on": "2021-01-01T00:00:00Z",
 	},
 ]}
+
+test_acceptable_bundle_is_acceptable {
+	is_acceptable(acceptable_bundle_ref) with data["task-bundles"] as bundle_data
+}
+
+test_unacceptable_bundle_is_unacceptable {
+	not is_acceptable("registry.img/unacceptable@sha256:digest") with data["task-bundles"] as bundle_data
+}
--- a/policy/release/java.rego
+++ b/policy/release/java.rego
@@ -5,10 +5,16 @@
 #   This package contains a rule to confirm that all Java dependencies
 #   were rebuilt in house rather than imported directly from potentially
 #   untrusted respositories.
+#   The result must be reported by a Task that has been loaded from an
+#   acceptable Tekton Bundle.
+#   See xref:release_policy.adoc#attestation_task_bundle_package[Task bundle checks].
+#   If the result is missing or provided via a task loaded from unacceptable no
+#   issue is reported.
 #
 package policy.release.java
 
 import data.lib
+import data.lib.bundles
 import future.keywords.in
 
 # METADATA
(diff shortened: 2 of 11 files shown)


policy/lib/bundles_test.rego
```diff
@@ -2,6 +2,16 @@ package lib.bundles
 
 import data.lib
 
+# used as reference bundle data in tests
+bundle_data := {"registry.img/accepable": [{
```
lcarva, line 6: `s/accepable/acceptable`

policy/lib/bundles_test.rego
```diff
@@ -236,3 +246,11 @@ task_bundles = {"reg.com/repo": [
 		"effective_on": "2021-01-01T00:00:00Z",
 	},
 ]}
+
+test_acceptable_bundle_is_acceptable {
+	is_acceptable(acceptable_bundle_ref) with data["task-bundles"] as bundle_data
+}
+
+test_unacceptable_bundle_is_unacceptable {
+	not is_acceptable("registry.img/unaccepable@sha256:digest") with data["task-bundles"] as bundle_data
```
lcarva, line 255: `s/unaccepable/unacceptable`

policy/release/java.rego
```diff
@@ -25,7 +31,14 @@ import future.keywords.in
 #       - redhat
 #       - rebuilt
 deny[result] {
-	results := lib.results_named(lib.java_sbom_component_count_result_name)
+	java_results := lib.results_named(lib.java_sbom_component_count_result_name)
+
+	results := [result |
+		result := java_results[_]
+		result[lib.key_bundle]
```
lcarva, line 38: This line can be omitted.

policy/release/java.rego
```diff
@@ -25,7 +31,14 @@ import future.keywords.in
...
+		bundle := result[lib.key_bundle]
+		bundles.is_acceptable(bundle)
```
lcarva, line 40: This would still allow a trusted bundle to include a task that spoofs the results, right? I think that's ok, just clarifying.

policy/release/lib/attestations.rego
```diff
@@ -47,12 +51,36 @@ results_named(name) = results {
 		result_map := json.unmarshal(result.value)
 
 		# Inject the task name so we can show it in failure messages
```
lcarva, line 53: Is this comment out of date?

policy/release/lib/attestations.rego
```diff
@@ -47,12 +51,36 @@ results_named(name) = results {
...
+# bundle
+task_data(task) = info {
+	not task.ref
+	not task.ref.kind == "Task"
+	info := {key_task_name: task.name}
+}
+
 # (Don't call it test_results since test_ means a unit test)
-results_from_tests := results_named(hacbs_test_task_result_name)
+results_from_tests = results {
```
lcarva, line 74: Is there a reason we can't just move this logic to `results_named`? Might as well make that function safer. Maybe a future PR.

policy/release/lib/attestations_test.rego
```diff
@@ -20,14 +21,49 @@ mock_tr_att_legacy := {"predicate": {"buildType": tr_build_type_legacy}}
 
 garbage_att := {"predicate": {"buildType": "garbage"}}
 
-# Used also in main_test and test_test
+# This is used throught the tests to generate an attestation of a PipelineRun
```
lcarva, line 24: `s/throught/through`

There are other occurrences in this file.

policy/release/lib/attestations_test.rego
```diff
@@ -20,14 +21,49 @@ mock_tr_att_legacy := {"predicate": {"buildType": tr_build_type_legacy}}
...
+# att_mock_helper_ref("result_name", {"value1": 1, "value2", "b"}, "task_name", "registry.io/name:tag...")
+# Make note of `bundle_data` and `acceptable_bundle_ref` in the data.lib.bundles
+# package that helps setup the acceptable bundle, for example:
+#
+# import data.lib
+# import data.lib.bundles
+# attestations := [lib.att_mock_helper_ref("RESULT_NAME", {...}, "task-name", bundles.acceptable_bundle_ref)]
+# {...} == deny
+#	with data["task-bundles"] as bundles.bundle_data
+#	with input.attestations as attestations
```
lcarva, line 55: These docs are really helpful!

policy/release/java.rego
zregvart: Yes, otherwise we'd have to link a Task bundle to a task to a result, which would be difficult with the current Task bundles (having multiple tasks within them at semi-random). Wrote [a comment](https://issues.redhat.com/browse/HACBS-995?focusedCommentId=21042997&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21042997) on this on the issue.

lcarva (approved):

PR #140, opened 2022-09-30
Required tasks policy with invocation parameters

zregvart:
This is built upon #139, so review that first.

The `sanity-label-check` Task can be parameterized with `POLICY_NAMESPACE` parameter, which can have values `required_checks` and `optional_checks`. When the task is run in the pipeline it is given then name `sanity-label-check` and `sanity-optional-label-check`. Note this is the given name for a `Task` within the `PipelineRun`. We assert the name of the referenced task from the Task Bundle, which in both cases is `sanity-label-check`.

So to distinguish between this two cases in addition to requiring a task by name, this allows to require a task by a parameter and value.

So the required task list now can include:
 - `sanity-label-check[POLICY_NAMESPACE=required_checks]`
 - `sanity-label-check[POLICY_NAMESPACE=optional_checks]`

This resolves the first part of HACBS-1166.

Ref. https://issues.redhat.com/browse/HACBS-1166

[1] https://issues.redhat.com/browse/HACBS-995

### DIFF
--- a/policy/release/tasks.rego
+++ b/policy/release/tasks.rego
@@ -5,6 +5,12 @@
 #   in the pipeline build for each image to be released.
 #   This package includes a set of rules to verify that the expected
 #   tasks ran in the pipeline when the image was built.
+#   Required tasks are listed by the names given to them within the Tekton
+#   Bundle image. Optionally invocation parameter of a Task can be also
+#   mandated by including the name and the value in square brackets following
+#   the name of the task. For example: ``name[PARAM=val]``. Only single parameter
+#   is supported, to assert multiple parameters repeat the required task
+#   definition for each parameter seperately.
 #   The Tasks must be loaded from an acceptable Tekton Bundle.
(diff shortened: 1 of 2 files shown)


policy/release/tasks.rego
```diff
@@ -64,18 +70,31 @@ deny[result] {
...
-		task_ref := att.predicate.buildConfig.tasks[_].ref
-		task_ref.kind == "Task"
-		bundle_ref := task_ref.bundle
+	attested_tasks_ary := {t |
+		task := att.predicate.buildConfig.tasks[_]
+		task.ref.kind == "Task"
+		bundle_ref := task.ref.bundle
 		bundles.is_acceptable(bundle_ref)
-		t := task_ref.name
+		t := _task_names(task)
```
lcarva, line 78: You can change this to `t := _task_names(task)[_]` to avoid the intermediate `attested_tasks_ary` variable.

lcarva (approved):

PR #168, opened 2022-10-31
Add initial check for SLSA Available requirement

lcarva:
https://issues.redhat.com/browse/HACBS-1275

### DIFF
--- a/policy/release/slsa_available.rego
+++ b/policy/release/slsa_available.rego
@@ -1,0 +1,37 @@
+#
+# METADATA
+# title: SLSA - Available
+# description: |-
+#   The SLSA Available requirement states the following:
+#
+#   "The provenance is available to the consumer in a format that the consumer accepts. The
+#   format SHOULD be in-toto SLSA Provenance, but another format MAY be used if both producer
+#   and consumer agree and it meets all the other requirements."
+#
+#   This package only accepts the in-toto SLSA Provenance format.
+#
+package policy.release.slsa_available
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+
+# METADATA
+# title: Attestation predicate type
+# description: |-
+#   The predicateType field of the attestation must indicate the in-toto SLSA Provenance format
+#   was used to attest the PipelineRun.
+# custom:
+#   short_name: unexpected_predicate_type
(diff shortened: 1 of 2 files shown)


policy/release/slsa_available.rego
```diff
@@ -0,0 +1,37 @@
...
+# custom:
+#   short_name: unexpected_predicate_type
+#   failure_msg: Attestation predicate type %q is not an expected type (%s)
+#   rule_data:
+#     allowed_predicate_types:
+#     - https://slsa.dev/provenance/v0.2
+deny contains result if {
+	some att in lib.pipelinerun_attestations
+	allowed_predicate_types := lib.rule_data(rego.metadata.rule(), "allowed_predicate_types")
+	not att.predicateType in allowed_predicate_types
```
lcarva, line 35: If the attestation does not include a `predicateType` at all, the deny rule is not triggered. I think that's ok, because we intend on verifying the attestation against a schema in https://issues.redhat.com/browse/HACBS-1276. So an attestation missing this field, should trigger a failure there. Also, the schema check is part of the same SLSA requirement, so it's likely going to be defined somewhere on this file as well.

Thoughts?

policy/release/slsa_available.rego
```diff
@@ -0,0 +1,37 @@
...
+# description: |-
+#   The SLSA Available requirement states the following:
+#
+#   "The provenance is available to the consumer in a format that the consumer accepts. The
+#   format SHOULD be in-toto SLSA Provenance, but another format MAY be used if both producer
+#   and consumer agree and it meets all the other requirements."
+#
+#   This package only accepts the in-toto SLSA Provenance format.
+#
+package policy.release.slsa_available
```
lcarva, line 13: Unclear if this package name (and/or filename) is the right format. Ideas are welcome :)

policy/release/slsa_available.rego
zregvart: Yeah, let's not over complicate the rule, this should be a concern for the schema check

policy/release/slsa_available.rego
zregvart: I think this works, the only alternative I could think of would be `policy.release.slsa.available` though I'm not sure if it makes any diffrence

zregvart (approved):

PR #171, opened 2022-11-02
Add support for different JSON types of results

lcarva:
Task results are not required to be a JSON object. This commit changes the data returned by `results_named` to better represent this.

This work is required for https://issues.redhat.com/browse/HACBS-1291

### DIFF
--- a/policy/release/lib/attestations.rego
+++ b/policy/release/lib/attestations.rego
@@ -23,6 +23,8 @@ key_task_name := "__task_name"
 
 key_bundle := "__bundle_name"
 
+key_value := "__value"
+
 # These are the ones we're interested in
 pipelinerun_attestations := [att |
 	att := input.attestations[_]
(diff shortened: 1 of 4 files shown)


policy/release/lib/attestations.rego
```diff
@@ -41,9 +43,11 @@ tasks_from_pipelinerun := [task |
 ]
 
 # All results from the attested PipelineRun with the provided name. Results are
-# expected to contain a JSON value. The JSON value will be augmented with a key
-# "__task_name" that will hold the name of the TaskRun where the named result
-# was found.
+# expected to contain a JSON value. The return object contains the following
+# keys:
```
lcarva, line 47: Since the actual result is neatly nested under its own key now, we could do away with the `__` prefix for the keys. Maybe a further enhancement is to do away with the `key_` variables as well:
```rego
{
  "task_name": "foobar",
  "bundle": "quay.io/...",
  "value": ...,
}
```

Thoughts?

zregvart (approved):

PR #174, opened 2022-11-04
Policy rule to assert SBOM was included in the built image

lcarva:
The `sbom-json-checks` task emits the standard `HACBS_TEST_OUTPUT` result. So if the SBOM is either missing or invalid, the result will indicate a failure. This should be covered by the existing policies in the `policy.release.test` package. However, we do want to make sure that this check is always performed. For this reason, the task is added to the list of required tasks.

As part of testing, I noticed that I was hitting the issue described in [HACBS-1366](https://issues.redhat.com/browse/HACBS-1366). I decided to fix that issue as well so I could be re-assured the new policy is working as expected. A unit test was added to ensure the previously breaking scenario is covered.

### DIFF
--- a/policy/release/tasks_test.rego
+++ b/policy/release/tasks_test.rego
@@ -3,6 +3,28 @@ package policy.release.tasks
 import data.lib
 import data.lib.bundles
 
+test_all_tasks_present {
+	lib.assert_empty(deny) with data["task-bundles"] as bundles.bundle_data
+		with input.attestations as [{"predicate": {
+			"buildType": lib.pipelinerun_att_build_types[0],
+			"buildConfig": {"tasks": [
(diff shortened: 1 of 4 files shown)


policy/release/tasks_test.rego
```diff
@@ -3,6 +3,29 @@ package policy.release.tasks
 import data.lib
 import data.lib.bundles
 
+test_all_tasks_present {
+	lib.assert_empty(deny) with data["task-bundles"] as bundles.bundle_data
+		with input.attestations as [{"predicate": {
```
zregvart, line 8: I think `input.attestation` can be generated via `_attestations_with_tasks(all_required_tasks)` that way we don’t need to maintain a second list of required tasks here

zregvart (approved):
Looks good, one small improvement in the comment

policy/release/tasks_test.rego
lcarva: Yes!

PR #176, opened 2022-11-07
Split data from policy rules

lcarva:
This allows someone to re-use the policies but provide a different set of data values to be used by the policy rules. For example, it is possible for someone to specify different allowed registries for base images.

At this point, this is meant for illustration purposes for the spike: https://issues.redhat.com/browse/HACBS-1329

### DIFF
--- a/policy/release/base_image_registries.rego
+++ b/policy/release/base_image_registries.rego
@@ -24,15 +24,10 @@ import data.lib.bundles
 # custom:
 #   short_name: disallowed_base_image
 #   failure_msg: Base image %q is from a disallowed registry
-#   rule_data:
-#     allowed_registry_prefixes:
-#     - registry.access.redhat.com/
-#     - registry.redhat.io/
 #
 deny contains result if {
-	allowed_registry_prefixes := lib.rule_data(rego.metadata.rule(), "allowed_registry_prefixes")
 	some image_ref in _base_images
-	not _image_ref_permitted(image_ref, allowed_registry_prefixes)
+	not _image_ref_permitted(image_ref, data.rule_data.allowed_registry_prefixes)
 	result := lib.result_helper(rego.metadata.chain(), [image_ref])
 }
 
--- a/data/rule_data.yml
+++ b/data/rule_data.yml
@@ -1,0 +1,5 @@
+---
+rule_data:
+  allowed_registry_prefixes:
+  - registry.access.redhat.com/
+  - registry.redhat.io/
--- a/policy/release/base_image_registries_test.rego
+++ b/policy/release/base_image_registries_test.rego
@@ -15,6 +15,7 @@ test_acceptable_base_images {
 		bundles.acceptable_bundle_ref,
 	)]
 	lib.assert_empty(deny) with data["task-bundles"] as bundles.bundle_data
+		with data.rule_data.allowed_registry_prefixes as ["registry.redhat.io/", "registry.access.redhat.com/"]
 		with input.attestations as attestations
 }
 
@@ -26,6 +27,7 @@ test_empty_base_images {
 		bundles.acceptable_bundle_ref,
 	)]
 	lib.assert_empty(deny) with data["task-bundles"] as bundles.bundle_data
+		with data.rule_data.allowed_registry_prefixes as ["registry.redhat.io/", "registry.access.redhat.com/"]
 		with input.attestations as attestations
 }
 
@@ -54,6 +56,7 @@ test_unacceptable_base_images {
 		},
 	}
 	lib.assert_equal(deny, expected) with data["task-bundles"] as bundles.bundle_data
+		with data.rule_data.allowed_registry_prefixes as ["registry.redhat.io/", "registry.access.redhat.com/"]
 		with input.attestations as attestations
 }
 
@@ -70,5 +73,6 @@ test_unacceptable_bundle {
 		"msg": "Base images result is missing",
 	}}
 	lib.assert_equal(deny, expected) with data["task-bundles"] as bundles.bundle_data
+		with data.rule_data.allowed_registry_prefixes as ["registry.redhat.io/", "registry.access.redhat.com/"]
 		with input.attestations as attestations
 }


policy/release/base_image_registries.rego
```diff
@@ -24,15 +24,10 @@ import data.lib.bundles
 # custom:
 #   short_name: disallowed_base_image
 #   failure_msg: Base image %q is from a disallowed registry
-#   rule_data:
-#     allowed_registry_prefixes:
-#     - registry.access.redhat.com/
-#     - registry.redhat.io/
 #
 deny contains result if {
-	allowed_registry_prefixes := lib.rule_data(rego.metadata.rule(), "allowed_registry_prefixes")
```
simonbaird, line 33: One idea:

We could add some logic to lib.rule_data so that it first looks for `data.values.allowed_registry_prefixes` and if that is not found then it falls back to the default rule_data defined in the annotations.

It mean the rule could be functional out of the box without any additional data, with a clear path to providing the custom data when it's required.

lcarva: My concern with providing default values is that it creates two paths for resolving data values. Since we control the policies repo, we'll have the motivation to always just use the default values for our purposes. This leaves room for the other mechanism to break over time without anyone actually noticing until someone other than us tries to use it.

zregvart: I think that is a valid concern and that it can be mitigated with a test that removes the annotations and relies on the data files alone. So with that I would have the default values in the annotations and a way to override them via data files.

zregvart (approved):
I think we're heading in the right direction, having the ability to easily customize the rule data is good for modularity and user experience

policy/release/base_image_registries.rego
lcarva: Fair point. We could do that with unit tests without much effort.

zregvart:
> Is this good enough to merge? Or should we have the fallback mechanism in place from the start and also make the change to the other policy rules?

Let's merge, we can easily followup on this

data/values.yml
```diff
@@ -0,0 +1,5 @@
+---
+values:
```
simonbaird, line 2: 1. what about "rule_data" as a top level key. values is a bit generic.
2. but actually, this might be the kind of thing that would lead to conftest throwing errors if it sees the key defined in two places. So we maybe need to invent a convention with a flat top level key structure like this:

```
---
rule_data_for_allowed_registry_prefixes:
- ...
```

policy/release/base_image_registries.rego
simonbaird: Yeah I'm not 100% sold on the "fallback to annotations" idea, but I guess it would make it easy to use one or the other depending on what works best for the particular rule.

Imagine a list of weekdays for example, might as well ship them with the rule definition. There's likely no need to override them in a user created data source.

simonbaird (changes requested):
I think it's better to figure out the data key clash problem in the context of the HACBS-1329 design work rather than merge early and often for this one.

data/values.yml
simonbaird: Maybe double underscore as a delimiter? `ec_rule__allowed_registry_prefixes`  or `rule_data__allowed_registry_prefixes`...

policy/release/base_image_registries.rego
lcarva: I'm in the minority here, but I think adding support for fallback adds little value. Today, we already need to provide data when evaluating the rules because some of the data we use is just too large to fit in the annotations, i.e. acceptable bundles.

By introducing a fallback mechanism, we basically have to double the amount of tests for policies that use data. I think that's unnecessary toil to support a feature we'll likely never benefit from.

data/values.yml
lcarva: `rule_data` is good name. Would it be a bit repetitive to type `data.rule_data` though? That's why I went with `values` which becomes `data.values`. I can easily change it, just double checking that you still like the name in context.

I made a quick test by creating another directory, `data2`, and modifying `make check-release` to also provide this data directory, e.g.:

```diff
diff --git a/Makefile b/Makefile
index dc0d07e..23418ee 100644
--- a/Makefile
+++ b/Makefile
@@ -200,6 +200,7 @@ check-release: ## Run policy evaluation for release
          --namespace $(RELEASE_NAMESPACE) \
          --policy $(POLICY_DIR) \
          --data $(DATA_DIR) \
+         --data $(DATA_DIR)2 \
          --no-fail \
          --output json
```

Then, `cp data/values.yml data2/`. Sure enough, conftest complains about conflicts:
```
🐚 make check-release
Error: running test: load: load documents: 1 error occurred during loading: data2/values.yml: merge error
make: *** [Makefile:199: check-release] Error 1
```

That's expected. Good, collisions are obvious. 

Next, I changed `data2/values.yml` to have a unique (nested) key:
```
🐚 diff data/values.yml data2/values.yml 
3c3
<   allowed_registry_prefixes:
---
>   allowed_registry_prefixes_test:
```
(Notice that both YAML files contain the same top-level key)

Conftest is happy to merge those two YAML files.

Does this address your second point?

There's a scenario that this may be a problem though. If a user decides to use two distinct policy sources which both use a common key but for different purposes, they can only satisfy one. I think the solution to this is that we should really have data sources mapped to a particular policy source, like you suggested in [HACBS-1329](https://issues.redhat.com/browse/HACBS-1329). Wdyt?

simonbaird: Oooh, so it does do a "deep" merge? That's interesting. I wasn't sure if it would do that.

policy/release/base_image_registries.rego
simonbaird: To be clear, I'm in favor of ditching the fallback idea of "look here, if not found look there", especially since the rego.metadata.chain() always needs to be inside the rule, which I guess means it would always have to be there in the function call.

Actually I suppose individual rules could define their own fallback mechanism if it was considered useful, but I think it's best not to make that the convention for everything.

data/values.yml
simonbaird: If conftest is smart enough to merge the keys under the top level key, that does remove my concern, and the motivation for suggesting a flat unique top level key.

I still don't like "values" as a key name though FWIW. I'd say "rule_data" is slightly preferable - even though it's only marginally less generic sounding, we are using it already in the annotations.

lcarva: Changed it to `rule_data`.

lcarva:
@simonbaird, I think your comments have been addressed. ptal.

PR #179, opened 2022-11-10
Add cache-dependency-path for setup-node

zregvart:
Seems that setup-node now looks at the lock file in the workspace root and `cache-dependency-path` needs to be provided for it to find the extension's lock file.


zregvart:
See https://github.com/hacbs-contract/ec-policies/actions/runs/3435622609

PR #187, opened 2022-11-17
Add SLSA - Source - Version Controlled policy

lcarva:
https://issues.redhat.com/browse/HACBS-1278

### DIFF
--- a/policy/release/slsa_source_version_controlled.rego
+++ b/policy/release/slsa_source_version_controlled.rego
@@ -1,0 +1,85 @@
+#
+# METADATA
+# title: SLSA - Source - Version Controlled
+# description: |-
+#   The SLSA requirement states the following:
+#
+#   "Every change to the source is tracked in a version control system
+#   that meets the following requirements:
+#
+#     [Change history] There exists a record of the history of changes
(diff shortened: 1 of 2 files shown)


policy/release/slsa_source_version_controlled.rego
```diff
@@ -0,0 +1,85 @@
...
+# description: |-
+#   Each entry in the predicate.materials array of the attestation uses
+#   a git URI.
+# custom:
+#   short_name: material_non_git_uri
+#   failure_msg: Material URI %q is not a git URI
+#
+deny contains result if {
+	some material in materials
+	not startswith(material.uri, "git+")
```
zregvart, line 61: I think we could support different syntax here, though probably we could hold off on that till someone tells us what other syntax exists. Then we could have this as rule data...

lcarva: Yeah. We should hold off on that until there's a requirement. FWIW, chains always adds the `git+` prefix: https://github.com/tektoncd/chains/blob/24dadcad3db7ed7ec82ad1bb25a34e9e38845c1f/pkg/chains/formats/intotoite6/attest/attest.go#L81-L87

PR #192, opened 2022-11-22
Validate collection consistency

zregvart:
This adds a convention check to make sure that the defined collections are consistent, i.e. they do not reference non-existent rules.

Ref. https://issues.redhat.com/browse/HACBS-1422


lcarva (approved):
Very nice!

PR #194, opened 2022-11-23
Pull lib from a different repository

lcarva:


lcarva:
The lib seems to be loaded properly, and `make check-release` works properly. However, I get some odd errors regarding missing test coverage.

This feels a bit awkward and I'm not sure if it's something we need right now. 

It would be much better, of course, if there was built-in support in rego for sharing policies as dependencies, but it's not quite there (and may not get there for a while, see stagnated issue [open-policy-agent/opa#3371](https://www.github.com/open-policy-agent/opa/issues/3371) )

Thoughts?

zregvart:
> Thoughts?

Not a fan of git submodules, though I think it can be made to work. What I think we want (in the long run) is a ec-sdk of sorts, i.e. a programming model and tooling around it. Perhaps the ec-sdk is best embedded in the ec-cli and from that point onward we use ec-cli in place of opa or conftest, with all of that would entail. An alternative might be to explore loading shared libraries into opa and conftest and ship the ec-sdk as one.

lcarva:
> Not a fan of git submodules, though I think it can be made to work. What I think we want (in the long run) is a ec-sdk of sorts, i.e. a programming model and tooling around it. Perhaps the ec-sdk is best embedded in the ec-cli and from that point onward we use ec-cli in place of opa or conftest, with all of that would entail. An alternative might be to explore loading shared libraries into opa and conftest and ship the ec-sdk as one.

Good points here. I think you capture the main problem: neither opa nor conftest support external libraries. Although creating an sdk-like project would help address this, I'm hesitant to do so _right now_ because:

1. We don't actually have a business requirement to do this. 
2. Users don't actually _need_ the libraries in this repo to create policies to work with the Enterprise Contract. We don't really have users yet so whether this would be helpful to users is not yet clear.
3. Splitting the code between two repos creates maintenance and development overhead. IMO this will hinder our ability to deliver.

I'll close this POC PR based on this. In the future, if we do have a requirement to create a shared library and there's still no answer from the opa community, the sdk approach should be investigated further.

PR #201, opened 2022-12-01
Use image.parse instead of regexp

zregvart:


lcarva (approved):
Nice! Thanks for doing this so fast.

PR #203, opened 2022-12-01
`image.equal_ref` function

zregvart:
Adds a function to determine if two container image references are equal. The function succeeds if the two supplied image references are string-equal, or if they're equal ignoring the tags provided that any of the image references contains the digest.


policy/lib/image.rego
```diff
@@ -28,6 +28,33 @@ parse(ref) = d {
...
+
+equal_ref(ref1, ref2) {
+	img1 := parse(ref1)
+	img2 := parse(ref2)
+
+	img1.digest != ""
+	object.remove(img1, ["tag"]) == object.remove(img2, ["tag"])
+}
+
+equal_ref(ref1, ref2) {
```
lcarva, line 50: I don't see why this block is needed. I commented it out and the tests still pass but code coverage is still 100% :thinking: 

I think it might be impossible for line 55 to ever evaluate to true. The evaluations on line 39 and line 47 capture all the possible scenarios.

lcarva (approved):
LGTM!

PR #205, opened 2022-12-05
Use time-based required tasks list in release

lcarva:
https://issues.redhat.com/browse/HACBS-1238

### DIFF
--- a/policy/release/tasks.rego
+++ b/policy/release/tasks.rego
@@ -13,17 +13,6 @@
 #   definition for each parameter seperately.
 #   The Tasks must be loaded from an acceptable Tekton Bundle.
 #   See xref:release_policy.adoc#attestation_task_bundle_package[Task bundle checks].
-# custom:
-#   tasks_required:
-#     rule_data:
-#       required_task_refs:
-#       - clamav-scan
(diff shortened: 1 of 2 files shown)


lcarva:
Still a couple things needed to finalize https://issues.redhat.com/browse/HACBS-1238:

1. Docs reference to time-based required tasks list.
2. Apply the same logic for pipeline definition validation.

But I think this PR is self-contained, and could be merged once reviewed.

policy/release/tasks.rego
```diff
@@ -100,3 +97,26 @@ _task_names(task, raw_name) = names if {
 
 	names := {name} | params
 }
+
+# The latest set of required tasks. Tasks here are not required right now
+# but will be required in the future.
+_latest_required_tasks contains task if {
+	some task in data["required-tasks"][0].tasks
```
zregvart, line 104: Should we sort in case `data["required-tasks"]` is not sorted by date?

lcarva: Good question. So far we don't do any sorting because we rely on ec-cli to do the right thing. That's a good idea though. I'll add it to a follow up PR and file a story to do the same for acceptable bundles.

PR #208, opened 2022-12-06
Share code for time-based required tasks

lcarva:

### DIFF
--- a/policy/lib/tekton/task_test.rego
+++ b/policy/lib/tekton/task_test.rego
@@ -1,0 +1,134 @@
+package lib.tkn
+
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+import data.lib.bundles
+
+test_latest_required_tasks if {
+	expected := {t | some t in _expected_latest.tasks}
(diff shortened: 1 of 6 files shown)


policy/lib/tekton/task_test.rego
```diff
@@ -0,0 +1,134 @@
...
+
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+import data.lib.bundles
+
+test_latest_required_tasks if {
+	expected := {t | some t in _expected_latest.tasks}
+	lib.assert_equal(expected, latest_required_tasks) with data["required-tasks"] as _time_based_required_tasks
```
zregvart, line 11: Do we need something like:
```suggestion
	lib.assert_equal(expected, latest_required_tasks) with data["required-tasks"] as _time_based_required_tasks with data["config"] as {"policy": {"when_ns": "2022-11-01T00:00:00Z"}}
```
To prevent time travel?

zregvart (approved):
LGTM

policy/lib/tekton/task_test.rego
lcarva: We're good until 2099 at which point we'll be retired and likely dead. This will be my legacy for the future generations.

PR #209, opened 2022-12-06
Update opa and conftest

lcarva:


policy/lib/bundles.rego
```diff
@@ -126,9 +126,7 @@ _stream(tag) = stream {
 	regex.match(_stream_regex, tag)
 	parts := split(tag, "-")
 	stream := parts[count(parts) - 1]
-} else = "default" {
-	true
-}
+} else = "default"
```
lcarva, line 129: I'm glad this is the new format.

zregvart (approved):

PR #210, opened 2022-12-07
Add hermetic build task policy

lcarva:
https://issues.redhat.com/browse/HACBS-1465

Also move some relevant helpers to `lib.tkn`.


zregvart (approved):
Looks good!

PR #216, opened 2022-12-14
Use default data in tests

lcarva:


simonbaird (approved):
LGTM, nice patch! I can't see any reason why not..

PR #218, opened 2022-12-14
deny remote Dockerfile

joejstuart:
deny remote url for Dockerfile

### DIFF
--- a/policy/release/buildah_build_task.rego
+++ b/policy/release/buildah_build_task.rego
@@ -1,6 +1,6 @@
 #
 # METADATA
-# title: Builah build task
+# title: Buildah build task
 # description: |-
 #   This package is responsible for verifying the buildah build task
 #
--- a/policy/release/buildah_build_task.rego
+++ b/policy/release/buildah_build_task.rego
@@ -1,6 +1,6 @@
 #
 # METADATA
-# title: builah_build_task
+# title: Builah build task
 # description: |-
 #   This package is responsible for verifying the buildah build task
 #
@@ -13,46 +13,40 @@ import future.keywords.in
 import data.lib
 
 # METADATA
-# title: dockerfile_param_not_included
+# title: Dockerfile param not included
 # description: |-
 #   This policy verifies that there is a dockerfile parameter
 # custom:
 #   short_name: dockerfile_param_not_included
 #   failure_msg: DOCKERFILE param is not included in the task
 deny contains result if {
-	buildah_task
 	not dockerfile_param
 	result := lib.result_helper(rego.metadata.chain(), [])
 }
 
 # METADATA
-# title: dockerfile_param_external_source
+# title: Dockerfile param external source
 # description: |-
 #   This policy verifies that the dockerfile is not an external source
 # custom:
 #   short_name: dockerfile_param_external_source
 #   failure_msg: DOCKERFILE param value (%s) is an external source
 deny contains result if {
-	buildah_task
 	dockerfile_param
 	_not_allowed_prefix(dockerfile_param)
 	result := lib.result_helper(rego.metadata.chain(), [dockerfile_param])
 }
 
 _not_allowed_prefix(search) if {
-	not_allowed := ["http", "https"]
+	not_allowed := ["http://", "https://"]
 	startswith(search, not_allowed[_])
 }
 
-params(param) if {
-
-}
 buildah_task := task if {
 	some task in lib.tasks_from_pipelinerun
 	task.name == "buildah"
-} else := false
+}
 
 dockerfile_param := param if {
-	buildah_task
 	param := lib.tkn.task_param(buildah_task, "DOCKERFILE")
-} else := false
+}
--- a/policy/release/buildah_build_task.rego
+++ b/policy/release/buildah_build_task.rego
@@ -1,0 +1,58 @@
+#
+# METADATA
+# title: builah_build_task
+# description: |-
+#   This package is responsible for verifying the buildah build task
+#
+package policy.release.buildah_build_task
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+
+# METADATA
+# title: dockerfile_param_not_included
+# description: |-
+#   This policy verifies that there is a dockerfile parameter
+# custom:
+#   short_name: dockerfile_param_not_included
+#   failure_msg: DOCKERFILE param is not included in the task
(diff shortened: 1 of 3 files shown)


policy/release/buildah_build_task.rego
```diff
@@ -0,0 +1,58 @@
...
+package policy.release.buildah_build_task
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+
+# METADATA
+# title: dockerfile_param_not_included
```
lcarva, line 16: Usually, the `title` contains spaces instead of a single word.

policy/release/buildah_build_task.rego
```diff
@@ -0,0 +1,58 @@
...
+#   failure_msg: DOCKERFILE param value (%s) is an external source
+deny contains result if {
+	buildah_task
+	dockerfile_param
+	_not_allowed_prefix(dockerfile_param)
+	result := lib.result_helper(rego.metadata.chain(), [dockerfile_param])
+}
+
+_not_allowed_prefix(search) if {
+	not_allowed := ["http", "https"]
```
lcarva, line 43: These should be `["http://", "https://"]`. Someone could have a path like `httpDockerfile` in their repo and that should be totally fine.

policy/release/buildah_build_task.rego
```diff
@@ -0,0 +1,58 @@
...
+	startswith(search, not_allowed[_])
+}
+
+params(param) if {
```
lcarva, line 47: Leftover?

policy/release/buildah_build_task.rego
```diff
@@ -0,0 +1,58 @@
...
+
+}
+buildah_task := task if {
+	some task in lib.tasks_from_pipelinerun
+	task.name == "buildah"
+} else := false
```
lcarva, line 53: I think `else := false` here is irrelevant. Let `buildah_task` not produce a value if the buildah task is not found.

Same for `dockerfile_param` below.

policy/release/buildah_build_task.rego
```diff
@@ -0,0 +1,58 @@
...
+
+# METADATA
+# title: dockerfile_param_not_included
+# description: |-
+#   This policy verifies that there is a dockerfile parameter
+# custom:
+#   short_name: dockerfile_param_not_included
+#   failure_msg: DOCKERFILE param is not included in the task
+deny contains result if {
+	buildah_task
```
lcarva, line 23: Since `dockerfile_param` will resolve `buildah_task`, this line can be removed.

Same for the other `deny` rule below.

policy/release/buildah_build_task.rego
```diff
@@ -0,0 +1,58 @@
...
+	not_allowed := ["http", "https"]
+	startswith(search, not_allowed[_])
+}
+
+params(param) if {
+
+}
+buildah_task := task if {
+	some task in lib.tasks_from_pipelinerun
+	task.name == "buildah"
```
lcarva, line 52: Important: we don't want to match this name. We want to match the name of the task reference, e.g.:
```json
{
  "name": "ignore-this",
  "ref": {"name": "buildah", "kind": "Task", "bundle": ... }
}
```

There are some utilities in `policy/lib/tekton/task.rego` that should be helpful here.

policy/release/buildah_build_task.rego
```diff
@@ -0,0 +1,58 @@
...
+} else := false
+
+dockerfile_param := param if {
+	buildah_task
```
lcarva, line 56: This line can be removed. Just let it be resolved on the line below when it's used.

lcarva (changes requested):

policy/release/buildah_build_task.rego
```diff
@@ -0,0 +1,58 @@
...
+# title: dockerfile_param_external_source
+# description: |-
+#   This policy verifies that the dockerfile is not an external source
+# custom:
+#   short_name: dockerfile_param_external_source
+#   failure_msg: DOCKERFILE param value (%s) is an external source
+deny contains result if {
+	buildah_task
+	dockerfile_param
+	_not_allowed_prefix(dockerfile_param)
```
lcarva, line 38: Nit: I'm not sure if a helper method is actually very helpful here. I think this rule can be rewritten more succinctly like this:

```rego
deny contains result if {                                                 
    disallowed := ["http://", "https://"]                                 
    startswith(dockerfile_param, disallowed[_])                           
    result := lib.result_helper(rego.metadata.chain(), [dockerfile_param])
}                                                                         

```

joejstuart: I met you half way. I kind of like the method name.

policy/release/buildah_build_task.rego
joejstuart: Fixed. PTAL when you can.

policy/release/buildah_build_task_test.rego
```diff
@@ -0,0 +1,49 @@
...
+		"msg": "DOCKERFILE param is not included in the task",
+	}}
+	lib.assert_equal(expected, deny) with input.attestations as [_attestation({})]
+}
+
+_attestation(params) = attestation if {
+	attestation := {"predicate": {
+		"buildType": lib.pipelinerun_att_build_types[0],
+		"buildConfig": {"tasks": [{
+			"name": "buildah",
```
lcarva, line 44: Nit: It would be great to name this something other than "buildah", so it becomes obvious if we do break the implementation.

lcarva (approved):

policy/release/buildah_build_task.rego
```diff
@@ -0,0 +1,52 @@
+#
+# METADATA
+# title: Builah build task
```
simonbaird, line 3: Typo: Buildah

simonbaird (approved):

policy/release/buildah_build_task_test.rego
joejstuart: Something like this? https://github.com/hacbs-contract/ec-policies/pull/218/commits/7f4534201ff63bca30bfbf29e72cfced840f0663

lcarva: Closer. But because `_attestation` is always called with `'buildah'` as the parameter, it still produces something that has the same value for `.name` and `.ref.name`. Just change the `.name` to something like `"ignored"` (the line I commented on)

PR #232, opened 2023-01-04
Ignore buildah policy for non pipeline run att

lcarva:
The policy rule dockerfile_param_not_included is only meant to be executed when there's a pipeline run attestation.

### DIFF
--- a/policy/release/buildah_build_task.rego
+++ b/policy/release/buildah_build_task.rego
@@ -20,6 +20,8 @@ import data.lib
 #   short_name: dockerfile_param_not_included
 #   failure_msg: DOCKERFILE param is not included in the task
 deny contains result if {
+	# Skip this rule if the buildah task is not present
+	buildah_task
 	not dockerfile_param
 	result := lib.result_helper(rego.metadata.chain(), [])
 }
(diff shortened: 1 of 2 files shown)


simonbaird:
I think it will work as expected, but for clarity of intent, I'm thinking maybe we could define another rule that fails if there is no buildah task found. That would also fai if there were no pipelinerun_attestations but the message would be more understandable.

lcarva:
> 

buildah is not always the build task used, e.g. java, node. We have [slsa_build_scripted_build](https://github.com/hacbs-contract/ec-policies/blob/main/policy/release/slsa_build_scripted_build.rego) to catch if a build task is not present. Is that sufficient?

policy/release/buildah_build_task.rego
```diff
@@ -20,6 +20,7 @@ import data.lib
 #   short_name: dockerfile_param_not_included
 #   failure_msg: DOCKERFILE param is not included in the task
 deny contains result if {
+	count(lib.pipelinerun_attestations) > 0
 	not dockerfile_param
```
simonbaird, line 24:
```suggestion
	buildah_task # there is a buildah task
	not dockerfile_param # ...with no dockerfile param
```

simonbaird: Does it make sense? IIUC that will pass and fail in exactly the same way, but the intent is more clear.

lcarva: It does. Updated. Thanks!

simonbaird (approved):

PR #237, opened 2023-01-06
Creating metadata files for artifact hub

joejstuart:
This adds the repo file for ownership verification and the pkg files for each package. The packages are split between release and pipeline.

### DIFF
--- a/policy/pipeline/artifacthub-pkg.yml
+++ b/policy/pipeline/artifacthub-pkg.yml
@@ -1,15 +1,14 @@
 # from https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml
 version: 0.0.1
 name: OPA Pipeline Policies
-displayName: OPA pipeline Policies
+displayName: OPA Pipeline Policies
 createdAt: 2023-01-05T21:27:55.549418+00:00
 description: OPA policies enforcing standards on a tekton pipeline definition
 readme: |
-  OPA policies enforcing standards on a tekton pipeline definition
+  OPA policies enforcing standards on a tekton pipeline definition. These policies
+  are a work in progress and not yet production ready.
 install: |
-  `conftest pull git::https://github.com/hacbs-contract/ec-policies.git//policy`
-  Configure your org-specific `--data` as necessary from [data/](https://github.com/hacbs-contract/ec-policies/tree/main/data)
-  Run your conftest command: `conftest verify --data data/
+  `conftest pull oci::https://quay.io/hacbs-contract/ec-pipeline-policy:latest`
 homeURL: https://hacbs-contract.github.io/ec-policies/
 keywords:
   - opa
--- a/policy/pipeline/artifacthub-pkg.yml
+++ b/policy/pipeline/artifacthub-pkg.yml
@@ -1,0 +1,19 @@
+# from https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml
+version: 0.0.1
+name: OPA Pipeline Policies
+displayName: OPA pipeline Policies
+createdAt: 2023-01-05T21:27:55.549418+00:00
+description: OPA policies enforcing standards on a tekton pipeline definition
+readme: |
+  OPA policies enforcing standards on a tekton pipeline definition
+install: |
+  `conftest pull git::https://github.com/hacbs-contract/ec-policies.git//policy`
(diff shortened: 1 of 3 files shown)


policy/pipeline/artifacthub-pkg.yml
```diff
@@ -0,0 +1,19 @@
+# from https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml
+version: 0.0.1
+name: OPA Pipeline Policies
+displayName: OPA pipeline Policies
+createdAt: 2023-01-05T21:27:55.549418+00:00
+description: OPA policies enforcing standards on a tekton pipeline definition
+readme: |
+  OPA policies enforcing standards on a tekton pipeline definition
+install: |
+  `conftest pull git::https://github.com/hacbs-contract/ec-policies.git//policy`
```
simonbaird, line 10: I think we want to use an oci:: url here, since that's the actual artifact.

policy/pipeline/artifacthub-pkg.yml
```diff
@@ -0,0 +1,19 @@
+# from https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml
+version: 0.0.1
+name: OPA Pipeline Policies
+displayName: OPA pipeline Policies
```
simonbaird, line 4:
```suggestion
displayName: OPA Pipeline Policies
```

policy/pipeline/artifacthub-pkg.yml
```diff
@@ -0,0 +1,19 @@
...
+createdAt: 2023-01-05T21:27:55.549418+00:00
+description: OPA policies enforcing standards on a tekton pipeline definition
+readme: |
+  OPA policies enforcing standards on a tekton pipeline definition
```
simonbaird, line 8: Can we add a sentence and mention stone soup, with maybe a disclaimer to say it's perhaps not prod ready.

simonbaird: It could also include the url to our docs page and/or or our github org.

policy/pipeline/artifacthub-pkg.yml
```diff
@@ -0,0 +1,19 @@
...
+install: |
+  `conftest pull git::https://github.com/hacbs-contract/ec-policies.git//policy`
+  Configure your org-specific `--data` as necessary from [data/](https://github.com/hacbs-contract/ec-policies/tree/main/data)
```
simonbaird, line 11: Do we have any org specific data for pipeline validation?

policy/pipeline/artifacthub-pkg.yml
simonbaird: Also I think the git url should end in `//policy/pipeline` if we do want to use a git url.

policy/pipeline/artifacthub-pkg.yml
```diff
@@ -0,0 +1,19 @@
...
+  Run your conftest command: `conftest verify --data data/
```
simonbaird, line 12: Looks like a back tick missing at the end of this line.

simonbaird (changes requested):
See comments above.

policy/pipeline/artifacthub-pkg.yml
joejstuart: I can add something about stone soup. On line #17 the homeUrl is defined. Is that enough?

policy/pipeline/artifacthub-pkg.yml
joejstuart: I guess i don't see any. I'll remove.

policy/pipeline/artifacthub-pkg.yml
joejstuart: Actually, is it a good idea to mention stone soup? I mean it could change pretty quick and the name isn't really anywhere on github

policy/pipeline/artifacthub-pkg.yml
joejstuart: you're right, I did update it to use the oci url though

policy/pipeline/artifacthub-pkg.yml
simonbaird: I think the boat has sailed, i.e. look at https://github.com/hacbs-contract/ for example.

simonbaird (approved):
Lgtm.

policy/pipeline/artifacthub-pkg.yml
simonbaird: But yeah, I think the url is good enough.

PR #241, opened 2023-01-09
Remove include exclude logic

lcarva:
This is now handled by the ec-cli.

https://issues.redhat.com/browse/HACBS-1303


simonbaird (approved):

antora/docs/modules/ROOT/templates/policy_configuration.hbs
```diff
@@ -189,16 +189,28 @@ The predefined collections are:
 |*Name*
 |*Description*
 
-{{#each collectionsData.rule_collections}}
-|`{{@key}}`
-|{{description}}
+| `slsa_1`
```
zregvart, line 192: A nice followup could be to have this text on the slsa rules (packages?) and have it pulled in as data into the template so it's up to date. We could have "empty" `slsa_N` packages with just the docs in the annotations perhaps...

zregvart (approved):

PR #244, opened 2023-01-10
Policy bundle docs, plus doc related fixes

simonbaird:


antora/ec-policies-antora-extension/package.json
```diff
@@ -16,7 +16,8 @@
   },
   "homepage": "https://github.com/hacbs-contract/ec-policies/tree/main/antora/ec-policies-antora-extension#readme",
   "dependencies": {
-    "@zregvart/opa-inspect": "latest"
+    "@zregvart/opa-inspect": "latest",
+    "glob": "^8.0.3"
```
zregvart, line 20: Shouldn’t we be using the [require helper](https://docs.antora.org/antora/latest/extend/extension-helpers/#require)? Doesn’t `glob` come from Antora? Something in the back of my head tells me that we should not add dependencies to the extensions as it’ll break the loading of the extension when we using relative path in `extensions.require` of the playbook.

zregvart (commented):
Not sure about adding dependencies in the extension.

zregvart:
Let's try to fix (if) any issues in a followup PR.

antora/ec-policies-antora-extension/package.json
simonbaird: Not sure, but somehow I needed this to make the build work locally.

PR #246, opened 2023-01-11
Add WARNING test result

arewm:
[ADR-0008](https://github.com/redhat-appstudio/book/pull/24) defines five different test result status types -- `SUCCESS`, `FAILURE`, `WARNING`, `SKIPPED` and `ERROR`. There was no policy defined for handling the `WARNING` result type.

Signed-off-by: arewm <arewm@users.noreply.github.com>

PR #247, opened 2023-01-11
Rule data changes

simonbaird:
[HACBS-1575](https://issues.redhat.com/browse/HACBS-1575) but also includes some other changes, in particular moving some defaults out of the rule_data.yml file.

### DIFF
--- a/antora/docs/modules/ROOT/templates/policy_configuration.hbs
+++ b/antora/docs/modules/ROOT/templates/policy_configuration.hbs
@@ -278,9 +278,23 @@ as a data source, for example:
 
 NOTE: If the data source contains policy rules, those will be ignored.
 
-NOTE: The data source must provide the full set of required data values. These
-are all the values defined in
+NOTE: If you replace the default data source entirely, you must provide the full set of required data values.
+These are all the values defined in
 link:https://github.com/hacbs-contract/ec-policies/blob/main/data/rule_data.yml[data/rule_data.yml]
 and
 link:https://github.com/hacbs-contract/ec-policies/blob/main/data/acceptable_tekton_bundles.yml[data/acceptable_tekton_bundles.yml]
-.
+
+NOTE: It's also possible to add an additional data source containing rule data
+defined under the `rule_data_custom` top level key. Data under this key will
+take precedence over data defined under the `rule_data` top level key. So, for
+example you could use the default data source and still define your own values
+for `allowed_registry_prefixes` by adding a second data source that includes
+a file such as:
+
+[source,yaml]
+----
+rule_data_custom:
+  allowed_registry_prefixes:
+    - trusted-registry.io/trusted-images/
+    - docker.io/acme-company/
+----
--- a/policy/lib/rule_data.rego
+++ b/policy/lib/rule_data.rego
@@ -22,6 +22,12 @@ rule_data_defaults := {
 		"SKIPPED",
 		"WARNING",
 	],
+	"failed_tests_results": [
+		"FAILURE",
+		"ERROR",
+	],
+	"skipped_tests_results": ["SKIPPED"],
+	"warned_tests_results": ["WARNING"],
 }
 
 # Returns the "first found" of the following:
--- a/policy/lib/rule_data.rego
+++ b/policy/lib/rule_data.rego
@@ -1,0 +1,23 @@
+package lib
+
+# Will define built-in default values here.
+# They can be overridden if required.
+#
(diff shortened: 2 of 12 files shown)


policy/lib/rule_data.rego
```diff
@@ -0,0 +1,27 @@
...
+# And falls back to an empty list if the key is not found anywhere.
+#
+rule_data(key_name) := value {
+	value := data.rule_data_custom[key_name]
+} else := value {
+	value := data.rule_data[key_name]
+} else := value {
+	value := rule_data_defaults[key_name]
+} else := value {
+	value := []
```
simonbaird, line 37: I wonder if there's an idiomatic good way to throw an error here.

Returning a null would be another alternative.

policy/lib/rule_data.rego
```diff
@@ -0,0 +1,38 @@
...
+	"_test_key_3": 10,
+}
+
+#
+# Returns the "first found" of the following:
+#   data.rule_data_custom[key_name]
+#   data.rule_data[key_name]
+#   rule_data_defaults[key_name]
+#
+# And falls back to an empty list if the key is not found anywhere.
```
lcarva, line 28: Is an empty list the right value? I suppose if we do reach the default value, something is up. Is there a `panic` in rego? :thinking:

policy/lib/rule_data.rego
simonbaird: Not sure what happens if I just removed line 25 and 26 entirely, but I'm worried the answer is rules start being silently skipped.

policy/lib/rule_data.rego
lcarva: All of the rule data we got are lists so a list is likely the right choice here.

policy/lib/rule_data.rego
```diff
@@ -0,0 +1,38 @@
...
+	"known_attestation_types": ["https://in-toto.io/Statement/v0.1"],
+	#
+	# Used in release/slsa_provenance_available
+	"allowed_predicate_types": ["https://slsa.dev/provenance/v0.2"],
+	#
+	# Used in release/slsa_build_build_service
+	"allowed_builder_ids": ["https://tekton.dev/chains/v2"],
+	#
+	# For the tests only. Not sure how else to get to 100% coverage.
+	"_test_key_3": 10,
```
lcarva, line 19: You could have a test that uses `rule_data_defaults` as is, and on that overwrites it by using `with`.

policy/lib/rule_data.rego
lcarva: I think you're right because that would mean rule_data produces no result.

policy/lib/rule_data.rego
simonbaird: The argument for a null is that it will probably be more likely to throw an error if you typoed the key for example.

zregvart (approved):
What a let down, try something weirder next time

simonbaird:
I think the `data.rule_data_custom` convention needs to be documented somewhere/somehow before this is out of draft.

zregvart:
Take note of #246 as well...

policy/lib/rule_data.rego
simonbaird: Let me try...

simonbaird:
> Take note of #246 as well...

Rebased, should be fine now.

lcarva (approved):
LGTM! Any reason to keep it in draft and not merge it?

simonbaird:
Got one more commit coming with a small docs addition.

antora/docs/modules/ROOT/templates/policy_configuration.hbs
```diff
@@ -278,9 +278,23 @@ as a data source, for example:
 
 NOTE: If the data source contains policy rules, those will be ignored.
 
-NOTE: The data source must provide the full set of required data values. These
-are all the values defined in
+NOTE: If you replace the default data source entirely you must provide the full set of required data values.
```
lcarva, line 281: Should there be a comma? `entirely, you`

lcarva (approved):

policy/lib/rule_data.rego
simonbaird: Works good. Thanks. I forgot you can use with for other stuff.

antora/docs/modules/ROOT/templates/policy_configuration.hbs
simonbaird: Done.

PR #248, opened 2023-01-11
Upload code coverage to codecov.io

zregvart:
Adds a Make target `coverage-report` that generates `coverage.json` in SimpleCov format using `hack/simplecov.rego` sourced from[1] and uploads it in pre-merge-ci workflow to codecov.io.

[1] https://github.com/open-policy-agent/contrib/tree/main/simplecov

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,10 @@ test: ## Run all tests in verbose mode and check coverage
 coverage: ## Show which lines of rego are not covered by tests
 	@opa test $(TEST_FILES) --coverage --format json | jq -r '.files | to_entries | map("\(.key): Uncovered:\(.value.not_covered)") | .[]' | grep -v "Uncovered:null" | cat
 
+.PHONY: coverage-report ## Generates coverage.json in SimpleCov format
+coverage-report: ## Show which lines of rego are not covered by tests
+	@opa test $(TEST_FILES) --coverage --format json | opa eval --format pretty --stdin-input --data hack/simplecov.rego data.simplecov.from_opa > coverage.json
+
 .PHONY: quiet-test
 quiet-test: ## Run all tests in quiet mode and check coverage
 	@opa test $(TEST_FILES)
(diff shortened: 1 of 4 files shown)


zregvart:
Report can be viewed [here](https://app.codecov.io/github/hacbs-contract/ec-policies/tree/pr%2Fcodecov)

hack/simplecov.rego
```diff
@@ -0,0 +1,52 @@
...
+
+not_covered_map(report) = ncm if {
+	not_covered := object.get(report, "not_covered", [])
+	ncm := {line: 0 |
+		some item in not_covered
+		some line in numbers.range(item.start.row, item.end.row)
+	}
+}
+
+to_lines(report) = lines if {
```
lcarva, line 29: nit: the [style guide](https://github.com/StyraInc/rego-style-guide#avoid-prefixing-rules-and-functions-with-get_-or-list) makes me think that `lines` is preferred over `to_lines`.

Makefile
```diff
@@ -146,7 +150,7 @@ fmt-check: ## Check formatting of Rego files
 	@opa fmt . --list --fail >/dev/null 2>&1
 
 .PHONY: ci
-ci: quiet-test opa-check conventions-check fmt-check ## Runs all checks and tests
+ci: coverage-report quiet-test opa-check conventions-check fmt-check ## Runs all checks and tests
```
lcarva, line 153: Should this be executed after all the other checks? (Not sure which order it will be executed now.)

lcarva (approved):
Fantastic!

Makefile
simonbaird: Though it makes it less useful for humans running make ci locally... :thinking: ...

simonbaird: Not sure why would we care about coverage.json for a PR that is failing.

zregvart: Folk would get graphical representation of the coverage and see what cases need covering, we don’t get that currently

PR #249, opened 2023-01-11
Simplify Makefile consumption for new contributors

arewm:
* conftest installation now works on macos
* conftest/opa will be installed if needed by another target

Signed-off-by: arewm <arewm@users.noreply.github.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,17 +46,17 @@ help: ## Display this help.
 ##@ Development
 
 .PHONY: test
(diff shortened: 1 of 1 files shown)


simonbaird:
Lgtm I think. I like the suggestion of using `go run -modfile tools/go.mod`  but that could be done in a later PR.

simonbaird (approved):

PR #251, opened 2023-01-12
Smarter bundle pushing

simonbaird:
Move the bundle pushing to a bash script that detects if a push is needed. Copy the technique used in build-definitions repo, see hack/build-and-push.sh.

### DIFF
--- a/data/rule_data.yml
+++ b/data/rule_data.yml
@@ -1,12 +1,15 @@
 ---
+
 rule_data:
   allowed_registry_prefixes:
   - registry.access.redhat.com/
   - registry.redhat.io/
+
   allowed_step_image_registry_prefixes:
   - quay.io/redhat-appstudio/
   - registry.access.redhat.com/
   - registry.redhat.io/
+
   allowed_java_component_sources:
   - redhat
   - rebuilt
--- a/hack/pr-infra-deployments.sh
+++ b/hack/pr-infra-deployments.sh
@@ -22,25 +22,20 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-BUNDLE_TAG="git-$(git rev-parse --short HEAD)"
+GIT_REPO_NAME=infra-deployments
+GIT_ORIGIN=hacbs-contract/${GIT_REPO_NAME}
+GIT_UPSTREAM=redhat-appstudio/${GIT_REPO_NAME}
 
-RELEASE_BUNDLE_DIGEST="$(skopeo manifest-digest <(skopeo inspect --raw "docker://quay.io/hacbs-contract/ec-release-policy:${BUNDLE_TAG}"))"
-RELEASE_BUNDLE_REF="quay.io/hacbs-contract/ec-release-policy:${BUNDLE_TAG}@${RELEASE_BUNDLE_DIGEST}"
-
-DATA_BUNDLE_DIGEST="$(skopeo manifest-digest <(skopeo inspect --raw "docker://quay.io/hacbs-contract/ec-policy-data:${BUNDLE_TAG}"))"
-DATA_BUNDLE_REF="quay.io/hacbs-contract/ec-policy-data:${BUNDLE_TAG}@${DATA_BUNDLE_DIGEST}"
-
-GIT_REF="$(git rev-parse HEAD)"
-
-# setup
+# Clone repo to a temp dir
 WORKDIR=$(mktemp -d)
 trap 'rm -rf "${WORKDIR}"' EXIT
 cd "${WORKDIR}" || exit 1
-
-gh repo clone hacbs-contract/infra-deployments
+gh repo clone "$GIT_ORIGIN"
 cd infra-deployments || exit 1
+
+# Setup key for access in the GH workflow
 if [ -n "${GITHUB_ACTIONS:-}" ]; then
-  git remote set-url origin git@github.com:hacbs-contract/infra-deployments.git
+  git remote set-url origin git@github.com:${GIT_ORIGIN}.git
   git config --global user.email "${GITHUB_ACTOR}@users.noreply.github.com"
   git config --global user.name "${GITHUB_ACTOR}"
   mkdir -p "${HOME}/.ssh"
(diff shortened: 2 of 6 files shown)


simonbaird:
The workflow is untested.

zregvart (commented):
Might be nice to create a single PR instead of two by checking out the existing branch from the fork and adding a commit on top of the commits already there

hack/pr-infra-deployments.sh
```diff
@@ -49,21 +49,18 @@ if [ -n "${GITHUB_ACTIONS:-}" ]; then
...
-#DATA_SOURCE_URL=git::https://github.com/hacbs-contract/ec-policies.git//data?ref=${GIT_REF}
+BRANCH_NAME=$REPO-update
+git checkout -b ${BRANCH_NAME} --track upstream/main
+
+SOURCE_URL=oci::https://${BUNDLE_REF}
 
 # replacements
-yq e -i '.configMapGenerator[] |= select(.name == "ec-defaults").literals[] |= select(. == "ec_policy_source=*") = "ec_policy_source='"${POLICY_SOURCE_URL}"'"' components/enterprise-contract/kustomization.yaml
-yq e -i '.configMapGenerator[] |= select(.name == "ec-defaults").literals[] |= select(. == "ec_data_source=*"  ) = "ec_data_source='"${DATA_SOURCE_URL}"'"'     components/enterprise-contract/kustomization.yaml
+yq e -i '.configMapGenerator[] |= select(.name == "ec-defaults").literals[] |= select(. == "'"${CM_KEY_NAME}"'"ec_policy_source=*") = "ec_policy_source='"${SOURCE_URL}"'"' components/enterprise-contract/kustomization.yaml
```
lcarva, line 59: fyi, it might simplify a bit to read the env var via `env(CM_KEY_NAME)`, see [docs](https://mikefarah.gitbook.io/yq/operators/env-variable-operators#read-string-environment-variable).  I believe you'd have to export the env var first though.

hack/push-bundles.sh
```diff
@@ -0,0 +1,115 @@
...
+  [ $1 == "data" ] && echo "config.json" || echo "artifacthub-pkg.yml"
+}
+
+function repo_name() {
+  [ $1 == "data" ] && echo "ec-policy-data" || echo "ec-$1-policy"
+}
+
+function conftest_push() {
+  if [ $1 == "data" ]; then
+    conftest push --policy '' --data data $2 || true
```
lcarva, line 49: Why ignore errors?

hack/push-bundles.sh
```diff
@@ -0,0 +1,115 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Pushes policy bundles to quay.io, but only if anything changed since
+# the last bundle was pushed.
+#
+set -e -o pipefail
```
lcarva, line 21: Add `-u` ?

hack/push-bundles.sh
```diff
@@ -0,0 +1,115 @@
...
+    done
+
+    # Remove some files
+    exclude_files=$(exclusions $b)
+    for f in $exclude_files; do
+      find $content_dir -name $f -delete
+    done
+
+    # Debug content
+    #find $tmp_dir
```
lcarva, line 94: It might be good to uncomment this so we have an audit trail in CI. It would help debugging there for sure.

hack/push-bundles.sh
```diff
@@ -0,0 +1,115 @@
...
+
+    # Now push
+    cd $tmp_dir || exit 1
```
lcarva, line 97: Why the `exit 1`?

zregvart: ShellCheck [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164), I think I originally put that in...

hack/push-bundles.sh
simonbaird: Good catch. I was mainly for my hacking since I was always getting the same error. I'll remove.

hack/push-bundles.sh
simonbaird: It's pretty long, but yeah ok.

data/rule_data.yml
```diff
@@ -1,12 +1,15 @@
...
 rule_data:
   allowed_registry_prefixes:
   - registry.access.redhat.com/
   - registry.redhat.io/
+
   allowed_step_image_registry_prefixes:
   - quay.io/redhat-appstudio/
   - registry.access.redhat.com/
   - registry.redhat.io/
+
```
simonbaird, line 12: Intentional to trigger a new bundle with a fresh digest.

policy/lib/rule_data.rego
```diff
@@ -1,7 +1,7 @@
 package lib
 
-# Will define built-in default values here.
-# They can be overridden if required.
+# Values in data.rule_data_custom or data.rule_data
+# will take precedence over these defaults.
```
simonbaird, line 4: Also meant to cause a fresh bundle build.

hack/push-bundles.sh
simonbaird: I copied the opts from the other script.

hack/pr-infra-deployments.sh
simonbaird: Interesting idea, but I think I'll leave it alone for now.

simonbaird:
New revision now will create a single PR when both bundles change. I still haven't properly tested it in the workflow, but I'm thinking let's test in prod.. :smiling_imp: .

hack/update-bundles.sh
```diff
@@ -0,0 +1,121 @@
...
+
+    # Show the content
+    cd $tmp_dir || exit 1
+    find . -type f
+
+    # Now push
+    conftest_push $b "$push_repo:$tag"
+
+    # Set the 'latest' tag
+    #skopeo copy --quiet docker://$push_repo:$tag docker://$push_repo:latest
```
lcarva, line 104: Intentional?

lcarva (approved):

simonbaird:
Seems to be working:

* https://github.com/redhat-appstudio/infra-deployments/pull/1148
* https://github.com/hacbs-contract/ec-policies/actions/runs/3906274271/jobs/6674261394

But, as @zregvart mentioned in #1136 there's a test failing. I filed https://issues.redhat.com/browse/HACBS-1613 about fixing that.

PR #254, opened 2023-01-13
Make sure the correct key is set

zregvart:
The value of the key needs to be taken from the `CM_KEY` variable.

### DIFF
--- a/hack/pr-infra-deployments.sh
+++ b/hack/pr-infra-deployments.sh
@@ -62,7 +62,7 @@ while [ $# -gt 0 ]; do
 
   # Update the yaml file with the new source url
   yq e -i \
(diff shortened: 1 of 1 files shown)


lcarva (approved):

hack/pr-infra-deployments.sh
```diff
@@ -62,7 +62,7 @@ while [ $# -gt 0 ]; do
 
   # Update the yaml file with the new source url
   yq e -i \
-    '.configMapGenerator[] |= select(.name == "ec-defaults").literals[] |= select(. == "'${CM_KEY}'=*") = "ec_policy_source='${SOURCE_URL}'"' \
+    '.configMapGenerator[] |= select(.name == "ec-defaults").literals[] |= select(. == "'"${CM_KEY}"'=*") = "'"${CM_KEY}"'"='"${SOURCE_URL}"'"' \
```
simonbaird, line 65: Oops, sorry about that!

PR #256, opened 2023-01-16
Use attachments in documentation

zregvart:
This way we can capture the version that was present at the build time, and allow for different versions of the documentation to point to different files or files with varying content.

Ref. https://issues.redhat.com/browse/HACBS-1574

### DIFF
--- a/antora/docs/rule_data.yml
+++ b/antora/docs/rule_data.yml
@@ -1,1 +1,0 @@
-../../data/rule_data.yml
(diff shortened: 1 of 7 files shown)


antora/docs/rule_data.yml
```diff
@@ -1 +0,0 @@
-../../data/rule_data.yml
```
lcarva, line 1: Do we also need to remove `antora/docs/acceptable_tekton_bundles.yml` ?

lcarva (approved):

simonbaird (approved):
Nice.

PR #268, opened 2023-01-18
Update docs to use --all-namespaces

lcarva:


lcarva:
This should've been included in https://github.com/hacbs-contract/ec-policies/pull/267

zregvart (approved):

PR #282, opened 2023-01-25
Rename rule `short_name`s to keep them unique

zregvart:


zregvart:
Nicer version in #284, this kept only if we wish to go down this route

lcarva:
https://github.com/hacbs-contract/ec-policies/pull/284 has my vote.

zregvart:
I'm going to close this one in favor of #284

PR #284, opened 2023-01-25
Make rule anchors unique

zregvart:
Prepend parts of the package, namely type (release, pipeline) and package name to the rule short name and form an `anchor` property that is used to make sure that the ids/anchors point to unique values.

### DIFF
--- a/antora/ec-policies-antora-extension/index.js
+++ b/antora/ec-policies-antora-extension/index.js
@@ -114,6 +114,7 @@ const helpers = {
           const title = a.annotations.title
           const description = a.annotations.description
           const shortName = a.annotations.custom.short_name
+          const anchor = `${a.path.slice(1).map(e => e.value).join('_')}_${shortName}`
           const warningOrFailure = isWarn ? "warning" : "failure"
           const failureMsg = a.annotations.custom.failure_msg
           const effectiveOn = a.annotations.custom.effective_on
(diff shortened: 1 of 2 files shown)


zregvart:
#282 is a bit uglier version that fixes the same issue

antora/ec-policies-antora-extension/index.js
```diff
@@ -114,6 +114,7 @@ const helpers = {
           const title = a.annotations.title
           const description = a.annotations.description
           const shortName = a.annotations.custom.short_name
+          const anchor = `${a.path[2].value}_${a.path[3].value}_${shortName}`
```
lcarva, line 117: Does `fullPath` hold pretty much the same value?

lcarva (approved):

antora/ec-policies-antora-extension/index.js
zregvart: It has dots and no `shortName` in it, so not the best option...

simonbaird: I was thinking `${packagePath}_${shortName}` might, or `${packagePath.replaceAll('.', '_')}_${shortName}`.

simonbaird (approved):

antora/ec-policies-antora-extension/index.js
```diff
@@ -114,6 +114,7 @@ const helpers = {
           const title = a.annotations.title
           const description = a.annotations.description
           const shortName = a.annotations.custom.short_name
+          const anchor = `${a.path.slice(1).map(e => e.value).join('_')}_${shortName}`
```
zregvart, line 117: Perhaps this version works?

simonbaird: Lgtm.

simonbaird (approved):

PR #296, opened 2023-01-27
Add operator-framework/opm to allowed registries

simonbaird:
This is needed in particular for FBC builds. As a short term
solution let's just add it to the default allowed list.

In the longer term there should be some thinking about what useful
defaults would be and how we might maintain different versions of
these lists, e.g. for RH internal use and for "5 clicks" users.

### DIFF
--- a/data/rule_data.yml
+++ b/data/rule_data.yml
@@ -5,6 +5,7 @@ rule_data:
   allowed_registry_prefixes:
   - registry.access.redhat.com/
   - registry.redhat.io/
+  - quay.io/operator-framework/opm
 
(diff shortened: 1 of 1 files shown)


simonbaird:
Not sure if this is a good idea or not.

See [related slack discussion](https://redhat-internal.slack.com/archives/C031J4KBFME/p1674838155271879).

simonbaird:
On slack it's being suggested that the build can use [registry.redhat.io/openshift4/ose-operator-registry:v4.x](http://registry.redhat.io/openshift4/ose-operator-registry:v4.x), so I think this is to be abandoned.

PR #301, opened 2023-01-31
Add missing task-bundles to acceptable list

lcarva:
These were missed due to the know timinig issue.

https://issues.redhat.com/browse/HACBS-1709


lcarva:
I verified these were correct by using a mix of `skopeo inspect` and quay's web UI to ensure the image did indeed exist and to check when it was created so I could set effective_on to 30 days from that.

joejstuart (approved):

PR #311, opened 2023-02-08
update to reflect the new format

joejstuart:
This pr https://github.com/hacbs-contract/ec-cli/pull/394 updated the format

### DIFF
--- a/data/acceptable_tekton_bundles.yml
+++ b/data/acceptable_tekton_bundles.yml
@@ -265,21 +265,21 @@ pipeline-bundles:
       tag: d2e1cbe0cab869ade97d9b6c7235a23797e9b119
 pipeline-required-tasks:
   nodejs:
-    effective_on: "2023-03-11T00:00:00Z"
-    tasks:
-      - clair-scan
-      - clamav-scan
-      - configure-build
-      - deprecated-image-check
-      - git-clone
-      - init
-      - prefetch-dependencies
-      - s2i-nodejs
-      - sanity-inspect-image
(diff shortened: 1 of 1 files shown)


data/acceptable_tekton_bundles.yml
```diff
@@ -265,21 +265,21 @@ pipeline-bundles:
       tag: d2e1cbe0cab869ade97d9b6c7235a23797e9b119
 pipeline-required-tasks:
   nodejs:
```
lcarva, line 267: How come nodejs is the only one here?

lcarva (approved):

data/acceptable_tekton_bundles.yml
joejstuart: Good question. I'm wondering if that's the only pipeline that's been updated since the track bundle changes went in?

lcarva: I doubt it... Usually all the pipelines are always updated. You can confirm this by looking at the recent PRs from build-definitions coming to this repo.

joejstuart: something may be wrong then. I'll check it out. thanks.

zregvart (approved):
Looks good, not sure if we want to wait to see if pipelines other than `nodejs` appear?

joejstuart:
> Looks good, not sure if we want to wait to see if pipelines other than `nodejs` appear?
I'm looking into that. It won't hurt to merge this though.

PR #314, opened 2023-02-09
Update docs for include/exclude with term

lcarva:
https://issues.redhat.com/browse/HACBS-1744


antora/docs/modules/ROOT/templates/policy_configuration.hbs
```diff
@@ -63,6 +63,20 @@ by a dot. Rule codes can be see in the documentation also. For example the
 xref::release_policy.adoc#unknown_att_type[Unknown attestation type found] rule
```
lcarva, line 63: This whole document should probably be moved over to ec-cli at some point.

zregvart (approved):

PR #331, opened 2023-02-15
make install command on a newline

joejstuart:
There are three install instructions that are showing up on the same line. This separates them to make them more readable.


lcarva (approved):
An explanation would be helpful!

zregvart (approved):

PR #358, opened 2023-03-01
Add cve policy rules

lcarva:
https://issues.redhat.com/browse/HACBS-1671

### DIFF
--- a/policy/lib/rule_data.rego
+++ b/policy/lib/rule_data.rego
@@ -28,6 +28,9 @@ rule_data_defaults := {
 	],
 	"skipped_tests_results": ["SKIPPED"],
 	"warned_tests_results": ["WARNING"],
+	#
+	# Used in release/cve.go
+	"restrict_cve_security_levels": ["critical", "high"],
 }
 
 # Returns the "first found" of the following:
--- a/policy/release/cve.rego
+++ b/policy/release/cve.rego
@@ -1,0 +1,82 @@
+#
+# METADATA
+# title: CVE checks
+# description: >-
+#   This package is responsible for verifying a CVE scan was performed during
+#   the build pipeline, and that the image under test does not contain CVEs
+#   of certain security levels.
+package policy.release.cve
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+
+# METADATA
+# title: Found CVE vulnerabilities
+# description: >-
+#   The SLSA Provenance attestation for the image is inspected to ensure CVEs
+#   of certain security levels have not been detected. If detected, this policy
+#   rule will fail. By default, only CVEs of critical and high security level
+#   cause a failure. This is configurable by the rule data key
+#   "restrict_cve_security_levels". The available levels are critical, high,
+#    medium, and low.
+# custom:
+#   short_name: found_cve_vulnerabilities
+#   failure_msg: Found %d CVE vulnerabilities of %s security level
+#   collections:
+#   - minimal
+deny contains result if {
+	some level, amount in _non_zero_levels("restrict_cve_security_levels")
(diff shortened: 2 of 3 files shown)


policy/lib/rule_data.rego
```diff
@@ -28,6 +28,8 @@ rule_data_defaults := {
 	],
 	"skipped_tests_results": ["SKIPPED"],
 	"warned_tests_results": ["WARNING"],
+	# Used in release/cve.go
+	"restrict_cve_security_levels": ["critical"],
```
arewm, line 32: You only have a restict/deny rule. Would it also be beneficial to create a warn rule as well? We can initially populate it with lower severity CVEs.

I don't know what it might look like eventually -- whether importants should be a deny or not.

lcarva: See the acceptance criteria in https://issues.redhat.com/browse/HACBS-1671. I initially had it as a warning, but Matt explicitly requested no warnings to be emitted.

mattreid: Is this only looking for Critical right now? The AC was expecting either critical or high to trigger a failure.
Generating warnings can be up for debate, basing this on our productization requirements, low and moderate aren't required to be patched, so the initial thinking was to minimize noise and not generate warnings that teams may never fix up. 
@arewm would you prefer to see warnings generated based on the presence of low/moderate cves?

lcarva: > Is this only looking for Critical right now? The AC was expecting either critical or high to trigger a failure, if present.

My bad! Fixed now.

It is going to be noisy, and cause confusion, if we emit those warnings since it's not an actionable item for the user. 

Another option is to make this configurable. By default we produce no warnings, but the user can configure it, with their own data bundle, to produce warnings for certain levels. We may be getting ahead of ourselves here though.

mattreid: > > Is this only looking for Critical right now? The AC was expecting either critical or high to trigger a failure, if present.
> 
> My bad! Fixed now.

Thanks!

> It is going to be noisy, and cause confusion, if we emit those warnings since it's not an actionable item for the user.
> 
> Another option is to make this configurable. By default we produce no warnings, but the user can configure it, with their own data bundle, to produce warnings for certain levels. We may be getting ahead of ourselves here though.

Yeah, that would be desirable, but if it adds a lot of complexity, we could come back to it. There was this bullet in the AC: ```The severity levels that trigger a violation, warning, or are ignored should be configurable by a policy creator. Meaning that if a customer wanted to go a step farther and ensure that no Moderate CVEs are present, they could easily set that up. ```

Would that require a different implementation?

lcarva: I should read the AC more carefully.... 

With the current implementation, someone could provide their own list of security levels, to achieve failing on "critical", "high", **and** "medium". In fact, they could fail only on "low" if they really want to. :smile: 

I don't think adding warning support as I suggested earlier adds too much complexity. And we can verify the behavior with the unit tests. Stay tuned.

mattreid: Does this need any other changes to get it included in the minimal policy collection, and thereby enforced as part of the default CR?

arewm: That is along the lines of what I was initially thinking -- creating an empty warning set to make it easily configurable. I just worked myself up to put _something_ in there by default.

lcarva: I believe this is all set now. Also added these new policy rules to the minimal collection. Thanks for the feedback!

arewm (approved):

policy/lib/rule_data.rego
mattreid: Thank you, Lui! Looking forward to seeing it in action soon when I run EC as an integration test.

simonbaird (approved):

PR #371, opened 2023-03-06
Add ART repo to list of allowed step images

lcarva:
https://issues.redhat.com/browse/HACBS-1948


lcarva:
Burr is reporting a new failure where it uses the image from ’image-registry.openshift-image-registry.svc:5000/openshift/cli`...

I'm a bit hesitant in adding that to the list of allowed prefixes. Although... the `openshift` namespace should be pretty locked down.

zregvart (approved):

PR #373, opened 2023-03-06
create policy for a build task

joejstuart:
Checking for the build_type label and that a label exists in the build task.

https://issues.redhat.com/browse/HACBS-1957

### DIFF
--- a/policy/build_task/labels.rego
+++ b/policy/build_task/labels.rego
@@ -4,7 +4,7 @@
 # description: |-
 #   Checks related to build tasks
 #
-package policy.task.build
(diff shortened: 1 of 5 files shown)


zregvart:
This applies to any task, should it apply to a build task only? Might be difficult to use this in a policy, one would need to include/exclude this per task.

joejstuart:
> This applies to any task, should it apply to a build task only? Might be difficult to use this in a policy, one would need to include/exclude this per task.

Moved to `policy/build_task`

zregvart (approved):

PR #374, opened 2023-03-07
Use image source revision to deduce tag

zregvart:
This adds the image source revision, `org.opencontainers.image.revision` annotation to the Rego OCI bundle's manifest and uses it to deduce the alternate tag for the `latest` tag of an image when updating the resources in infra-deployments.

We need to keep the source revision (git commit id) within the bundle because not all commits end up pushing an updated bundle. So when we run the update of infra-deployment resources we do not know which commit id-based tag corresponds to the `latest` tag of the Rego OCI bundle we wish to update to.

An alternative to list all tags, fetch manifests and compare those with the `latest` tag is overly expensive (by number of requests) and requires authentication against quay.io to fetch a manifest by tag.

Ref. https://issues.redhat.com/browse/HACBS-1820

### DIFF
--- a/hack/update-bundles.sh
+++ b/hack/update-bundles.sh
@@ -71,6 +71,12 @@ function ensure_unique_file() {
   fi
 }
 
+tmp_oci_dirs=()
+function cleanup() {
+  rm -rf "${tmp_oci_dirs[@]}"
+}
+trap cleanup EXIT
+
 for b in $BUNDLES; do
   # Find the git sha where the source files were last updated
   src_dirs=$(bundle_src_dirs $b)
@@ -92,7 +98,7 @@ for b in $BUNDLES; do
 
     # Prepare a temp dir with the bundle's content
     tmp_dir=$(mktemp -d -t ec-bundle-$b.XXXXXXXXXX)
-    trap 'rm -rf "${tmp_dir}"' EXIT
+    tmp_oci_dirs+=("${tmp_dir}")
     content_dir=$tmp_dir/$(bundle_subdir $b)
     mkdir $content_dir
     for d in $src_dirs; do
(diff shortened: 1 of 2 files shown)


hack/update-bundles.sh
```diff
@@ -120,6 +120,13 @@ for b in $BUNDLES; do
     # Now push
     conftest_push $b "$push_repo:$tag"
 
+    # Add OCI annotations to the bundle image
+    tmpdir="$(mktemp -d --tmpdir)"
```
simonbaird, line 124: Do we want to trap cleanup this one also?

hack/update-infra-deployments.sh
```diff
@@ -33,7 +33,18 @@ RELEASE_POLICY_REF='quay.io/hacbs-contract/ec-release-policy:latest'
...
-  digest="$(skopeo inspect --raw  "docker://${img}" | sha256sum | awk '{print $1}')"
+  manifest="$(mktemp --tmpdir)"
+  function cleanup() {
+    # shellcheck disable=SC2317
+    rm "${manifest}"
+  }
+  trap cleanup RETURN
+  skopeo inspect --raw  "docker://${img}" > "${manifest}"
+  revision="$(jq -r '.annotations["org.opencontainers.image.revision"]' "${manifest}")"
+  if [[ -n "${revision}" && "${revision}" != null ]]; then
```
simonbaird, line 44: It's a string null here right? For readability maybe:

```suggestion
  if [[ -n "${revision}" && "${revision}" != "null" ]]; then
```

simonbaird (approved):

hack/update-infra-deployments.sh
zregvart: No need to quote non-space-containing strings in shell...

hack/update-bundles.sh
zregvart: Added a common trap

hack/update-infra-deployments.sh
simonbaird: Yeah, I know it's not needed, but when I read this code I spent a few moments wondering if a bare null had some significance in bash that I didn't remember. It doesn't of course, but we could consider optimizing to reduce brain cycles. :smile:

hack/update-bundles.sh
```diff
@@ -120,6 +126,14 @@ for b in $BUNDLES; do
     # Now push
     conftest_push $b "$push_repo:$tag"
 
+    # Add OCI annotations to the bundle image
+    tmp_oci_dir="$(mktemp -d --tmpdir)"
+    tmp_oci_dirs+=("${tmp_oci_dir}")
+    skopeo copy docker://"$push_repo:$tag" dir:"${tmp_oci_dir}"
+    manifest="$(jq -c '. += { "annotations": { "org.opencontainers.image.revision": "'"${last_update_sha}"'" } }' "${tmp_oci_dir}/manifest.json")"
```
simonbaird, line 133: Not sure exactly what the jq is doing, but I'll assume it's right.

simonbaird (approved):

hack/update-bundles.sh
zregvart: `-c` is for compact output (not indented), and the rest is adding `annotations["org.opencontainers.image.revision"]` to the manifest. Kinda emphasizes that we need [HACBS-1958](https://issues.redhat.com/browse/HACBS-1958)

PR #379, opened 2023-03-08
Add default cve config to rule_data file

simonbaird:
Show what the defaults are but leave them commented out. The idea is to make it easy for someone to modify the file to suit their preference, but avoid the possibility that the config values might obscure the real defaults.

### DIFF
--- a/data/rule_data.yml
+++ b/data/rule_data.yml
@@ -20,3 +20,7 @@ rule_data:
   allowed_java_component_sources:
   - redhat
   - rebuilt
+
+  # Defaults for reference. Levels are "critical", "high", "medium", and "low".
+  #restrict_cve_security_levels: ["critical", "high"]
+  #warn_cve_security_levels: []


data/rule_data.yml
```diff
@@ -20,3 +20,7 @@ rule_data:
   allowed_java_component_sources:
   - redhat
   - rebuilt
+
+  # Defaults for reference. Levels are "critical", "high", "medium", and "low".
```
lcarva, line 24: What about the other defaults in `policy/lib/rule_data.rego`? Do we need to mention them here as well? It's a bit confusing how there are two places to provide "default" values. So maybe, at least, we can mention here that some other defaults are in `policy/lib/rule_data.rego`?

lcarva (approved):

data/rule_data.yml
simonbaird: Yeah maybe. It's a bit inconsistent now. Will move to draft to consider.

simonbaird: Maybe documenting the defaults nicely is a better way to do it.

lcarva: FWIW, I thought I did document this sufficiently on the rule description, but maybe that's a good place to start?

simonbaird:
Replaced by https://github.com/hacbs-contract/ec-policies/pull/393 .

PR #382, opened 2023-03-09
Validate pipeline bundles being tracked

lcarva:


.github/workflows/pre-merge-ci.yaml
```diff
@@ -25,6 +25,14 @@ jobs:
     - name: Run checks
       run: make ci
 
+    - name: Validate added bundles
+      run: |
+        set -o errexit
+        set -o pipefail
+        set -o nounset
```
zregvart, line 32: Not sure we need this, doesn't hurt, so nitpickery

hack/validate-acceptable-bundles.sh
```diff
@@ -0,0 +1,70 @@
...
+set -o errexit
+set -o pipefail
+set -o nounset
+
+bundles_file='data/acceptable_tekton_bundles.yml'
+
+function list_pipeline_bundles() {
+    < "${1}" yq \
+        '."pipeline-bundles" | to_entries | .[] | .key + "@" + .value[].digest' | \
+        sort -u
```
zregvart, line 34: (nitpick, for fun) you could use [`unique`](https://mikefarah.gitbook.io/yq/operators/unique)

Something like:

```suggestion
    < "${1}" yq \
        '[."pipeline-bundles" | to_entries | .[] | .key + "@" + .value[].digest] | unique | .[]'
```

hack/validate-acceptable-bundles.sh
```diff
@@ -0,0 +1,70 @@
...
+}
+
+origin_bundles="$(list_pipeline_bundles <(curl -s "https://raw.githubusercontent.com/hacbs-contract/ec-policies/main/${bundles_file}"))"
+pr_bundles="$(list_pipeline_bundles "${bundles_file}")"
+new_bundles="$(comm -13 <(echo "${origin_bundles}") <(echo "${pr_bundles}"))"
```
zregvart, line 39: TIL `comm`, neat

hack/validate-acceptable-bundles.sh
```diff
@@ -0,0 +1,70 @@
...
+
+all_success=true
+for ref in ${new_bundles}; do
+    # Verify the image is accessible
+    skopeo inspect --raw "docker://${ref}" > /dev/null
```
zregvart, line 44: Do we want to error out if it isn't, or continue?

```suggestion
    skopeo inspect --raw "docker://${ref}" > /dev/null || { all_success=false; continue; }
```

zregvart (approved):
:+1: for the emoji!

hack/validate-acceptable-bundles.sh
lcarva: Yes! Also need `sort`

hack/validate-acceptable-bundles.sh
lcarva: Good call. Updated with a similar approach

.github/workflows/pre-merge-ci.yaml
lcarva: Removed it. Thanks

PR #385, opened 2023-03-15
Make sure we have ec and tkn commands in workflows

zregvart:


zregvart:
Works nicely https://github.com/hacbs-contract/ec-policies/actions/runs/4425572160 vs https://github.com/hacbs-contract/ec-policies/actions/runs/4426681364

PR #388, opened 2023-03-16
Move required-tasks to their own file

lcarva:


lcarva:
Marked as draft until https://github.com/hacbs-contract/ec-cli/pull/522 is merged and there's a snapshot release from it.

PR #393, opened 2023-03-20
Mention the built in defaults in rule data yaml

simonbaird:
Make it a little easier for users to know there are some other rule
data keys that they could override if they want to.

The initial motivation was to make it easier for users to be aware
of the two rule data keys related to the cve policy, since
customizing those is a thing that users might want to do. This is
documented in the rule itself, but I figure there's no harm in
making it discoverable in other ways.

Also set the default value of warn_cve_security_levels explicitly to
an empty list. It makes no functional difference, but it makes it
easier for someone reading the code to understand what is going on.

(This does bring up the question of why some keys have default
values in rule_data.rego and others don't, but let's consider that
later.)

### DIFF
--- a/data/rule_data.yml
+++ b/data/rule_data.yml
@@ -20,3 +20,6 @@ rule_data:
   allowed_java_component_sources:
   - redhat
   - rebuilt
+
+  # See also the additional default rule data values defined in
+  # https://github.com/hacbs-contract/ec-policies/blob/main/policy/lib/rule_data.rego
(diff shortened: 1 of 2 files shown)


data/rule_data.yml
```diff
@@ -20,3 +20,6 @@ rule_data:
   allowed_java_component_sources:
   - redhat
   - rebuilt
+
+  # See also the additional default rule data values defined in
+  # https://github.com/hacbs-contract/ec-policies/blob/main/policy/lib/rule_data.rego#L6
```
lcarva, line 25: We probably will never change this, but if we do add/remove lines to that file, it may break this link. Maybe just omit the line number altogether?

lcarva (approved):

PR #403, opened 2023-03-23
Add one more higher precedence rule data key

simonbaird:
The plan is that EC will create a data file using data found in the ECP CR. This provides a convenient way for a user to specify some custom data without needing to create a data source.

JIRA: HACBS-2042


zregvart (commented):
This works, but I think we should not have `rule_data.rego` at all. We are getting tightly coupled between ec-policies and ec-cli with this approach.

zregvart (approved):
Let's move forward with this

PR #427, opened 2023-04-05
Show solution annotation in docs output

simonbaird:
I added one solution as a pilot just so we can see that it works. Adding solution text for other rules is still todo.

JIRA: [HACBS-2049](https://issues.redhat.com/browse/HACBS-2049)


simonbaird:
=This is a piece of HACBS-2049, but there's another PR needed for ec-cli.

zregvart (approved):

lcarva (approved):

PR #431, opened 2023-04-12
Describe how a bundle becomes acceptable

lcarva:
https://issues.redhat.com/browse/HACBS-1809


zregvart (commented):
Perhaps this could include an example of creating a data bundle with custom trusted Tekton bundles? Then this document can be less about RHTAP and more about policies and trusted Tekton bundles, with a mention of RHTAP default trusted Tekton bundles.

lcarva:
@zregvart , reworked with your suggestion. Thanks!

zregvart (approved):
Supercool

PR #432, opened 2023-04-13
Adjust directories to fix make docs-preview

simonbaird:
The antora files docs in the main docs repo were moved down one directory recently.


lcarva:
Nice! :pray:

PR #436, opened 2023-04-17
Add provenance_materials policy package

lcarva:
https://issues.redhat.com/browse/HACBS-1302

### DIFF
--- a/policy/lib/tekton/task.rego
+++ b/policy/lib/tekton/task.rego
@@ -114,6 +114,16 @@ build_task(attestation) := task if {
 	count(trim_space(image_digest)) > 0
 }
 
+git_clone_task(attestation) := task if {
+	some task in tasks(attestation)
+
+	commit := task_result(task, "commit")
+	count(trim_space(commit)) > 0
+
+	url := task_result(task, "url")
+	count(trim_space(url)) > 0
+}
+
 # task_data returns the data relating to the task. If the task is
 # referenced from a bundle, the "bundle" attribute is included.
 task_data(task) = info if {
--- a/policy/release/provenance_materials.rego
+++ b/policy/release/provenance_materials.rego
@@ -1,0 +1,79 @@
+#
+# METADATA
+# title: Provenance Materials
+# description: >-
+#   This package provides rules for verifying the contents of the materials section
+#   of the SLSA Provenance attestation.
+#
+package policy.release.provenance_materials
+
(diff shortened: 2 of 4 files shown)


policy/release/provenance_materials.rego
```diff
@@ -0,0 +1,79 @@
...
+
+# METADATA
+# title: Task git-clone missing
+# description: >-
+#   The attestation must contain a git-clone task with the expected commit and url results.
+# custom:
+#   short_name: missing_git_clong_task
+#   failure_msg: Task git-clone not found
+#   collections:
+#   - minimal
```
lcarva, line 25: These seem like a good candidate for the minimal collection since this is expected to pass with the most basic rhtap pipeline.

policy/release/provenance_materials.rego
```diff
@@ -0,0 +1,79 @@
...
+}
+
+# METADATA
+# title: Git repo materials mismatch
+# description: >-
+#   The result of the git-clone task must be included in the materials section of the SLSA
+#   provenance attestation.
+# custom:
+#   short_name: git_repo_materials_mismatch
+#   failure_msg: Entry in materials for the git repo %q at commit %q not found
```
joejstuart, line 40: Replace `at` with `and`?

joejstuart: or maybe `or` since it's checking for both

joejstuart (approved):

policy/lib/tekton/task.rego
```diff
@@ -114,6 +114,16 @@ build_task(attestation) := task if {
 	count(trim_space(image_digest)) > 0
 }
 
+git_clone_task(attestation) := task if {
```
zregvart, line 117: I know we're all in on git, but we could do some cheap future proofing for when we need to support something else, say:
```suggestion
scm_fetch_task(attestation) := task if {
```
And for other mentions of git...

zregvart (approved):

policy/lib/tekton/task.rego
lcarva: I think `git` is right in this context since that's all that we support. For example, Chains the `git+` prefix unconditionally!

policy/release/provenance_materials.rego
lcarva: Thanks! I changed it to `and` because both must be present.

PR #438, opened 2023-04-19
Revise rule titles and short names

simonbaird:
The idea is that the title and name should describe what is being checked, rather than describe the failure condition. This means the title text makes more sense when used in the UI if the rule is passing or if the rule is failing.

I'm also changing the short name for consistency, even though the short name is not as exposed via the web UI.

JIRA: [HACBS-2095](https://issues.redhat.com/browse/HACBS-2095)

### DIFF
--- a/policy/release/base_image_registries.rego
+++ b/policy/release/base_image_registries.rego
@@ -14,14 +14,15 @@ import future.keywords.in
 import data.lib
 
 # METADATA
-# title: Restrict registry of base images
+# title: Base image comes from permitted registry
 # description: >-
 #   The base images used when building a container image must come from a known set
-#   of trusted registries to reduce potential supply chain attacks. This policy
-#   defines trusted registries as registries that are fully maintained by Red Hat
-#   and only contain content produced by Red Hat.
+#   of trusted registries to reduce potential supply chain attacks. By default this
+#   policy defines trusted registries as registries that are fully maintained by Red
+#   Hat and only contain content produced by Red Hat. The list of permitted registries
+#   can be customized by setting the `allowed_registry_prefixes` list in the rule data.
 # custom:
-#   short_name: disallowed_base_image
+#   short_name: base_image_permitted
 #   failure_msg: Base image %q is from a disallowed registry
 #   collections:
 #   - minimal
@@ -33,12 +34,13 @@ deny contains result if {
 }
 
 # METADATA
-# title: Base images must be provided
+# title: Base image task result was provided
 # description: >-
 #   The attestation must provide the expected information about which base images
-#   were used during the build process.
+#   were used during the build process. The base image information is expected to
+#   be found in a task result called `BASE_IMAGES_DIGESTS`.
 # custom:
-#   short_name: base_images_missing
+#   short_name: base_image_info_found
 #   failure_msg: Base images result is missing
 #   collections:
 #   - minimal
@@ -54,12 +56,12 @@ deny contains result if {
 }
 
 # METADATA
-# title: Missing rule data
+# title: Allowed base image registry prefixes list was provided
 # description: >-
-#   The policy rules in this package require the allowed_registry_prefixes
+#   The policy rules in this package require the `allowed_registry_prefixes`
 #   rule data to be provided.
 # custom:
-#   short_name: missing_rule_data
+#   short_name: allowed_registries_provided
 #   failure_msg: Missing required allowed_registry_prefixes rule data
 #   collections:
 #   - minimal
--- a/policy/release/cve.rego
+++ b/policy/release/cve.rego
@@ -5,6 +5,7 @@
 #   This package is responsible for verifying a CVE scan was performed during
 #   the build pipeline, and that the image under test does not contain CVEs
 #   of certain security levels.
+#
 package policy.release.cve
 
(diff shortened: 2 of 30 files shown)


simonbaird:
WIP. I'm about 60% through the text changes and 0% through the test fixing.

policy/release/cve.rego
```diff
@@ -42,28 +44,30 @@ deny contains result if {
...
 	some level, amount in _non_zero_levels("warn_cve_security_levels")
 	result := lib.result_helper_with_term(rego.metadata.chain(), [amount, level], level)
 }
 
 # METADATA
-# title: Missing CVE scan results
+# title: CVE scan results found
 # description: >-
 #   The clair-scan task results have not been found in the SLSA Provenance
 #   attestation of the build pipeline.
```
mattreid, line 61: Just a note here, that we're looking at introducing the CRDA task in some pipeline configurations, so there could be multiple tasks (or different tasks) providing vulnerability results in the future.

policy/release/base_image_registries.rego
```diff
@@ -33,12 +34,13 @@ deny contains result if {
 }
 
 # METADATA
-# title: Base images must be provided
+# title: Base image information was provided
 # description: >-
 #   The attestation must provide the expected information about which base images
-#   were used during the build process.
+#   were used during the build process. The base image information is expected to
+#   be found in a task result called `BASE_IMAGES_DIGESTS`.
```
arewm, line 41: (not to be addressed in this PR, but commenting as I see it) Are task results like these configurable to make it easier to modify the policies for various applications as others might adopt it?

arewm (changes requested):
Many of the titles and shortnames use a singular form (i.e. task) when the are really checking to ensure that multiple items are in conformance. Should these forms be changed to plural (and plural possessive) variants?

I think that singular forms make more sense for the negatives as we are looking for even a single bad occurrence. When referencing the positive, we want to ensure that _all_ are matching the policy.

simonbaird:
>

simonbaird:
Ready for review now I think. Moving out of draft.

arewm:
I didn't review all of the items, but the comments I did make can be generalized to other places where appropriate.

policy/release/cve.rego
simonbaird: I think it's clear enough from the description and the solution text that it's clair doing the scanning. I'd rather keep the back end detail out of the title for now. Introducing another source of CVEs will be a non-trivial change for the rego logic here I think, so let's revisit the rule titles then.

policy/release/base_image_registries.rego
```diff
@@ -14,14 +14,15 @@ import future.keywords.in
...
-# title: Restrict registry of base images
+# title: Base image comes from permitted registry
 # description: >-
 #   The base images used when building a container image must come from a known set
-#   of trusted registries to reduce potential supply chain attacks. This policy
-#   defines trusted registries as registries that are fully maintained by Red Hat
-#   and only contain content produced by Red Hat.
+#   of trusted registries to reduce potential supply chain attacks. By default this
+#   policy defines trusted registries as registries that are fully maintained by Red
+#   Hat and only contain content produced by Red Hat. The list of permitted registries
```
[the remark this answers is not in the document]
simonbaird: It's a good question. We are kind of in-between a "RH and RHTAP specific" and "General purpose" now. The plan is to become more of the latter over time, but I think it's beyond the scope for this PR.

policy/release/base_image_registries.rego
simonbaird: Yes and no. Yes in the sense that you can write your own custom rego to use whatever field you want. No, in the sense that you can't easily use this rule with a different task result name.

simonbaird: If we did want to make it more customizable, the var
```
build_base_images_digests_result_name := "BASE_IMAGES_DIGESTS"
```
could switched to rule data, e.g.
```
build_base_images_digests_result_name := lib.rule_data("base_image_digests_result_name")
```
with a suitable default value.

Then the user would have a nice way to use this same rule with their own preferred task result name.

policy/release/base_image_registries.rego
```diff
@@ -14,14 +14,15 @@ import future.keywords.in
...
 
 # METADATA
-# title: Restrict registry of base images
+# title: Base image comes from permitted registry
 # description: >-
 #   The base images used when building a container image must come from a known set
-#   of trusted registries to reduce potential supply chain attacks. This policy
-#   defines trusted registries as registries that are fully maintained by Red Hat
-#   and only contain content produced by Red Hat.
+#   of trusted registries to reduce potential supply chain attacks. By default this
```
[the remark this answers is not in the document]
simonbaird: I think generally yes, but I'll file another story rather than delay this further. Scope here is supposed to be titles and names. The bonus description tweaks are technically out of scope. (I did considered moving them to another PR, but let's keep it loose.)

simonbaird:
Re "Many of the titles and shortnames use a singular form", you're right. I did actually realize this at some point and start to use the plural. Will give it another quick review and see how consistent I was.

simonbaird:
A few last minute tweaks. I'll probably merge this once it goes green.

PR #447, opened 2023-04-26
Don't push already pushed tags

zregvart:
The `https://quay.io/api/v1/repository/hacbs-contract/tag?specificTag` API call requires authentication that we do not provide. This results in valid JSON and `curl` exiting with `0` as the `401` HTTP status does not result in non zero exit code with `curl`.

The `jq` following it will not find the tag in the JSON as it parses the JSON error response and the bundle will be pushed regardless of the tag existence.

This, for unknown reason, results in a different manifest digest overwriting the previous tag.

To fix this, this uses `skopeo list-tags` instead, which will use the existing Docker config credentials as needed.

### DIFF
--- a/hack/update-bundles.sh
+++ b/hack/update-bundles.sh
@@ -23,7 +23,6 @@ set -o pipefail
 set -o nounset
 
 REPO_ORG=hacbs-contract
-QUAY_API_URL=https://quay.io/api/v1/repository/$REPO_ORG
 ROOT_DIR=$( git rev-parse --show-toplevel )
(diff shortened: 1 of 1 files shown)


zregvart:
Note the two runs:

1. https://github.com/enterprise-contract/ec-policies/actions/runs/4801818781
2. https://github.com/enterprise-contract/ec-policies/actions/runs/4792018155

That push the same `git-8629680` tag for ec-policy-data bundle and result in different digests.

simonbaird (approved):

PR #449, opened 2023-04-26
Support new TEST_OUTPUT task result

simonbaird:
Fall back to the old name if required to make the transition more smooth. Eventually we can remove the deprecated name entirely.

JIRA: [HACBS-2159](https://issues.redhat.com/browse/HACBS-2159)

### DIFF
--- a/policy/release/lib/attestations.rego
+++ b/policy/release/lib/attestations.rego
@@ -14,7 +14,11 @@ taskrun_att_build_types := [
 	"https://tekton.dev/attestations/chains@v2",
 ]
 
-hacbs_test_task_result_name := "HACBS_TEST_OUTPUT"
+# (We can't call this test_task_result_name since anything prefixed
+# with test_ is treated as though it was a test.)
+task_test_result_name := "TEST_OUTPUT"
+
+task_test_result_name_deprecated := "HACBS_TEST_OUTPUT"
 
 java_sbom_component_count_result_name := "SBOM_JAVA_COMPONENTS_COUNT"
 
@@ -65,7 +69,15 @@ unmarshal(raw) = value {
 
 # (Don't call it test_results since test_ means a unit test)
 results_from_tests = results {
-	results := results_named(hacbs_test_task_result_name)
+	# First find results using the new task result name
+	_results := results_named(task_test_result_name)
+
+	# If some were found then return them
+	count(_results) > 0
+	results := _results
+} else = results {
+	# Try the older task result name, return it even if empty
+	results := results_named(task_test_result_name_deprecated)
 }
 
 # Check for a task by name. Return the task if found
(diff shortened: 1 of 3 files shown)


policy/release/lib/attestations.rego
```diff
@@ -14,7 +14,9 @@ taskrun_att_build_types := [
 	"https://tekton.dev/attestations/chains@v2",
 ]
 
-hacbs_test_task_result_name := "HACBS_TEST_OUTPUT"
+hacbs_test_task_result_name := "TEST_OUTPUT"
```
lcarva, line 17: Update to not has `hacbs` in the variable name?

policy/release/lib/attestations.rego
```diff
@@ -65,7 +67,15 @@ unmarshal(raw) = value {
 
 # (Don't call it test_results since test_ means a unit test)
 results_from_tests = results {
```
lcarva, line 69: Do we need to support when a mix of `HACBS_TEST_OUTPUT` and `TEST_OUTPUT` are used?

simonbaird: I was thinking let's not do that. I think it will make this patch a bit more complicated, and it seems unlikely it would happen.

I guess if it does happen, the impact is that ec will report a "required test not found" violation and someone will need to update their task definitions.

Wdyt?

lcarva: Users can update the tasks in their pipelines individually, so it's not that much of a corner case. But maybe it's not something we need.

policy/release/lib/attestations.rego
simonbaird: Done.

policy/release/lib/attestations.rego
```diff
@@ -14,7 +14,11 @@ taskrun_att_build_types := [
 	"https://tekton.dev/attestations/chains@v2",
 ]
 
-hacbs_test_task_result_name := "HACBS_TEST_OUTPUT"
+# (We can't call this test_task_result_name since anything prefixed
+# with test_ is treated as though it was a test.)
+task_test_result_name := "TEST_OUTPUT"
```
zregvart, line 19: Could we do it like this:

```suggestion
task_test_result_name := ".*TEST_OUTPUT"
```

And then change `results_named` not to test for equality but via `regex.match`? This is similar to type hinting in Chains.

simonbaird: It's a reasonable idea, but I don't want to do it here. I'd rather keep this PR as minimal as possible.

simonbaird: Thinking about it some more, that implementation would probably be even more minimal, but it's functionally different, e.g. (a contrived example) IGNORE_THIS_TEST_OUTPUT would be included.

Also I think we can unblock some other PRs by merging this sooner.

simonbaird:
Let's merge to unblock https://github.com/redhat-appstudio/build-definitions/pull/392 .

PR #536, opened 2023-07-07
Add redhat collection

lcarva:
https://issues.redhat.com/browse/HACBS-2370


simonbaird:
Lgtm. Is it every rule, or most rules?

lcarva:
> Lgtm. Is it every rule, or most rules?

Most rules. I skipped the `external_parameters` because that's for SLSA Provenance v1.0, and the github one for obvious reasons.

PR #543, opened 2023-07-13
Use golang modules to version and run cli tools

zregvart:
This way we can have Dependabot updates for tooling, akin to the setup in ec-cli.


zregvart:
In 1be96b878edce60a88ac6a2e8b4ef315c6d18884 I've made dependency replacements that fixed the several vulnerabilities, one remains in using Tekton Pipeline <0.49.0, the Tekton CLI is on 0.47.0 and we should not force the use of 0.49.0 as this will most likely result in compile errors. I think we should be okay to use it as is and update once Tekton CLI uses >=0.49.0 of Tekton Pipelines.

simonbaird (approved):

Makefile
```diff
@@ -210,17 +214,17 @@ INPUT_DIR=./input
...
 	  --policy $(POLICY_DIR) \
 	  --data $(DATA_DIR) \
 	  --no-fail \
 	  --output json
 
 .PHONY: check-pipeline
-check-pipeline: soft-install-tools ## Run policy evaluation for pipeline definition
-	@conftest test $(INPUT_FILE) \
+check-pipeline: ## Run policy evaluation for pipeline definition
+	@${CONFTEST} test $(INPUT_FILE) \
```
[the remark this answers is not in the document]
zregvart: No difference, I just forgot myself and went with the shell default, I'll switch to round ones to be in line with the rest of the file

hack/update-bundles.sh
```diff
@@ -55,9 +56,10 @@ function repo_name() {
 
 function conftest_push() {
   if [ $1 == "data" ]; then
-    $DRY_RUN_ECHO conftest push --policy '' --data data $2
+    $DRY_RUN_ECHO ${CONFTEST} push --policy '' --data data $2
   else
-    $DRY_RUN_ECHO conftest push --policy policy $2
+    pwd
```
lcarva, line 61: Intended?

lcarva (approved):
:heart:

hack/update-bundles.sh
zregvart: I was troubleshooting go modules, one can't point to a `go.mod` outside of the project tree, i.e. in `$DIR/a` one can't do `go run -modfile $DIR/b/go.mod`

lcarva:
Not sure why dependency review failed.

zregvart:
> Not sure why dependency review failed.

A vulnerable version of Tekton Pipeline was [detected](https://github.com/enterprise-contract/ec-policies/actions/runs/5554506622#summary-15046036377), details in the [comment above](https://github.com/enterprise-contract/ec-policies/pull/543#issuecomment-1634139147).

zregvart:
Merging, seems okayish now

PR #546, opened 2023-07-13
Check if provided string is a valid JSON

zregvart:
I could not create a unit test that would fail without `json.is_valid`, which is odd. I did test this with acceptance tests on https://github.com/enterprise-contract/ec-cli/pull/801 and it did help with that.


simonbaird (approved):
Lgtm.

PR #551, opened 2023-07-14
Add labels policy rules

lcarva:
https://issues.redhat.com/browse/HACBS-2372

### DIFF
--- a/data/rule_data.yml
+++ b/data/rule_data.yml
@@ -25,3 +25,84 @@ rule_data:
   - output-image
   # See also the additional default rule data values defined in
   # https://github.com/enterprise-contract/ec-policies/blob/main/policy/lib/rule_data.rego
+
+  # Used in release/labels.rego. See the policy.release.labels docs for an explanation
+  # of how these are used.
+  deprecated_labels:
+  - name: INSTALL
+    replacement: install
+  - name: Architecture
+    replacement: architecture
+  - name: BZComponent
+    replacement: com.redhat.component
+  - name: Name
+    replacement: name
+  - name: RUN
+    replacement: run
+  - name: Release
+    replacement: release
+  - name: UNINSTALL
+    replacement: uninstall
+  - name: Version
+    replacement: version
+
+  # Used in release/labels.rego. See the policy.release.labels docs for an explanation
+  # of how these are used.
+  required_labels:
+  - name: architecture
+    description: Architecture the software in the image should target.
+  - name: build-date
+    description: Date/Time image was built as RFC 3339 date-time.
+  - name: com.redhat.component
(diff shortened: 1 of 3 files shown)


zregvart (approved):

data/rule_data.yml
```diff
@@ -25,3 +25,75 @@ rule_data:
...
+
+  optional_labels:
+  - name: maintainer
+    description: >-
+      The name and email of the maintainer (usually the submitter).
+      Should contain `@redhat.com` or `Red Hat`.
+  - name: summary
+    description: A short description of the image.
+
+  inherited_labels:
```
arewm, line 90: It isn't clear to me whether these labels are _okay_ to be inherited or not.

policy/release/labels.rego
```diff
@@ -0,0 +1,155 @@
...
+	not is_fbc
+} else := lib.rule_data("fbc_required_labels")
+
+optional_labels := lib.rule_data("optional_labels") if {
+	not is_fbc
+} else := lib.rule_data("fbc_optional_labels")
+
+inherited_labels := lib.rule_data("inherited_labels") if {
+	not is_fbc
+} else := lib.rule_data("fbc_inherited_labels")
```
arewm, line 147: This pattern seems like it might explode easily. As an example, once we add bundle images, they might have different rules. Are you okay with this precedence?

policy/release/labels.rego
```diff
@@ -0,0 +1,155 @@
+#
+# METADATA
+# description: >-
+#   Check if the image has the expected labels set. The rules in this package
+#   distinguish file-based catalog (FBC) images from all other images. When
+#   checking an FBC image, a policy rule may use a different set of rule data.
+#
+package policy.release.labels
```
arewm, line 8: I will assume that I don't need to look at how you defined the label-based interactions.

arewm (changes requested):

policy/release/labels.rego
lcarva: I think it's ok for now. We can revisit it when we do add other image types.

policy/release/labels.rego
lcarva: As a reviewer of this PR? Or do you mean in a broader sense as a user/author reading the docs?

arewm: As a reviewer.

policy/release/labels.rego
mattreid: We'll likely be making changes to the labels we choose to care about when we're closer to shipping containers that users can actually directly interact with. https://issues.redhat.com/browse/RHTAP-409 

A first pass to help our services get to production makes sense to me, and we can revise from there as needed.

simonbaird:
Very cool!

PR #553, opened 2023-07-14
Ignore certain result attributes in unit tests

lcarva:
https://issues.redhat.com/browse/HACBS-2374


lcarva:
I think ignoring these attributes is harmless, thus this PR. Let me know if you think this is not a good idea, and we can certainly drop this PR.

simonbaird (approved):
Awesome.

zregvart (approved):

PR #555, opened 2023-07-17
Use tools from go.mod in hack/validate-acc...sh

zregvart:


zregvart:
Merging to unblock #554

lcarva (commented):
LGTM!

PR #556, opened 2023-07-17
Fixup acceptable bundles

lcarva:
Introduce a new script, `hack/check-acceptable-tekton-bundles.sh`, to check if the latest acceptable bundle records match the latest images in the registry.

The script generated most of the changes in `data/acceptable_tekton_bundles.yml`.

The sast-go and sast-java-sec-check tasks had a change that caused their definition to be equal to one of its previous versions. Because of this `ec track bundle` did not re-add entries for those bundles, as it is expected. The issue is that in this case, this behavior doesn't make sense as we want the old digest to be treated as the most up-to-date task definition. I manually fixed up those entries. `ec track bundle` should be smarter about this. This will be done in: https://issues.redhat.com/browse/HACBS-2394

### DIFF
--- a/data/acceptable_tekton_bundles.yml
+++ b/data/acceptable_tekton_bundles.yml
@@ -503,6 +503,9 @@ task-bundles:
       effective_on: "2022-12-31T00:00:00Z"
       tag: "0.1"
   quay.io/redhat-appstudio-tekton-catalog/task-git-clone:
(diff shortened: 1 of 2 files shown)


lcarva:
> Should this be run in CI?

Good idea! Filed https://issues.redhat.com/browse/HACBS-2395

PR #560, opened 2023-07-20
HACBS-2319 update source link for builtin rule set

robnester-rh:
This commit fixes source links for builtin rule collection entries. These source links now point to the correct locations within the ec-cli repository.

### DIFF
--- a/antora/docs/modules/ROOT/templates/_rules.hbs
+++ b/antora/docs/modules/ROOT/templates/_rules.hbs
@@ -24,6 +24,13 @@
 * Code: `{{ packageInfo.shortName }}.{{ shortName }}`
 {{#if effectiveOn}}
 * Effective from: `{{ effectiveOn }}`{{/if}}
-* https://github.com/enterprise-contract/ec-policies/blob/main/{{ file }}#L{{ row }}[Source, window="_blank"]
+{{#if packageInfo.shortName}}
+    {{#if (isBuiltIn packageInfo.shortName)}}
+    * https://github.com/enterprise-contract/ec-cli/blob/main/docs/{{ file }}#L{{ row }}[Source, window="_blank"]
 
+    {{else}}
+    * https://github.com/enterprise-contract/ec-policies/blob/main/{{ file }}#L{{ row }}[Source, window="_blank"]
+
+    {{/if}}
+{{/if}}
 {{/each}}
--- a/antora/ec-policies-antora-extension/index.js
+++ b/antora/ec-policies-antora-extension/index.js
@@ -6,6 +6,11 @@ const opa = import("@zregvart/opa-inspect")
 // Helpers for handlebars templates
 const hbsHelpers = {
 
+  // Determines if value starts with 'builtin'
+  isBuiltIn: (a) => {
+    return a.startsWith("builtin")
+  },
+
   // Convert "foo" to "FOO"
   toUpper: (s) => {
     return s.toUpperCase()


antora/docs/modules/ROOT/templates/_rules.hbs
```diff
@@ -24,6 +24,13 @@
 * Code: `{{ packageInfo.shortName }}.{{ shortName }}`
 {{#if effectiveOn}}
 * Effective from: `{{ effectiveOn }}`{{/if}}
-* https://github.com/enterprise-contract/ec-policies/blob/main/{{ file }}#L{{ row }}[Source, window="_blank"]
+{{#if packageInfo.shortName}}
+    {{#if (isInArray packageInfo.shortName '["builtin_attestation", "builtin_image"]')}}
```
zregvart, line 28: We could future proof this a bit, and do something like:

```suggestion
    {{#if isBuiltin packageInfo.shortName }}
```

robnester-rh: I'd originally taken this approach but upon further consideration decided to do a more generic function in case we need to evaluate array membership in other situations.

zregvart: With this any new builtin check would need a change here, which for me is a big drawback of this approach.
The implementation of `isBuiltin` could look at the `builtin_` prefix and having it in the template as `isBuiltin` makes the template clearer. We currently don't need a `is*` function, so when such requirement occurs we can refactor to something like this`.

robnester-rh: I appreciate the feedback. We currently only have a single place (here) where we need to check array membership, but this approach is able handle any use case whereas the `isBuiltIn` approach would only evaluate a single use case. I'm going to proceed with this implementation. Feel free to follow-up if you feel it's worthwhile.

antora/ec-policies-antora-extension/index.js
```diff
@@ -6,6 +6,18 @@ const opa = import("@zregvart/opa-inspect")
 // Helpers for handlebars templates
 const hbsHelpers = {
 
+  // Determines if value is in array
+  isInArray: (value, array) => {
+    if (Array.isArray(array)) {
+      return array.includes(value);
+    } else if (typeof array === 'string') {
```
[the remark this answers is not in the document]
simonbaird: This is just a random idea, not a blocker for the PR. I was looking at https://stackoverflow.com/questions/18903003/passing-arbitrary-arguments-to-handlebars-helpers and now I'm doubtful that it will be possible.

antora/ec-policies-antora-extension/index.js
```diff
@@ -6,6 +6,18 @@ const opa = import("@zregvart/opa-inspect")
...
+      // Parse the array string into an actual array
+      const parsedArray = JSON.parse(array);
+      return parsedArray.includes(value);
+    }
+    return false;
+  },
+
```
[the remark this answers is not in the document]
simonbaird: Since we're in plain javascript, I guess here this would work here:

    isBuiltin: (value) => {
      return hbsHelpers.isInArray(value, ["builtin_attestation", "builtin_image"])
    }

Which I think provides another way to ditch the need for json.Parse.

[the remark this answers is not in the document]
simonbaird: Alternative version:

```
isBuiltin: (val) => {
  return val.startsWith("builtin_")
},
```

simonbaird (changes requested):
I'd like to see a version where the helper is called `isBuiltin` to keep the messy logic out of the template.

zregvart (approved):
LGTM

simonbaird (approved):

PR #583, opened 2023-07-27
Check OLM bundle image's manifests has pinned refs

zregvart:
Given the content of the OLM bundle image manifest files in input and the presence of the `operators.operatorframework.io.bundle.manifests.v1` label on the image this checks if all the references mentioned in the Python checker[1] are pinned, i.e. they contain the digest.

Ref. https://issues.redhat.com/browse/HACBS-2379

[1] https://github.com/containerbuildsystem/operator-manifest#pull-specifications

### DIFF
--- a/policy/release/olm.rego
+++ b/policy/release/olm.rego
@@ -1,0 +1,121 @@
+#
+# METADATA
+# description: >-
+#   Checks for Operator Lifecycle Manager (OLM) bundles.
+#
+package policy.release.olm
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+import data.lib.image
+
+olm_manifestv1 := "operators.operatorframework.io.bundle.manifests.v1"
+
+# METADATA
+# title: Unpinned images in OLM bundle
+# description: >-
+#   Checks the OLM bundle image for the presence of unpinned image references.
+#   Unpinned image pull refernces are references to images found in
+#   link:https://osbs.readthedocs.io/en/latest/users.html#pullspec-locations[varying
+#   locations] that do not contain a digest -- uniquely identifying the version of
+#   the image being pulled.
+# custom:
+#   short_name: unpinned_references
+#   failure_msg: The %q image reference is not pinned at %s.
+#   solution: >-
+#     Update the OLM bundle replacing the unpinned image reference with pinned image
+#     reference. Pinned image reference contains the image digest.
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	manifestDir := input.image.config.Labels[olm_manifestv1]
+
+	some path, manifest in input.image.files
+
+	# only consider files in the manifest path as determined by the OLM manifest v1 label
+	startswith(path, manifestDir)
+
+	# only consider this API prefix, disregard the version
+	startswith(manifest.apiVersion, "operators.coreos.com/")
+
+	# only consider CSV manifests
+	manifest.kind == "ClusterServiceVersion"
+
+	some i in all_image_ref(manifest)
+	i.ref.digest == "" # unpinned image references have no digest
+
+	result := lib.result_helper_with_term(rego.metadata.chain(), [image.str(i.ref), i.path], image.str(i.ref))
+}
+
+_name(o) := n if {
+	n := o.name
+} else := "unnamed"
+
+# Finds all image references and their locations (paths). Returns all image
+# references (parsed into components) found in locations as specified by:
+# https://github.com/containerbuildsystem/operator-manifest/blob/f24cd9374f5ad9fed04f47701acffa16837d940e/README.md#pull-specifications
+# and https://osbs.readthedocs.io/en/latest/users.html#pullspec-locations
+all_image_ref(manifest) := [e |
+	# NOTE: use comprehensions in here, trying to set a value for `imgs` that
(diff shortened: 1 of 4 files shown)


policy/release/olm.rego
```diff
@@ -0,0 +1,99 @@
...
+# Checks if the image is a OPA bundle
+olm_bundle if {
+	input.image.config.Labels[olm_manifestv1]
+}
+
+_name(o) := n if {
+	n := o.name
+} else := "unnamed"
+
+# Finds all image references and their locations (paths)
```
simonbaird, line 51: You could add a comment here with the link to documentation with explanations for all these. It might help future debuggers of this code.

policy/release/olm.rego
```diff
@@ -0,0 +1,99 @@
...
+all_image_ref(manifest) := [e |
+	some imgs in [
+		[r |
+			some i, related in manifest.spec.relatedImages
```
simonbaird, line 55: ..or maybe some comments on each of these to help the reader understand each one.

simonbaird (approved):
Nice.

policy/release/olm.rego
zregvart: Added a reference to the [README.md](https://github.com/containerbuildsystem/operator-manifest/blob/f24cd9374f5ad9fed04f47701acffa16837d940e/README.md#pull-specifications)

policy/release/olm.rego
zregvart: I think reference to the README above should be enough, I don't think the lookups are difficult to grasp, what they need to be is documented in that README. Would be nicer to point to a specification, but this is what we have...

policy/release/olm.rego
```diff
@@ -0,0 +1,114 @@
...
+	result := lib.result_helper_with_term(rego.metadata.chain(), [image.str(i.ref), i.path], image.str(i.ref))
+}
+
+_name(o) := n if {
+	n := o.name
+} else := "unnamed"
+
+# Finds all image references and their locations (paths). Returns all image
+# references (parsed into components) found in locations as specified by:
+# https://github.com/containerbuildsystem/operator-manifest/blob/f24cd9374f5ad9fed04f47701acffa16837d940e/README.md#pull-specifications
```
lcarva, line 56: Alternatively, consider adding those to the package docs. We should probably copy from https://github.com/containerbuildsystem/operator-manifest#pull-specifications instead of just linking to it given that we are re-implementing that.

zregvart:
Updated with changes in https://github.com/enterprise-contract/ec-cli/pull/866, i.e. `image.manifests` -> `image.files`

policy/release/olm.rego
```diff
@@ -0,0 +1,114 @@
...
+			some i, related in manifest.spec.relatedImages
+			r := {"path": sprintf("spec.relatedImages[%d].image", [i]), "ref": image.parse(related.image)}
+		],
+		[r |
+			manifest.metadata.annotations.containerImage
+			r := {"path": "annotations.containerImage", "ref": image.parse(manifest.metadata.annotations.containerImage)}
+		],
+		[r |
+			some key, val in manifest.metadata.annotations
+			ref := image.parse(val)
```
lcarva, line 79: `metadata.annotations[]` is really tricky. You basically have to traverse down the object. Whenever a value of type string is found, it must be parsed to extract any image refs. This could be a paragraph with multiple image refs.

policy/release/olm_test.rego
```diff
@@ -0,0 +1,76 @@
+package policy.release.olm
```
lcarva, line 1: See https://github.com/containerbuildsystem/operator-manifest/blob/master/tests/test_operator.py for the use cases that must be supported.

lcarva:
@chmeliik, given that you were the original implementer of all these rules in the python version, your feedback here would be greatly appreciated! :pray:

policy/release/olm.rego
chmeliik: If it's helpful, OSBS has its own docs about this https://osbs.readthedocs.io/en/latest/users.html#pullspec-locations (and also links to the python implementation for extracting pullspecs from annotations)

One thing to note about metadata.annotations: both OSBS and operator-manifest process not only the top-level `.metadata.annotations`, but also any `<arbitrary path through the manifest> | .metadata.annotations`

lcarva (approved):

policy/release/olm.rego
```diff
@@ -0,0 +1,121 @@
...
+			some i, related in manifest.spec.relatedImages
+			r := {"path": sprintf("spec.relatedImages[%d].image", [i]), "ref": image.parse(related.image)}
+		],
+		[r |
+			manifest.metadata.annotations.containerImage
+			r := {"path": "annotations.containerImage", "ref": image.parse(manifest.metadata.annotations.containerImage)}
+		],
+		[r |
+			some _, values in walk(manifest)
+			some key, val in values.annotations
```
zregvart, line 84: Just noting that this will try to process any key named `annotations`, perhaps this should be (but perhaps safer)?
```suggestion
			some key, val in values.metadata.annotations
```

lcarva: oh good point. Yeah, according to Adam's comment that would be the right behavior.

zregvart: Ended up being safe...

lcarva (approved):

PR #584, opened 2023-07-27
Disable telemetry in harden-runner

zregvart:


zregvart:
There was a quite a bit of workflows here, so I created a PR. Will merge now.

PR #597, opened 2023-08-01
Adopt the new input format for attestations

lcarva:
This change allows policy rules to be applied to either `input.attestations[].statement` or `input.attestations[]`. The first having higher precedence. At some point in the future, we'll drop support for the second. This is deferred to allow users to update the ec-cli version being used.

https://issues.redhat.com/browse/HACBS-2423

### DIFF
--- a/policy/release/lib/attestations.rego
+++ b/policy/release/lib/attestations.rego
@@ -23,29 +23,36 @@ java_sbom_component_count_result_name := "SBOM_JAVA_COMPONENTS_COUNT"
 build_base_images_digests_result_name := "BASE_IMAGES_DIGESTS"
 
 # These are the ones we're interested in
-pipelinerun_attestations := [att |
+pipelinerun_attestations := [statement |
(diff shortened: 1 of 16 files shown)


policy/release/lib/attestations.rego
```diff
@@ -23,30 +23,37 @@ java_sbom_component_count_result_name := "SBOM_JAVA_COMPONENTS_COUNT"
...
+taskrun_attestations := [statement |
 	att := input.attestations[_]
-	att.predicate.buildType == taskrun_att_build_types[_]
+	statement := _statement(att)
+	statement.predicate.buildType == taskrun_att_build_types[_]
 ]
 
+_statement(att) = statement {
+	statement := att.statement
+} else = att
```
lcarva, line 55: I verified that if I comment out the else-clause all tests still pass with the exception of the ones that explicitly check the older format.

zregvart (approved):

PR #663, opened 2023-09-04
SLSA source correlation check

zregvart:
Given the attested source code reference and the source image code reference this check will make sure of their equality.

A warning is issued for source image data not including the source reference.

A failure will be issued if no source reference is present in the materials section or if the materials does not match the expected image source reference for SLSA Provenance v0.2.

Similarly, a failure will be issued if no source reference is present in the resolvedDependencies or if the resolvedDependencies does not match the expected image source reference for SLSA Provenance v1.0.

Ref. https://issues.redhat.com/browse/HACBS-2247

### DIFF
--- a/policy/release/slsa_source_correlated.rego
+++ b/policy/release/slsa_source_correlated.rego
@@ -1,0 +1,149 @@
+#
+# METADATA
+# title: SLSA - Verification model - Source
+# description: >-
+#   SLSA v1 verification model states:
+#
+#   "...artifacts are verified to ensure they meet the producer defined
+#   expectations of where the package source code was retrieved from..."
+#
+#   This package correlates the provided source code reference with the source
+#   code referenced in the attestation.
+#
+package policy.release.slsa_source_correlated
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+
+# opa fmt will transform "\u0000" into "\x00" which subsequently can't be parsed
+# by OPA, see https://github.com/open-policy-agent/opa/issues/6220
+nul := base64.decode("AA==")
+
+# METADATA
+# title: Expected source code reference
+# description: >-
+#   Warn if the expected source code reference is not provided.
+# custom:
+#   short_name: expected_source_code_reference
+#   failure_msg: Expected source code reference was not provided for verification
+#   solution: >-
+#     Provide the expected source code reference in inputs.
+warn contains result if {
+	source := object.get(input, ["image", "source"], {})
+	count(source) == 0
+
+	result := lib.result_helper(rego.metadata.chain(), [])
+}
+
+# METADATA
+# title: Source reference
+# description: >-
+#   Attestation contains source reference.
+# custom:
+#   short_name: attested_source_code_reference
+#   failure_msg: The attested material contains no source code reference
+#   solution: >-
+#     Check that the attestation creation process includes the source code reference
+#     in the predicate.materials for SLSA Provenance v0.2, or in
+#     predicate.buildDefinition.resolvedDependencies for SLSA Provenance v1.0
+#     attestations. Check that the Version Control System prefix is the list of the
+#     supported VCS types in rule data (`supported_vcs` key).
+#   collections:
+#   - minimal
+#   - slsa1
+#   - slsa2
+#   - slsa3
+#   - redhat
+#   depends_on:
+#   - attestation_type.known_attestation_type
+#
+deny contains result if {
+	count(_source_references) == 0
+
+	result := lib.result_helper(rego.metadata.chain(), [])
+}
+
+# METADATA
+# title: Expected source code reference
+# description: >-
+#   Verify that the provided source code reference is the one being attested.
+# custom:
+#   short_name: expected_source_code_reference
+#   failure_msg: The expected source code reference %q is not attested
+#   solution: >-
+#     The source code reference in the attestation doesn't match the expected and
+#     provided source code reference. Make sure that the provided source code
+#     reference is correct, and if it is make sure that the build process is
+#     configured to retrieve the source code from the appropriate source code
+#     repository. Make sure that the source code reference is pointing to a
+#     explicit revision not to a symbolic identifier, e.g. a branch or tag name.
+#   collections:
+#   - minimal
+#   - slsa1
+#   - slsa2
+#   - slsa3
+#   - redhat
+#   depends_on:
+#   - attestation_type.known_attestation_type
+#
+deny contains result if {
+	count(_source_references) > 0
+
+	some vcs_type, vcs_info in input.image.source
+
+	# e.g. git+https://github.com/...
+	expected_vcs_uri := sprintf("%s+%s", [vcs_type, vcs_info.url])
+	expected_revision := vcs_info.revision
+	expected_sources := {
+		sprintf("%s@sha1:%s", [expected_vcs_uri, expected_revision]),
+		sprintf("%s@gitCommit:%s", [expected_vcs_uri, crypto.sha1(sprintf("commit %d%s%s", [count(expected_revision), nul, expected_revision]))]),
+	}
+
+	# TODO: this is rather loose, this checks that the expected source is
+	# one of the attested sources, thus allowing also the inclusion of
+	# unexpected source
+	count(expected_sources & _source_references) == 0
+
+	some attested_source in _source_references
+
+	result := lib.result_helper_with_term(rego.metadata.chain(), [sprintf("%s@%s", [expected_vcs_uri, expected_revision])], attested_source)
+}
+
+# SLSA Provenance v0.2
+_source_references contains ref if {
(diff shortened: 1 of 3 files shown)


policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+package policy.release.slsa_source_correlated
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+
+# METADATA
+# title: Expected source code reference not provided
```
lcarva, line 22: I think we want the title to be positive. Something like: "Expected source code reference".

Same for the `short_name`.

The reasoning is that these values are used whether or not the check fails. In contrast, the failure message should also be in the negative form since it only applies if there is indeed a failure/warning.

policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+# description: >-
+#   Warn if the expected source code reference is not provided.
+# custom:
+#   short_name: missing_expected_source_code_reference
+#   failure_msg: Expected source code reference was not provided for verification
+# solution: >-
```
lcarva, line 28: This should go under `custom`.

policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+#   Provide the expected source code reference in inputs.
+warn contains result if {
+	source := object.get(input, ["image", "source"], {})
+	count(source) == 0
```
lcarva, line 32: This is cool. I didn't know that's how you'd check for an empty object.

policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+	result := lib.result_helper(rego.metadata.chain(), [])
+}
+
+# METADATA
+# title: Source reference material not provided
```
lcarva, line 37: Consider: "Source reference material"

policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+# description: >-
+#   Attested materials contain source reference.
+# custom:
+#   short_name: material_code_reference
+#   failure_msg: The attested material contains no source code reference
+#   solution: >-
+#     Check that the attestation creation process includes the source code reference
+#     in the predicate.materials for SLSA Provenance v0.2, or in
+#     predicate.buildDefinition.resolvedDependencies for SLSA Provenance v1.0
+#     attestations. Check that the SCM prefix is  the list of the supported SCM
```
lcarva, line 47: Extra white space in `is  the list` ?

policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+
+# METADATA
+# title: Source reference material not provided
+# description: >-
+#   Attested materials contain source reference.
+# custom:
+#   short_name: material_code_reference
+#   failure_msg: The attested material contains no source code reference
+#   solution: >-
+#     Check that the attestation creation process includes the source code reference
```
lcarva, line 44: This is a solid solution.

policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+#   - attestation_type.known_attestation_type
+#
+deny contains result if {
+	count(_source_references) == 0
+
+	result := lib.result_helper(rego.metadata.chain(), [])
+}
+
+# METADATA
+# title: Provided source code reference is the one that is being attested
```
lcarva, line 65: Maybe "Expected source code reference" ?

policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+#   - minimal
+#   - slsa1
+#   - slsa2
+#   - slsa3
+#   - redhat
+#   depends_on:
+#   - attestation_type.known_attestation_type
+#
+deny contains result if {
+	some scm_type in object.keys(input.image.source)
```
lcarva, line 88: Nit: You could do this instead:

```rego
some scm_type, scm_info in input.image.source
...
expected_scm_uri := sprintf("%s+%s", [scm_type, scm_info.url])
```

See https://www.openpolicyagent.org/docs/latest/policy-reference/#objects-1

policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+
+	# e.g. git+https://github.com/...
+	expected_scm_uri := sprintf("%s+%s", [scm_type, input.image.source[scm_type].url])
+	expected_revision := input.image.source[scm_type].revision
+	expected_source := sprintf("%s@%s", [expected_scm_uri, expected_revision])
+
+	some attested_source in _source_references
```
lcarva, line 95: Should there be a rule that ensures there is some overlap between `_source_references` and the keys of `input.image.source` ? If there isn't, then I believe no violations/warnings will be reported which doesn't seem right.

policy/release/slsa_source_correlated.rego
```diff
@@ -0,0 +1,130 @@
...
+
+	expected_source != attested_source
```
lcarva, line 97: Instead of an exact match, should this just make sure that one of the items in `_source_references` match? IIUC, resolvedDependencies and materials may include other things. Maybe this?

```rego
expected_source in _source_references
```

policy/release/slsa_source_correlated_test.rego
```diff
@@ -0,0 +1,128 @@
+package policy.release.slsa_source_correlated
+
+import future.keywords.in
+
+import data.lib
+
```
lcarva, line 6: Maybe add a success case where no warnings and no violations are reported?

policy/release/slsa_source_correlated.rego
zregvart: Not following, `_source_references` is a set of strings, it is empty if no source references are attested, that case should be covered by `attested_source_code_reference` rule. Why would we check the keys of `input.image.source` those should be a given, no?

policy/release/slsa_source_correlated.rego
zregvart: I don't think so, there is one source reference in `input.image.source`, there should be only one source reference in the attestation. Having more than one violates the expectation of the one source reference from the `input.image.source`.

policy/release/slsa_source_correlated_test.rego
zregvart: The `test_*_happy_day` tests should cover this

policy/release/slsa_source_correlated.rego
lcarva: I think Chains will add git+ resources to resolvedDependencies if the git resolver is used: https://github.com/tektoncd/chains/blob/70c8c7de563ba3d6c1e65a1e4e21c83335fa432a/pkg/chains/formats/slsa/v2alpha2/internal/resolved_dependencies/resolved_dependencies_test.go#L519-L547

policy/release/slsa_source_correlated_test.rego
lcarva: ack. Missed that.

policy/release/slsa_source_correlated.rego
lcarva: You're right. I was thinking if rule data `supported_vcs` didn't contain `git` for example. I was worried that this would not trigger a violation. But I see now that it does.

policy/release/slsa_source_correlated.rego
zregvart: Awkward, brings to light the importance of https://github.com/tektoncd/chains/issues/914, I _think_ I can workaround that by eliminating the references from the steps from the ones considered, the PipelineRun attestation I worked with had a oci bundle IIRC without a resolver IIRC (old school). I'll experiment a bit and see...

zregvart: So, actually not, at least not with Chains v0.17.0 the `resolvedDependencies` contains the images used in the steps, not the git references to the task definitions. See: https://gist.github.com/zregvart/7cc23e92ba37333f70c07af7671cac5a#file-attestation-v1-json-L144-L170

lcarva: We learned that Chains will eventually populate resolvedDependencies with git-resolver repos. We also learned that we don't have a way of distinguishing those from an application source code.

I think we should allow for other git things to be in resolvedDependencies until we have a definitive way of distinguishing them.

zregvart: Using intersection of expected and attested now

zregvart:
Latest update is that [`gitCommit`](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#gitcommit-gittree-gitblob-gittag) in DigestSet (`predicate.resolvedDependencies.digest.gitCommit`) is supported.

lcarva (approved):

PR #688, opened 2023-09-12
Add schedule policy rules

lcarva:
https://issues.redhat.com/browse/HACBS-2608

### DIFF
--- a/policy/release/schedule.rego
+++ b/policy/release/schedule.rego
@@ -1,0 +1,53 @@
+#
+# METADATA
+# title: Schedule related checks
+# description: >-
+#   Rules that verify the current date conform to a given schedule.
+#
+package policy.release.schedule
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.lib
+
+# METADATA
+# title: Weekday Restriction
+# description: >-
+#   Check if the current weekday is allowed based on the rule data value from the key
+#   `disallowed_weekdays`. By default, the list is empty in which case *any* weekday is allowed.
+# custom:
+#   short_name: weekday_restriction
+#   failure_msg: '%s is a disallowed weekday: %s'
+#   solution: Try again on a different weekday.
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	today := lower(time.weekday(lib.time.effective_current_time_ns))
+	disallowed := {lower(w) | some w in lib.rule_data("disallowed_weekdays")}
+	count(disallowed) > 0
+	today in disallowed
+	result := lib.result_helper(rego.metadata.chain(), [today, concat(", ", disallowed)])
+}
+
+# METADATA
+# title: Date Restriction
+# description: >-
+#   Check if the current date is not allowed based on the rule data value
+#   from the key `disallowed_dates`. By default, the list is empty in which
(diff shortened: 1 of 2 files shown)


policy/release/schedule.rego
```diff
@@ -0,0 +1,60 @@
...
+#   case *any* day is allowed.
+# custom:
+#   short_name: date_restriction
+#   failure_msg: '%s is a disallowed date: %s'
+#   solution: Try again on a different day.
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	today := time.format(lib.time.effective_current_time_ns)
```
zregvart, line 49: Could use this

```suggestion
	today := time.format([lib.time.effective_current_time_ns, "UTC", "2006-01-02"])
```

policy/release/schedule.rego
```diff
@@ -0,0 +1,60 @@
...
+	disallowed := lib.rule_data("disallowed_dates")
+	count(disallowed) > 0
+
+	matches := [d |
+		some d in disallowed
+		startswith(today, d)
+	]
+	count(matches) > 0
```
zregvart, line 57: With above this could be:

```suggestion
	today in disallowed
```

policy/release/schedule_test.rego
```diff
@@ -0,0 +1,104 @@
...
+	lib.assert_empty(deny) with data.rule_data.disallowed_weekdays as ["friday"]
+		with data.config.policy.when_ns as monday
+}
+
+test_date_restriction {
+	year_violation := {{
+		"code": "schedule.date_restriction",
+		"msg": "2023-01-01T00:00:00Z is a disallowed date: 2023",
+	}}
+	lib.assert_equal_results(deny, year_violation) with data.rule_data.disallowed_dates as ["2023"]
```
zregvart, line 68: I don't think it makes much sense to support date ranges like this, best have the dates explicitly stated in full, a typo in the configuration could lead to wide denies

zregvart (commented):
Not sure about date ranges...

policy/release/schedule_test.rego
rhartman93: I think ranges would be useful, typos are always a risk, and you'd learn pretty quickly if every release attempt is getting blocked.

zregvart: And blocking on a 10day/month/01-09,10-12month/year range is okay? That just seems off to me. One could provide a list of dates and do arbitrary ranges instead, would that be okay?

rhartman93: Oh I missed some of the context, yeah blocking a whole year doesn't seem useful, arbitrary ranges is more what I had in mind. Days of the week, weekends, arbitrary ranges

policy/release/schedule.rego
lcarva: Nice! I missed this in the docs. Thanks

policy/release/schedule_test.rego
lcarva: Yeah, agree. I've changed it so now you have to use an explicit date, e.g. `2023-01-01`. Let's see how far we can get with this for now before adding support for ranges.

zregvart (approved):

PR #691, opened 2023-09-13
Update tasks operations so they support slsav1 attestations.

joejstuart:
This modifies the task library to handle tasks from the resolvedDepencies list in a slsav1 attestation.

This ONLY updates the task library and the `buildah_task` policy. More policies need to be updated, but I'm breaking them up into separate PRs to reduce the size.

https://issues.redhat.com/browse/HACBS-2246


zregvart (approved):
LGTM

PR #700, opened 2023-09-17
Update the slsav1 task format

joejstuart:
This is a follow-up to this https://github.com/enterprise-contract/ec-policies/pull/691. The task format is different than I expected. This PR fixes that mistake.

### DIFF
--- a/policy/lib/refs.rego
+++ b/policy/lib/refs.rego
@@ -30,6 +30,14 @@ task_ref(task) = i {
 		"kind": lower(_param(r, "kind", "task")),
 	}
 } else = i {
+	r := _ref(task)
+	r.resolver == "git"
+	i := {
+		"url": _param(r, "url", ""),
+		"revision": _param(r, "revision", ""),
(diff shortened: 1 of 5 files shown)


policy/lib/refs.rego
```diff
@@ -29,6 +29,14 @@ task_ref(task) = i {
...
 		"kind": lower(_param(r, "kind", "task")),
 	}
+} else = i {
+	r := _ref(task)
+	r.resolver == "git"
+	i := {
+		"url": _param(r, "url", ""),
+		"revision": _param(r, "revision", ""),
+		"pathInRepo": _param(r, "pathInRepo", ""),
+	}
```
zregvart, line 39: Seems we need coverage for this branch

zregvart (approved):
LGTM

policy/lib/refs.rego
joejstuart: thanks

PR #709, opened 2023-09-20
Update tasks policy to include the slsav1 attestation format

joejstuart:
This updates the time based required task list and the task status policies.

https://issues.redhat.com/browse/HACBS-2246

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -4,3 +4,7 @@ rules:
   style:
     todo-comment:
       level: warning
+    file-length:
+      level: warning
+    line-length:
+      max-line-length: 130
--- a/policy/lib/tekton/pipeline.rego
+++ b/policy/lib/tekton/pipeline.rego
@@ -35,4 +35,9 @@ pipeline_label_selector(pipeline, selector) := value if {
 	label == selector
 }
 
+pipeline_label_selector(pipeline, selector) := value if {
+	some label, value in pipeline.predicate.buildDefinition.internalParameters.labels
+	label == selector
+}
+
 pipeline_name := input.metadata.name
--- a/policy/lib/tekton/task.rego
+++ b/policy/lib/tekton/task.rego
@@ -28,7 +28,7 @@ tasks(obj) := {task |
 	_slsa_task(task)
 }
 
-# task from a slsav0.1 or slsav0.2 attestation
+# task from a slsav0.2 attestation
 _slsa_task(task) if {
 	task_ref := refs.task_ref(task)
 	task_ref.kind == "task"
@@ -95,7 +95,7 @@ task_names(task) := names if {
 	names := {name} | params
 }
 
-# task name from a v0.1 and v0.2 attestation
+# task name from a v0.2 and v1.0 attestation
 task_name(task) := refs.task_ref(task).name
 
 # _task_params returns an object where keys are parameter names
@@ -113,7 +113,7 @@ _task_params(task) := params if {
 	}
 }
 
-# handle params from a taskRun
+# handle params from a slsav1.0 attestation
 _task_params(task) := params if {
 	task.spec.params
 	params := {name: value |
--- a/policy/release/tasks.rego
+++ b/policy/release/tasks.rego
@@ -40,7 +40,7 @@ import data.lib.tkn
 #   - attestation_type.known_attestation_type
 #
 deny contains result if {
-	some att in lib.pipelinerun_attestations
+	some att in _pipelineruns
 	count(tkn.tasks(att)) == 0
 	result := lib.result_helper(rego.metadata.chain(), [])
 }
@@ -64,11 +64,11 @@ deny contains result if {
 #   - tasks.pipeline_has_tasks
 #
 deny contains result if {
-	some att in lib.pipelinerun_attestations
+	some att in _pipelineruns
 	some task in tkn.tasks(att)
-	status := _status(task)
+	some status in _status(task)
 	status != "Succeeded"
-	result := lib.result_helper_with_term(rego.metadata.chain(), [task.name, status], task.name)
+	result := lib.result_helper_with_term(rego.metadata.chain(), [tkn.task_name(task), status], tkn.task_name(task))
 }
 
 # METADATA
@@ -166,7 +166,7 @@ deny contains result if {
 # _missing_tasks returns a set of task names that are in the given
 # required_tasks, but not in the PipelineRun attestation.
 _missing_tasks(required_tasks) := {task |
-	some att in lib.pipelinerun_attestations
+	some att in _pipelineruns
 	count(tkn.tasks(att)) > 0
 
 	some task in required_tasks
@@ -176,7 +176,7 @@ _missing_tasks(required_tasks) := {task |
 # get the future tasks that are pipeline specific. If none exists
 # get the default list
 latest_required_tasks := task_data if {
-	some att in lib.pipelinerun_attestations
+	some att in _pipelineruns
 	count(tkn.tasks(att)) > 0
 	task_data := tkn.latest_required_pipeline_tasks(att)
 } else := task_data if {
@@ -186,7 +186,7 @@ latest_required_tasks := task_data if {
 # get current required tasks. fall back to the default list if
 # no label exists in the attestation
 current_required_tasks := task_data if {
-	some att in lib.pipelinerun_attestations
+	some att in _pipelineruns
 	count(tkn.tasks(att)) > 0
 	task_data := tkn.current_required_pipeline_tasks(att)
 } else := task_data if {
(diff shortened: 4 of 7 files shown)


policy/release/tasks.rego
```diff
@@ -197,11 +196,40 @@ current_required_tasks := task_data if {
...
+	condition.status == "True"
+	status := "Succeeded"
+}
+
+_slsav1_status(condition) := status if {
+	condition.status == "False"
+	status := "Failed"
+}
+
+_pipelinerun := att if {
```
joejstuart, line 223: I think I can move this or something similar to the `attestations.rego` lib, but wanna do that in a separate PR.

policy/lib/tekton/pipeline.rego
```diff
@@ -35,6 +35,11 @@ pipeline_label_selector(pipeline, selector) := value if {
 	label == selector
 }
 
+pipeline_label_selector(pipeline, selector) := value if {
```
lcarva, line 38: I noticed that this file is only tested indirectly. I think we should have a `pipeline_test.rego` file that tests all the rules in this file that don't start with `_`. Could we create one as part of this PR to just cover `pipeline_label_selector`?

policy/release/tasks.rego
```diff
@@ -40,7 +40,7 @@ import data.lib.tkn
 #   - attestation_type.known_attestation_type
 #
 deny contains result if {
-	some att in lib.pipelinerun_attestations
+	some att in _pipelinerun
```
lcarva, line 43: `_pipelineruns` ?

policy/release/tasks.rego
```diff
@@ -197,11 +197,40 @@ current_required_tasks := task_data if {
...
 	count(tkn.tasks(att)) > 0
 	task_data := tkn.required_task_list(att)
 }
 
-_status(task) := status if {
+_status(task) = status if {
+	not task.status.conditions
 	status := task.status
+} else = status if {
+	some condition in task.status.conditions
```
lcarva, line 209: The status is actually a list of conditions?!

policy/release/tasks.rego
```diff
@@ -197,11 +197,40 @@ current_required_tasks := task_data if {
...
 required_pipeline_task_data := task_data if {
-	some att in lib.pipelinerun_attestations
+	some att in _pipelinerun
 	count(tkn.tasks(att)) > 0
 	task_data := tkn.required_task_list(att)
 }
 
-_status(task) := status if {
+_status(task) = status if {
+	not task.status.conditions
```
lcarva, line 206: Consider reversing the order and adding a comment to explain the different use case:

```rego
_status(task) := status if {
  # Handle SLSA Provenance v1.0
  some condition in task.status.conditions
  condition.type == "Succeeded"
  status := _slsav1_status(condition)
} else := status if {
  # Handle SLSA Provenance v0.2
  status := task.status
} else := "MISSING"

File: policy/lib/tekton/task.rego [hunk 6]
```diff
@@ -99,7 +99,7 @@ task_names(task) := names if {
 	names := {name} | params
 }
 
-# task name from a v0.1 and v0.2 attestation
+# task name from a v0.1, v0.2 and v1.0 attestation
```
lcarva, line 102: I don't know if `v0.1` is true. I don't think we ever dealt with a version of Chains that emitted that.

policy/release/tasks.rego
```diff
@@ -197,11 +197,40 @@ current_required_tasks := task_data if {
...
+_slsav1_status(condition) := status if {
+	condition.status == "False"
+	status := "Failed"
+}
+
+_pipelinerun := att if {
+	att := [a |
+		some a in lib.pipelinerun_attestations
+	]
+	count(att) > 0
```
lcarva, line 228: Why not just return an empty list?

policy/release/tasks.rego
```diff
@@ -197,11 +197,40 @@ current_required_tasks := task_data if {
...
+}
+
+_pipelinerun := att if {
+	att := [a |
+		some a in lib.pipelinerun_slsa_provenance_v1
```
lcarva, line 233: What if an image has both? Maybe something like this is safer?

```rego
v1_0 := [a |
  some a in lib.pipelinerun_slsa_provenance_v1
]
v0_1 := [a |
  some a in lib.pipelinerun_attestations
]

attestations := array.concat(v1_0, v0_1)
```

policy/release/tasks_test.rego
```diff
@@ -16,19 +16,21 @@ test_no_tasks_present if {
...
+	lib.assert_equal_results(deny, expected) with input.attestations as _slsav1_attestations_with_tasks([], [])
 }
 
 test_failed_tasks if {
 	expected := {
 		{
 			"code": "tasks.successful_pipeline_tasks",
-			"msg": "Pipeline task \"av-scanner-p\" did not complete successfully, \"Failed\"",
-			"term": "av-scanner-p",
+			"msg": "Pipeline task \"av-scanner\" did not complete successfully, \"Failed\"",
```
lcarva, line 27: Why not use the pipeline task name in the error message?

policy/release/tasks.rego
joejstuart: Here's what I noticed in the taskRun. Maybe I need to check `reason` also?
```json
"status": {
    "conditions": [
      {
        "type": "Succeeded",
        "status": "True",
        "lastTransitionTime": "2023-09-15T19:17:45Z",
        "reason": "Succeeded",
        "message": "All Steps have completed executing"
      }
    ],
```

policy/release/tasks.rego
joejstuart: I get this error when trying that
```
eval_conflict_error: complete rules must not produce multiple outputs
```

policy/release/tasks_test.rego
joejstuart: I did change that I guess out of convenience. The rule that it tests was printing `task.name` which doesn't exist in slsav1.0 and we already have a helper that pulls the task name from `task.ref.name` which I switched to. Shoudln't be hard to add a helper for the pipeline task name also. Here's the code. https://github.com/enterprise-contract/ec-policies/pull/709/files#diff-e9948535fcfae404a22da4f7fc963cff6fa6faa48fd26a6df42f5f073c6eb3edR71

policy/release/tasks.rego
lcarva: Wait, this is in the SLSA Provenance right? Not the resource from the cluster?

Also, it would be really good if you could share an example SLSA Provenance v1.0 generated by Chains as a gist(?) in this PR :pray:

policy/release/tasks.rego
lcarva: Yeah, https://github.com/enterprise-contract/ec-policies/pull/709#discussion_r1333468696 should address that.

policy/release/tasks_test.rego
lcarva: If it doesn't make sense in 1.0, then let's change it. The important thing is that we are always evaluating against the `ref.name`. The error message could display either. Which one is better is subjective.

policy/release/tasks.rego
joejstuart: This is a good question! I guess I was thinking one attestation per build, but I don't think there's any downside to this and we're protected in case it does happen.

policy/release/tasks.rego
joejstuart: And if it's a list, can there be multiple entries? idk, I'm going to look into this more.

lcarva: I think it should be ok if it this rule produces multiple results. It just means that whatever calls `_status` may also produce multiple results.

policy/release/tasks.rego
joejstuart: Interesting. If I switch the order and remove `not task.status.conditions`. The code doesn't work. I still need the `not`.

policy/release/tasks.rego
joejstuart: maybe it could have been simpler, but I ended up with this. The slsav0.2 status is not an array, but the v1.0 is and the way it was written it only returned one value no matter what.
https://github.com/enterprise-contract/ec-policies/pull/709/commits/e668fc09bf2f7f10da036d48694e6a85c76a108a

policy/release/tasks.rego
joejstuart: completely rewritten in https://github.com/enterprise-contract/ec-policies/pull/709#discussion_r1333462257

.regal/config.yaml
```diff
@@ -4,3 +4,7 @@ rules:
   style:
     todo-comment:
       level: warning
+    file-length:
+      level: warning
+    line-length:
+      max-line-length: 130
```
joejstuart, line 10: This conflicted with `make `fmt`, so increased it.

lcarva: ack. I can't tell where `opa fmt` sets this to 130 though.

Nit: it would be nice to have a separate commit for this.

policy/release/tasks.rego
```diff
@@ -195,11 +195,51 @@ current_required_tasks := task_data if {
...
-	status := task.status
-} else = "MISSING"
+	# Handle SLSA Provenance v0.2
+	task.status
+	not task.status.conditions
+	status := [s |
+		s := task.status
+	]
+
+	# if task.status = [], we want ["MISSING"] returned
```
lcarva, line 211: Missing test case to verify this.

lcarva (approved):

policy/release/tasks.rego
joejstuart: good catch. fixed.

PR #733, opened 2023-09-29
slsav1.0 refactor

joejstuart:
Refactor all policies that verify an attestation to work with slsav1.0.

- **IMPORTANT**: The task information here is pulled from the `content` field in `resolvedDependencies`. It is the `taskRun` that is stored there.
- This mostly changes libraries to support accessing v0.2 and v1.0 attestations and updating the policy tests to support v1.0.
- Update regal lint to warn on line-length too long. `opa fmt` in many cases is ok with a line-length that `regal lint` is not. So there is a conflict between the two.

https://issues.redhat.com/browse/HACBS-2303

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -7,7 +7,11 @@ rules:
     file-length:
       level: warning
     line-length:
+<<<<<<< HEAD
       max-line-length: 130
   idiomatic:
     no-defined-entrypoint:
       level: ignore
+=======
+      level: warning
+>>>>>>> c4d3317 (regal lint is in conflict with opa fmt on the line-length)
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -7,11 +7,7 @@ rules:
     file-length:
       level: warning
     line-length:
-<<<<<<< HEAD
-      max-line-length: 130
+      level: warning
   idiomatic:
     no-defined-entrypoint:
       level: ignore
-=======
-      level: warning
->>>>>>> c4d3317 (regal lint is in conflict with opa fmt on the line-length)
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -8,6 +8,8 @@ rules:
       level: warning
     line-length:
       level: warning
+    rule-length:
+      max-rule-length: 50
   idiomatic:
     no-defined-entrypoint:
       level: ignore
--- a/policy/lib/refs.rego
+++ b/policy/lib/refs.rego
@@ -37,7 +37,7 @@ task_ref(task) := i {
 		"url": _param(r, "url", ""),
 		"revision": _param(r, "revision", ""),
 		"pathInRepo": _param(r, "pathInRepo", ""),
-		"name":  object.get(r, "name", ""),
+		"name": object.get(r, "name", ""),
 		"kind": lower(object.get(r, "kind", "task")),
 	}
 } else = i {
--- a/policy/lib/refs.rego
+++ b/policy/lib/refs.rego
@@ -37,7 +37,7 @@ task_ref(task) := i {
 		"url": _param(r, "url", ""),
 		"revision": _param(r, "revision", ""),
 		"pathInRepo": _param(r, "pathInRepo", ""),
-		"name": _param(r, "pathInRepo", ""),
+		"name":  object.get(r, "name", ""),
 		"kind": lower(object.get(r, "kind", "task")),
 	}
 } else = i {
--- a/policy/lib/refs.rego
+++ b/policy/lib/refs.rego
@@ -38,6 +38,7 @@ task_ref(task) := i {
 		"revision": _param(r, "revision", ""),
 		"pathInRepo": _param(r, "pathInRepo", ""),
 		"name": _param(r, "pathInRepo", ""),
+		"kind": lower(object.get(r, "kind", "task")),
 	}
 } else = i {
 	# Handle local reference
--- a/policy/lib/refs.rego
+++ b/policy/lib/refs.rego
@@ -37,6 +37,7 @@ task_ref(task) := i {
 		"url": _param(r, "url", ""),
 		"revision": _param(r, "revision", ""),
 		"pathInRepo": _param(r, "pathInRepo", ""),
+		"name": _param(r, "pathInRepo", ""),
 	}
 } else = i {
 	# Handle local reference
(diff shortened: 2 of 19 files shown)


policy/lib/refs.rego
```diff
@@ -36,6 +36,8 @@ task_ref(task) := i {
 		"url": _param(r, "url", ""),
 		"revision": _param(r, "revision", ""),
 		"pathInRepo": _param(r, "pathInRepo", ""),
+		"name": _param(r, "pathInRepo", ""),
```
zregvart, line 39: I'm looking at a attestation with the git resolver, and the task has a `name` property, could we use that instead?

```suggestion
		"name": object.get(r, "name", ""),
```

joejstuart: Yes, good catch.

lcarva: Actually... `r` here should be the [taskRef](https://gist.github.com/zregvart/7cc23e92ba37333f70c07af7671cac5a#file-attestation-v1-json-L21). This is always going to be an empty string.

policy/lib/refs.rego
```diff
@@ -36,6 +36,8 @@ task_ref(task) := i {
 		"url": _param(r, "url", ""),
 		"revision": _param(r, "revision", ""),
 		"pathInRepo": _param(r, "pathInRepo", ""),
+		"name": object.get(r, "name", ""),
```
lcarva, line 39: Let's not do this.

policy/lib/refs_test.rego
```diff
@@ -73,6 +73,7 @@ test_slsav1_local_ref {
 
 test_git_resolver_in_slsav1_pipelinerun {
 	ref := {"spec": {"taskRef": {
+		"name": "git-clone",
```
lcarva, line 76: Yeah, see this not a valid taskRef AFAICT.

policy/lib/tekton/task.rego
```diff
@@ -126,14 +126,32 @@ _task_params(task) := params if {
 # task_param returns the value of the given parameter in the task.
 task_param(task, name) := _task_params(task)[name]
 
+# slsa v0.2 results
+_task_results(task) := task.results
+
+# slsa v1.0 results
+_task_results(task) := task.status.taskResults
```
lcarva, line 133: I don't think this is right. AFAIK, results will appear under `.predicate.runDetails.byproducts`, and I'm not sure if this already includes results from TaskRuns.

policy/lib/tekton/task.rego
```diff
@@ -126,14 +126,32 @@ _task_params(task) := params if {
...
 	result_name := _key_value(result, "name")
 	result_name == name
 	value := _key_value(result, "value")
 }
 
+# slsa v0.2 task steps
+task_steps(task) := task.steps
+
+# slsa v1.0 task steps
+task_steps(task) := task.status.taskSpec.steps
```
lcarva, line 147: I must be missing something. I don't see `status` for a task in v1.

policy/lib/refs.rego
joejstuart: This is actually the taskRef from the taskRun that's stored in resolvedDependencies. The information from externalParameters can change based on the pipeline type. inline, etc

policy/lib/tekton/task.rego
joejstuart: I think only the pipeline results are in byproducts. This is from the taskRun in resolvedDependencies.

policy/lib/tekton/task.rego
joejstuart: In the new buildType, the taskRun is stored under resolvedDependencies in the `content` field. This information is taken from there. Here is an example. https://gist.github.com/joejstuart/dce324211434b19990b50e2efe83f548#file-gistfile1-txt-L198

policy/lib/refs_test.rego
joejstuart: I don't have an example git resolver, but this [1] is what I'm referring to here. I think it's right? Although, I don't see a `name` in the taskRun example here [2]

[1] https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1/taskref_types.go#L20
[2] https://tekton.dev/docs/pipelines/git-resolver/

policy/lib/tekton/task.rego
```diff
@@ -126,14 +126,32 @@ _task_params(task) := params if {
...
+
+# slsa v0.2 step image
+task_step_image_ref(step) := step.environment.image
+
+# slsa v1.0 step image
+task_step_image_ref(step) := step.image
```
lcarva, line 153: Is it `image` or `imageID`?

policy/lib/tekton/task_test.rego
```diff
@@ -469,3 +484,118 @@ slsav1_attestation_local_spec := {
...
+_marshal_slsav1_results(results) := [r |
+	some result in results
+	r := {"name": result.name, "type": result.type, "value": json.marshal(result.value)}
+]
+
+resolved_dependencies(tasks) := [r |
+	some task in tasks
+	r := {
+		"name": "pipelineTask",
+		"content": json.marshal(task),
```
lcarva, line 599: We need to base64 encode this, right?

policy/lib/tekton/task.rego
joejstuart: imageID. Good catch!

policy/lib/tekton/task_test.rego
joejstuart: yes, good catch. that's a pretty big mistake. I'll fix.

policy/lib/tekton/task.rego
lcarva: We should update the tests as well.

.regal/config.yaml
```diff
@@ -7,7 +7,9 @@ rules:
     file-length:
       level: warning
     line-length:
-      max-line-length: 130
+      level: warning
+    rule-length:
+      max-rule-length: 50
```
lcarva, line 12: what?! :sweat_smile:

policy/lib/tekton/task.rego
```diff
@@ -56,7 +56,7 @@ _maybe_tasks(pipeline) := _tasks if {
 # handle tasks from a slsav1 attestation
 _maybe_tasks(slsav1) := _tasks if {
 	slsav1.predicate.buildDefinition
-	_tasks := {json.unmarshal(dep.content) |
+	_tasks := {json.unmarshal(base64.decode(dep.content)) |
```
lcarva, line 59: Let's do a follow up on this, but I think if this is not valid json, or valid base64, then it just crashes opa (and ec along the way).

.regal/config.yaml
lcarva: oh this is rule length, not line length.

lcarva (approved):

.regal/config.yaml
joejstuart: lol. The default is 30. Seemed just as arbitrary as 50.

PR #748, opened 2023-10-06
Fix and disable step_image_registires

lcarva:
Starting with version 0.17, Tekton Chains adds the `oci://` prefix to the image references for each task step in the SLSA Provenance.

This change modifies the step_image_registries policies to take this into account.

Ref: RHTAPBUGS-857

This also removes step_image_registries from every collection as we know it is currently broken.

Ref: HACBS-1972


joejstuart (approved):

robnester-rh (approved):
LGTM

PR #757, opened 2023-10-10
Get list of required tasks from build task labels

lcarva:
Ref: EC-53


zregvart (approved):

policy/lib/tekton/pipeline.rego
```diff
@@ -4,6 +4,7 @@ import future.keywords.contains
 import future.keywords.if
 import future.keywords.in
 
+import data.lib
 import data.lib.time
 
 pipeline_label := "pipelines.openshift.io/runtime"
```
joejstuart, line 10: The label in the task is `build.appstudio.redhat.com/build_type`. This should be updated to that. 

An example.
https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L6

lcarva: Thank you! In the process, should I remove support for looking at the labels on the slsa provenance for the pipeline, and only look at the labels for the task?

lcarva: I kept it since it's just used as a fallback.

lcarva:
Added a new set of required tasks named `docker` as that's what the buildah task uses. This is just a copy of the `generic` set.

I used this image to test: `quay.io/redhat-user-workloads-stage/lucarval-tenant/devfile-sample-code-with-quarkus-2/devfile-sample-code-with-quarkus-whez@sha256:7897148683f15816427059f7e2def81a61884e54fca68f98cd7808d4ae1bf138`

joejstuart (approved):

PR #765, opened 2023-10-16
Add script to generate attestations

lcarva:
This commit adds the record.sh script. It uses Tekton Chains to generate different versions of SLSA Provenance. The results are then saved. This is meant to assist in writing policies by providing real data as a reference point. It is a baby step towards acceptable tests.

Ref: EC-170


lcarva:
There's a discussion about whether or not this data should reside in its own repo. Looking for feedback on that idea.

integration/record.sh
```diff
@@ -0,0 +1,336 @@
+#!/bin/env bash
```
joejstuart, line 1: any reason you chose the integration directory instead of keeping the scripts under `hack/`?

integration/record.sh
```diff
@@ -0,0 +1,336 @@
...
+function rand_string() {
+    local size
+    local s
+
+    size="${1-5}"
+    set +o pipefail
+    s="$(tr -dc a-z </dev/urandom | head -c ${size})"
+    set -o pipefail
+    echo -n "${s}"
+}
```
zregvart, line 62: Bit simpler with `openssl rand -hex 5`

integration/record.sh
```diff
@@ -0,0 +1,336 @@
...
+#
+#   kubectl -n default create secret docker-registry $USER \
+#     --from-file=.dockerconfigjson=$HOME/.docker/config.json
+#   kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "'$USER'"}]}'
+#   kubectl patch serviceaccount default -p '{"secrets": [{"name": "'$USER'"}]}'
+#
+#
+# Usage: ./integration/record.sh
+#   The environment variable IMAGE_URL must point to an OCI repository, e.g. quay.io/lucarval/spam
+#   You must have write access to the repository.
```
zregvart, line 43: Could we spin up a local registry?

integration/record.sh
```diff
@@ -0,0 +1,336 @@
...
+    echo -n "${s}"
+}
+
+# Update the Chains configuration and restart the controller.
+# https://github.com/tektoncd/chains/blob/main/docs/config.md#chains-configuration-1
+function configure_chains() {
+    local data
+    data="${1}"
+    # Make sure Chains is setup how we want it.
+    oc patch -n tekton-chains configmap/chains-config --type merge --patch \
```
zregvart, line 70: Could we use `kubectl` instead of `oc`?

integration/record.sh
```diff
@@ -0,0 +1,336 @@
...
+  download_attestation "${image_digest}" > "${output_dir}/attestation.json"
+}
+
+#########
+# Setup #
+#########
+log "Using repository ${IMAGE_URL}"
+cd "$(git root)"
+# Just a quick check to make sure there is an active connection to the cluster.
+oc get serviceaccount > /dev/null
```
zregvart, line 266: `kubectl cluster-info` could work here as well, something like:
```suggestion
echo |kubectl cluster-info
```

zregvart (commented):
Looks good, does the job, left some nitpicks...

joejstuart:
> There's a discussion about whether or not this data should reside in its own repo. Looking for feedback on that idea.

My 2cts, I think this is a good place for them now. 

Just some things that came to mind, since they need a kind cluster and it would be good to use these in some sort of e2e test, maybe the generation of them is moved to the ec-cli acceptance tests at some point.

robnester-rh (approved):
lgtm, although I do like @zregvart's comments wrt using `kubectl` vs `oc` where applicable.

integration/record.sh
lcarva: I was thinking ahead of time and trying to eventually turn this into some sort of integration test. I'll rename it.

integration/record.sh
lcarva: Ideally, yes. But I don't know how to make that accessible from within the cluster by Chains. Probably not worth chasing that for this PR.

lcarva:
> > There's a discussion about whether or not this data should reside in its own repo. Looking for feedback on that idea.
> 
> My 2cts, I think this is a good place for them now.
> 
> Just some things that came to mind, since they need a kind cluster and it would be good to use these in some sort of e2e test, maybe the generation of them is moved to the ec-cli acceptance tests at some point.

I think given the size of the data this creates, and the fact that this repo gets cloned pretty often during policy evaluation, there's a benefit in keeping it small. I'll move it https://github.com/enterprise-contract/hacks.

Eventually, we should have acceptance tests in this repo, and it should have a more sophisticated way of generating this data dynamically. I don't know if those tests belong in ec-cli though.

lcarva:
Implemented the suggestions and merged the changes in the hacks repo: https://github.com/enterprise-contract/hacks/pull/37

Closing this.

PR #766, opened 2023-10-17
Handle partial input.image.source data

zregvart:
The code assumed if `input.image.source` was present that the data structure would always contain `url` and `revision` properties. Those might be omitted as they're user-controlled inputs, e.g. from the ApplicationSnapshot.
If those properties were omitted then the helper function `_expected_sources` would return empty set as it would try to operate on undefined data, and iterating over an empty set would lead to no checks being performed.
This uses `object.get` instead of direct access to return a default value if the properties are missing, so that the set should be populated and the checks should fail to match.

ref: https://issues.redhat.com/browse/RHTAPBUGS-883

### DIFF
--- a/policy/release/slsa_source_correlated.rego
+++ b/policy/release/slsa_source_correlated.rego
@@ -138,8 +138,8 @@ _expected_sources contains expected_source if {
 	some vcs_type, vcs_info in input.image.source
 
 	# e.g. git+https://github.com/...
-	expected_vcs_uri := sprintf("%s+%s", [vcs_type, vcs_info.url])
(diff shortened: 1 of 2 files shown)


zregvart:
I want to be sure this is safe to merge so the e2e tests don't fail if we have unpinned policy version in use somewhere. I'll convert this to draft and make modifications to the e2e tests before merging this just to be sure...

zregvart:
Let's wait for https://github.com/redhat-appstudio/e2e-tests/pull/829 to get merged

zregvart:
The change in e2e tests was merged, I think this should be risk free now.

PR #773, opened 2023-10-18
Ensure unknown tasks fail acceptable bundle checks

simonbaird:
The attestation_task_bundle.task_ref_bundles_acceptable check should produce a failure if the task bundle reference is not found at all in the available `data.task-bundles` data. Previously it was only failing if the repo was present in data.task-bundles and the task's digest wasn't in the list.

Also some extra test coverage is added to exercise the different failure conditions.

Resolves: https://issues.redhat.com/browse/EC-206

### DIFF
--- a/policy/lib/bundles.rego
+++ b/policy/lib/bundles.rego
@@ -90,7 +90,8 @@ bundle(task) := refs.task_ref(task).bundle
 # _collection returns an array representing the full list of records to
 # be taken into consideration when evaluating policy rules for bundle
 # references. Any irrelevant records are filtered out from the array.
+# (The else condition is for when data["task-bundles"][ref.repo] doesn't exist.)
 _collection(ref) := items {
 	full_collection := data["task-bundles"][ref.repo]
 	items := time_lib.acceptable_items(full_collection)
-}
+} else := []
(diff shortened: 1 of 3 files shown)


policy/lib/bundles.rego
```diff
@@ -93,4 +93,7 @@ bundle(task) := refs.task_ref(task).bundle
 _collection(ref) := items {
 	full_collection := data["task-bundles"][ref.repo]
 	items := time_lib.acceptable_items(full_collection)
+} else := items {
+	# In case data["task-bundles"][ref.repo] doesn't exist
+	items := []
 }
```
simonbaird, line 99: Could write it like this, wdyt? I thought the longer one was more readable.. :shrug: .

```suggestion
} else := [] # In case data["task-bundles"][ref.repo] doesn't exist
```

policy/pipeline/task_bundle_test.rego
```diff
@@ -42,7 +42,7 @@ test_bundle_unpinned {
 test_bundle_reference_valid {
 	tasks := [{
 		"name": "my-task",
-		"taskRef": {"bundle": "quay.io/redhat-appstudio/hacbs-templates-bundle:latest@sha256:abc"},
+		"taskRef": {"bundle": "reg.com/repo:latest@sha256:abc"},
```
simonbaird, line 45: Required to make existing tests pass.

policy/lib/bundles.rego
zregvart: Not really important... I prefer the shorter version

zregvart (approved):

PR #774, opened 2023-10-19
Delete Red Hat data

lcarva:
Ref: EC-149


lcarva:
Needs a bit more work in `hack/update-bundles.sh`. I also need to figure out how to make this not break the antora extension when building the website.

lcarva:
Made this change to make it easier to validate changes: https://github.com/enterprise-contract/ec-policies/pull/775

lcarva:
Also need to update the symlinks for attachments to point to the example data instead. And, update some of the text on ec-cli docs accordingly.

zregvart (approved):

PR #787, opened 2023-10-31
Consider tags when looking for out of date bundles

zregvart:
When looking for out of date bundles now tags are considered as well, so the newer bundle needs to be for the same tag, not for the same exact reference (assuming that the reference also contains the digest) and it needs to be newer than the bundle used (newer effective_on).

There are a number of assumptions in this:
 - the order of bundles in acceptable bundles is by descending `effective_on`; every newer bundle is listed before the examined task bundle
 - bundles in acceptable bundles and the task bundle have tags, and the tags are used to denote versions (e.g. v1, v2) or are unique for each version (e.g. if git id is used it is for the last change of the definitions for the files that are in the bundle, not for the whole repository)

reference: EC-223


policy/lib/bundles.rego
```diff
@@ -32,9 +32,23 @@ out_of_date_task_bundle(tasks) := {task |
...
+	b.tag == match_bundle.tag
+
+	# it is not the same reference
+	not is_equal(b, ref)
+
+	# and it is newer
+	time.parse_rfc3339_ns(b.effective_on) > time.parse_rfc3339_ns(match_bundle.effective_on)
 }
 
 # Returns a subset of tasks that do not use an acceptable bundle reference.
```
lcarva, line 54: Do we need to add some special handling in `unacceptable_task_bundle` and `is_acceptable` as well? If there's enough churn for tag A, tag B may get buried down the list and trigger a violation when used.

Consider making the changes in `_collection` so it covers the different cases:
```rego
_collection(ref) := items {
	full_collection := data["task-bundles"][ref.repo]
    tag_collections := _collection_by_tag(full_collection, tag)
	items := time_lib.acceptable_items(full_collection)
} else := []

_collection_by_tag(all_items, tag) := items {
   items := [item |
     some item in items
     item.tag == tag
   ]
   count(items) > 0
} else := all_items
```

zregvart: Not sure about changing `_collection`, the `tag` then must be passed in somehow...

My brain is a bit slow today. I think it might cover (most/all?) cases if I add a `is_equal` variant that matches on both tag and digest.

Pushed [1fb7ef1](https://github.com/enterprise-contract/ec-policies/pull/787/commits/1fb7ef1831484b4f0fdc298a6811021b501f1081) to demonstrate the effects.

I'm not 100% sure of this...

lcarva: > Not sure about changing _collection, the tag then must be passed in somehow...

Isn't the tag an attribute of `ref`, i.e. `ref.tag`? (I didn't do that in my example code above)

lcarva: The `is_equal` approach may work, but it may have its own challenges. For example, we want to take the tag into account only if at least one record in the collection has a matching tag. Otherwise, we want to consider the whole collection as is. This may be challenging to do when comparing each record in the collection individually.

policy/lib/bundles_test.rego
```diff
@@ -112,22 +113,22 @@ test_is_equal {
 task_bundles := {"reg.com/repo": [
```
lcarva, line 113: We probably want to use a data set that contains different tag values.

zregvart:
@lcarva this is now a completely different take, the code has no semblance to the previous version. I've restarted the work from `main` and pushed to the same branch the previous code was on.

policy/lib/bundles.rego
```diff
@@ -15,83 +18,165 @@ disallowed_task_reference(tasks) := {task |
...
 	some task in tasks
-	bundle(task) == ""
+	refs.task_ref(task).bundle == ""
 }
 
 # Returns a subset of tasks that use bundle references not pinned to a digest.
 unpinned_task_bundle(tasks) := {task |
 	some task in tasks
-	ref := image.parse(bundle(task))
+	ref := image.parse(_bundle_ref(task, data["task-bundles"]))
```
lcarva, line 27: This function probably doesn't need to use `_bundle_ref` since it only cares about the digests.

policy/lib/bundles.rego
```diff
@@ -15,83 +18,165 @@ disallowed_task_reference(tasks) := {task |
...
 
-# Returns if the required task-bundles data is missing
-default missing_task_bundles_data := false
+_is_unacceptable(ref) {
+	not _is_acceptable(ref)
+}
 
-missing_task_bundles_data {
-	count(data["task-bundles"]) == 0
+_is_unacceptable(ref) {
```
lcarva, line 61: It's a bit confusing the `_is_unacceptable` is not just the inverse of `_is_acceptable`. Maybe a misnomer? 

Looking at the definition of `_is_acceptable`, I think we should include the `_is_expired` there.

policy/lib/bundles.rego
```diff
@@ -15,83 +18,165 @@ disallowed_task_reference(tasks) := {task |
...
+	_newer_version_exists(ref, data["task-bundles"])
```
lcarva, line 62: Why does having a newer version make it unacceptable? Unless it is not expired, it completely fine to use - meaning it is acceptable.

`_newer_version_exists` should be used in the context of "it's fine to use this, but be mindful that there is an update available". That to me seems to fall under the unacceptable category. It's just an informational warning (is that a thing?!) to the user.

policy/lib/bundles.rego
```diff
@@ -15,83 +18,165 @@ disallowed_task_reference(tasks) := {task |
...
 }
 
-# Returns whether or not the ref matches the digest of the record.
-is_equal(record, ref) := match {
-	ref.digest != ""
-	match := record.digest == ref.digest
+# Out of date references are those that are acceptable, meaning that their
+# reference is recorded in acceptable task bundles data and no newer version
+# exists or the reference has expired
+_is_out_of_date(ref) {
```
lcarva, line 82: This makes sense to me. The implementation clearly defines what `_is_out_of_date` means :+1:

policy/lib/bundles.rego
```diff
@@ -15,83 +18,165 @@ disallowed_task_reference(tasks) := {task |
...
-# in case the digest is blank for the ref. This is a weaker comparison as,
-# unlike digests, tags are not immutable entities. It is expected that a
-# missing digest results in a warning whenever possible.
-is_equal(record, ref) := match {
-	ref.digest == ""
-	match := record.tag == ref.tag
+_is_out_of_date(ref) {
+	_is_acceptable(ref)
+	not _newer_version_exists(ref, data["task-bundles"])
+	_is_expired(ref, data["task-bundles"])
```
lcarva, line 91: Hmm.. If it's expired, it should be treated as if it's not even in the list. In this case, it's not out of date, it's unacceptable.

policy/lib/bundles.rego
```diff
@@ -15,83 +18,165 @@ disallowed_task_reference(tasks) := {task |
...
 }
 
-bundle(task) := refs.task_ref(task).bundle
+# Evaluates to true if the tasks bundle reference is found in the acceptable
+# task bundles data, matched by digest, and it is not in effect, meaning that
+# it's effective_on is in the future, i.e. it has expired
```
lcarva, line 96: "in the past"?

policy/lib/bundles.rego
```diff
@@ -15,83 +18,165 @@ disallowed_task_reference(tasks) := {task |
...
+	records := acceptable[ref.repo]
+
+	some record in records
+
+	# consider all records, if a match is found via exact digest and it's
+	# effective_on is greater than current time in effect (remember it is
+	# configurable) we deem the bundle reference out of date
+	record.digest == ref.digest
+
+	time_lib.effective_current_time_ns > time.parse_rfc3339_ns(record.effective_on)
```
lcarva, line 108: What if it's the latest entry but its effective_on was yesterday? That shouldn't be counted as expired.

policy/lib/bundles.rego
```diff
@@ -15,83 +18,165 @@ disallowed_task_reference(tasks) := {task |
...
+	some record in records
+
+	# consider all records, if a match is found via exact digest and there
+	# exists a newer record for the same tag but it is newer, i.e. has greater
+	# effective_on value
+	record.digest == ref.digest
+
+	some other in records
+
+	record.tag == other.tag
```
lcarva, line 129: We should be careful here. If a tag doesn't match, we should take into account all the records in the list.

simonbaird:
I don't understand this well, but I like the more towards making it more readable.

policy/lib/bundles.rego
zregvart: In that case the `_is_expired` as used in `_is_unacceptable` (the branch where
a record for the bundle does exist) will not be
evaluated because `_newer_version_exists` will be first to evaluate to `false`

policy/lib/bundles.rego
zregvart: This is used from `out_of_date_task_bundle` to create a list of Task bundle
references that are out of date, in this case there is no newer version, so we
would not emit a warning that an out of date bundle was used, with our
definition of what acceptable bundle is, in such case there would not be a
violation or a warning; reminder: this is a branch were the record for the task
bundle exists and it is the latest version i.e. it is acceptable, but it has a
effective on in the past i.e. it has expired

policy/lib/bundles.rego
zregvart: Remember it's a logical AND here, so to be unacceptable a newer version has to
exist and the version used needs to have expired.

policy/lib/bundles.rego
zregvart: I think I should rename `_is_acceptable` to `_record_exists`, it is confusing this way.

policy/lib/bundles.rego
lcarva: Yup, you're right.

policy/lib/bundles.rego
lcarva: Those are the three scenarios I can think of. The first one is probably the one that is tripping me. We probably need some work to make sure that behaves properly.

Also, I think the tests need to be enhanced to explicitly check those scenarios.

zregvart: I've added more tests and a specialization of `_newer_version_exists` when there are no same value tags.

policy/lib/bundles.rego
```diff
@@ -26,73 +29,188 @@ unpinned_task_bundle(tasks) := {task |
...
+# Determine the image reference of the task bundle, if the provided task bundle
+# image reference doesn't have the tag within it try to lookup the tag from the
+# acceptable task bundles data
+_bundle_ref(task, _) := ref if {
+	ref := bundle(task)
+	img := image.parse(ref)
+	img.tag != ""
+}
+
+_bundle_ref(task, acceptable) := ref if {
```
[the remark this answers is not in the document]
zregvart: Functions take the data as arguments, rules access the data directly. This is in alignment with https://docs.styra.com/regal/rules/style/external-reference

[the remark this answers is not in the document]
lcarva: Aha! That makes sense. There are quite a few functions that do hard code to data, e.g. `_is_expired`, `_newer_version_exists`, `_record_exists`.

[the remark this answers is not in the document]
zregvart: Yeah, I noticed that after I wrote the comment so I pushed a bit of that, still refactoring this but I'll keep this in mind...

PR #788, opened 2023-10-31
Support appending to data lists

simonbaird:
The goal is to allow users to append to rule data without overwriting the original rule data. This can be used to add a custom task to data.task-bundles for example.


simonbaird:
Move this to draft since the task-bundles is actually a map not an array, which I had forgotten about.

PR #789, opened 2023-10-31
Support _one of_ for required tasks

zregvart:
This adds support for the notion of _one of_ required tasks from the required tasks list.

For example, given required tasks data as:

```yaml
 - A
 - [B, C, D]
 - E
```

Tasks "A", at least one of "B", "C" or "D", and "E" are required.

### DIFF
--- a/policy/release/tasks.rego
+++ b/policy/release/tasks.rego
@@ -78,7 +78,7 @@ deny contains result if {
 #   in the PipelineRun attestation.
 # custom:
 #   short_name: required_tasks_found
-#   failure_msg: Required task %q is missing
+#   failure_msg: '%s is missing'
 #   solution: >-
 #     Make sure all required tasks are in the build pipeline. The required task list
 #     is contained as xref:ec-cli:ROOT:configuration.adoc#_data_sources[data] under the key 'required-tasks'.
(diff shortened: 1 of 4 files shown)


zregvart:
Created as draft, support for the pipeline rules needs to be added.

policy/release/tasks.rego
```diff
@@ -169,10 +169,31 @@ _missing_tasks(required_tasks) := {task |
...
 
-	some task in required_tasks
-	not task in tkn.tasks_names(att)
+	some required_task in required_tasks
+	some task in _any_missing(required_task, tkn.tasks_names(att))
 }
 
+_any_missing(required, tasks) := missing if {
+	# one of required tasks is required
+	is_array(required)
```
lcarva, line 178: Could use a helper method to avoid having to implement the logic for both cases, e.g.
```rego
_any_missing(required, tasks) := missing if {
  req := setfy(required)
  tsk := setfy(tasks)
  ...
}

setfy(o) := s if {
  s := {v | some v in o}
} else := {o}
```

This could maybe reside in `policy/lib/set_helpers.rego`.

lcarva (approved):

policy/release/tasks.rego
zregvart: Turns out we already have that helper there :)

simonbaird:
Nice.

policy/release/tasks.rego
lcarva: I think it would've been nice to unify the two use cases (array vs non-array) which is what I was trying to drive at with this comment. But anyways, the current approach works just as well :wink:

PR #792, opened 2023-11-07
Use `ec opa` as drop-in replacement for `opa`

zregvart:
Given https://github.com/enterprise-contract/ec-cli/issues/1071 we can now use `ec opa` that includes the required custom Rego builtin functions.

This depends on https://github.com/enterprise-contract/ec-cli/pull/1149

reference: EC-242


zregvart:
Draft until https://github.com/enterprise-contract/ec-cli/pull/1149 is merged and the `replace` can be removed from `go.mod`.

go.mod
```diff
@@ -1,10 +1,12 @@
 module github.com/enterprise-contract/ec-policies
 
-go 1.21
+go 1.21.2
+
+toolchain go1.21.3
 
 require (
+	github.com/enterprise-contract/ec-cli v0.0.0-20231107103908-d7e137f0b777
```
lcarva, line 8: dependabot won't keep these up to date. (That's probably ok, just calling it out)

lcarva (approved):

go.mod
zregvart: Seems that Renovate [will](https://github.com/renovatebot/renovate/blob/f9d1cade0804a36f40bd0afb327fe4aee4c69023/lib/modules/datasource/go/index.spec.ts#L163)

PR #793, opened 2023-11-07
Add support for informative tests

lcarva:
Some tests are required to be executed, but their results do not have to be successful. They are meant to inform the user that their results need some attention, but they are non-blocking. An example of such test is a certain code scanner which may produce a high number of false positives. These tests are referred to as informative.

This commit modifies the `test` package to more easily support these types of tests.

Also, since `test.no_skipped_tests` has been modified to emit a violation instead of a warning, an effective date to a month from now has been added. This should allow users to gradually address any issues before they become blocking.

Ref: EC-224

### DIFF
--- a/policy/release/test.rego
+++ b/policy/release/test.rego
@@ -91,18 +91,17 @@ deny contains result if {
 }
 
 # METADATA
-# title: All required tests passed
+# title: No tests failed
 # description: >-
-#   Confirm that all the tests in the test results
-#   have a successful result. A successful result is one that isn't a
-#   "FAILURE" or "ERROR". This will fail if any of the tests failed and
-#   the failure message will list the names of the failing tests.
+#   Produce a violation if any non-informative tests have their result set to "FAILED".
+#   The result type is configurable by the "failed_tests_results" key, and the list
+#   of informative tests is configurable by the "informative_tests" key in the rule data.
 # custom:
-#   short_name: required_tests_passed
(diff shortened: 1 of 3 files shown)


policy/release/test.rego
```diff
@@ -91,32 +91,79 @@ deny contains result if {
...
-#   "FAILURE" or "ERROR". This will fail if any of the tests failed and
-#   the failure message will list the names of the failing tests.
+#   Produce a violation if any non-informative tests have their result set to "FAILED".
+#   The result type is configurable by the "failed_tests_results" key, and the list
+#   of informative tests is configurable by the "informative_tests" key in the rule data.
 # custom:
-#   short_name: required_tests_passed
-#   failure_msg: "Test %q did not complete successfully"
+#   short_name: no_failed_tests
+#   failure_msg: "Test %q failed"
```
zregvart, line 101: Observation: I was wondering if we'd need a transition period for anyone addressing the `required_tests_passed` rule, I guess not really but it is an interesting question that we might want to consider. What happens when we remove a rule? Do we have any compatibility guarantees that we need to uphold?

zregvart (approved):

policy/release/test.rego
lcarva: Yeah. It's an interesting one. It would be interesting to see if we can provide aliases for policy rules. Filed https://github.com/enterprise-contract/ec-policies/issues/795 as a follow up.

PR #796, opened 2023-11-09
Do not rely on build task label for FBC images

zregvart:
When computing the key to select the set of required tasks we look at the built task's or the pipeline label. Given that the build task for the FBC and the Docker builds are the same, i.e. have the same label. We cannot rely on the task's label, instead we this uses the image label `operators.operatorframework.io.index.configs.v1` to classify a attestation as being an attestation of a FBC image, and subsequently select the set of required tasks for the `fbc` pipeline.

reference: EC-85

### DIFF
--- a/policy/lib/tekton/pipeline.rego
+++ b/policy/lib/tekton/pipeline.rego
@@ -30,9 +30,13 @@ required_task_list(pipeline) := pipeline_data if {
 # pipeline_label_selector is a specialized function that returns the name of the
 # required tasks list that should be used.
 pipeline_label_selector(pipeline) := value if {
+	not is_fbc # given that the build task is shared between fbc and docker builds we can't rely on the task's label
+
 	# Labels of the build Task from the SLSA Provenance v1.0 of a PipelineRun
 	value := build_task(pipeline).metadata.labels[task_label]
 } else := value if {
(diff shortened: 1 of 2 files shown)


policy/lib/tekton/pipeline.rego
```diff
@@ -44,6 +48,17 @@ pipeline_label_selector(pipeline) := value if {
 } else := value if {
 	# Labels from a Tekton Pipeline definition
 	value := pipeline.metadata.labels[pipeline_label]
+} else := value if {
```
joejstuart, line 51: nit: is this last part needed? the fbc pipeline contains the label `pipelines.openshift.io/runtime` so the above two should find it if it's a valid fbc pipeline.

joejstuart (approved):

lcarva (approved):

policy/lib/tekton/pipeline.rego
zregvart: Perhaps, I wasn't sure myself. It helps for very custom pipelines, though it might be an overhead that we don't need right now?

joejstuart: Maybe just leave it. fbc seems like a special case.

Issue #746, opened 2023-10-04
Create policy rules to process an SBOM


lcarva:
# Description

Once https://github.com/enterprise-contract/ec-cli/issues/1070 and https://github.com/enterprise-contract/ec-cli/issues/1071 are completed, let's introduce some policy rules that use the newly added rego function to assist in processing a CycloneDX SBOM. See [example](https://gist.github.com/lcarva/94da611c0aa5b358821cd469ac384918) policy rule.

The policy rules should be relatively basic for this story. Consider something similar to what was done for [SPDX SBOMs](https://enterprisecontract.dev/docs/ec-policies/release_policy.html#sbom_spdx_package).

As a stretch goal, consider writing the policy rules in such a way that the SBOM could be fetched from the SLSA Provenance, or from the list of attestations.

# Acceptance Criteria

- Policy rules exist that can process the CycloneDX SBOM referenced in the SLSA Provenance (, or as loaded file from the image).

PR #807, opened 2023-11-20
Add policy rules for CycloneDX SBOM

lcarva:
Resolves: #746


policy/release/sbom_cyclonedx.rego
```diff
@@ -0,0 +1,79 @@
...
+
+	# https://cyclonedx.org/specification/overview/#recognized-predicate-type
+	statement.predicateType == "https://cyclonedx.org/bom"
+	sbom := statement.predicate
+]
+
+# _is_valid is true if the given CycloneDX SBOM has certain fields. This is not an exhaustive schema
+# check. It mostly ensures the fields used by the policy rules in this package have been set.
+_is_valid(sbom) if {
+	is_array(sbom.components)
```
zregvart, line 78: Perhaps for future refinement we could use [`json.match_schema`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-object-jsonmatch_schema) here

zregvart (approved):

policy/release/sbom_cyclonedx.rego
lcarva: Great suggestion! I added some initial support but filed https://github.com/enterprise-contract/ec-policies/issues/808 as a follow up.

lcarva:
> Very nice. Really clean and nice rego. Lgtm.
> 
> One thing I was wondering, should some of these rules be included in a collection, perhaps the redhat collection or perhaps minimal?

Yes! Added to both.

policy/release/sbom_cyclonedx_schema.rego
zregvart, line 1: :heart:

PR #818, opened 2023-11-29
Tekton slsa refactor wip

simonbaird:


simonbaird:
I'm gonna put this on hold, with the optimistic idea that I'll come back to it in the future. I think there are some useful ideas here.

Some notes for when/if this gets picked up again:

* There is nice neat code for v1.0 and v0.2 tekton tasks in SLSA attestations, but nothing for tekton tasks in pipeline definitions. Often we reuse the same logic for all three of those, hence without some nice neat code for "tasks in a pipeline definition" we're left with new and old code mixed together, which is not really an improvement. I think it's fixable, but I didn't want to invest more time.
* There is too much repeated code: for the attributes that are derived from other attributes, the code should be shared rather than copy pasted. I figure there should be a nice way to do it, but didn't actually figure out what that is yet.
* There are a lot of tests that are sensitive to the task format, and try to test both formats. Ideally we can decouple that so only the logic is tested, not the task parsing. I think it's doable, but I was finding it very slow going.

PR #821, opened 2023-12-06
Use oras to build OPA bundles

### LINKED ISSUE #817: Allow pushing bundles with custom rego function
We [tried pushing](https://github.com/enterprise-contract/ec-policies/pull/815#pullrequestreview-1754935597) an OCI bundle that contains policy rules that use a custom rego function (`ec.purl.parse`). It was not successful: 

```
2023/11/29 12:01:33 pushing bundle to: quay.io/zregvart_redhat/ec-beta-policy:git-23b1223
Error: push bundle: pushing layers: load: loading policies: get compiler: 2 errors occurred:
policy/beta/packages.rego:39: rego_type_error: undefined function ec.purl.parse
policy/beta/packages.rego:42: rego_type_error: undefined function ec.purl.parse
exit status 1
````

That is because the bundle is being created by executing `conftest push`. The conftest CLI doesn't know anything about the ec custom rego functions.

We do have `ec opa` which wraps the opa cli to avoid this. However, we don't have a wrapper for conftest itself. We could add one. Or we could try switching from conftest bundles to opa bundles.

**Acceptance Criteria**

* It is possible to push OCI bundles containing policy rules that use ec's custom rego functions.

### PR DESCRIPTION
**Pull request by lcarva**
The previous approach of using `conftest push` does not work now that EC provides custom rego functions. We could wrap the `conftest` command in EC CLI, like we did for OPA via `ec opa`. However, that is a bit heavy handed.

The OPA docs[1] suggest using `opa bundle` to generate a tarball with the compiled policy rules. Then, use `oras push` to upload the tarball to an OCI registry.

The problem with the approach above is that when the OCI image containing the policy rules is fetched, the output is a tarball instead of the expected rego files. This causes conftest to not see the policy rules. I considered adding support in the EC CLI to handle this use case by simply expanding any tarballs that may have been fetched. But... this means older version of the EC CLI won't be able to consume the newer policy bundles.

Instead, this commit opts to use `oras push` directly on the rego files. This creates a policy bundle that works as expected by the EC CLI.

There is benefit in running the `opa build` command since it compiles the policy rules and verifies their integrity. For example, it checks if the policy rules use an unknown function. This is useful. So, as a sanity check, we now run `ec opa bundle` to ensure integrity, but the output is discarded.

[1] https://www.openpolicyagent.org/docs/latest/management-bundles/#building-and-publishing-policy-containers

Fixes #817

### DIFF
--- a/hack/update-bundles.sh
+++ b/hack/update-bundles.sh
@@ -24,8 +24,9 @@ set -o nounset
 
 REPO_PREFIX="${REPO_PREFIX-quay.io/enterprise-contract/}"
 ROOT_DIR=$( git rev-parse --show-toplevel )
-BUNDLES="release pipeline task build_task"
-CONFTEST="go run github.com/open-policy-agent/conftest"
+BUNDLES="beta release pipeline task build_task"
+OPA="go run github.com/enterprise-contract/ec-cli opa"
+ORAS="go run oras.land/oras/cmd/oras"
 
 # For example:
 #   ENSURE_UNIQUE=1 DRY_RUN=1 hack/update-bundles.sh
@@ -49,10 +50,6 @@ function exclusions() {
 
 function repo_name() {
   echo "ec-$1-policy"
-}
-
-function conftest_push() {
-  $DRY_RUN_ECHO ${CONFTEST} push --policy policy $2
 }
 
 function ensure_unique_file() {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: hack/update-bundles.sh [hunk 1]
```diff
@@ -119,16 +116,13 @@ for b in $BUNDLES; do
...
-
-    # Add OCI annotations to the bundle image
-    tmp_oci_dir="$(mktemp -d --tmpdir)"
-    tmp_oci_dirs+=("${tmp_oci_dir}")
-    skopeo copy docker://"$push_repo:$tag" dir:"${tmp_oci_dir}" ${skopeo_cp_args}
-    manifest="$(jq -c '. += { "annotations": { "org.opencontainers.image.revision": "'"${last_update_sha}"'" } }' "${tmp_oci_dir}/manifest.json")"
-    echo "${manifest}" > "${tmp_oci_dir}/manifest.json"
-    $DRY_RUN_ECHO skopeo copy dir:"${tmp_oci_dir}" docker://"$push_repo:$tag" ${skopeo_cp_args}
+    ${ORAS} push "$push_repo:$tag" ${src_dirs} \
+      --annotation "org.opencontainers.image.revision=${last_update_sha}"
```
simonbaird, line 125: Removing the manual annotation add is a nice bonus!

zregvart:
> But... this means older version of the EC CLI won't be able to consume the newer policy bundles.

Could we distinguish those two cases via media type, and publish both variants for a bit? We could benefit from having a single layer, fetching multiple layers (one for each file) takes a bit longer...

lcarva:
> > But... this means older version of the EC CLI won't be able to consume the newer policy bundles.
> 
> Could we distinguish those two cases via media type, and publish both variants for a bit? We could benefit from having a single layer, fetching multiple layers (one for each file) takes a bit longer...

We certainly can! But I don't think we should. Besides the additional overhead of maintaining two versions of the bundle (and having to figure out which one was used when a user reports an issue), having a single layer is not necessarily better for performance. As with most things it depends. For instance, if the registry storage backends is something like S3 storage (like quay.io is), then downloading any one blob is capped at a certain value. But if you have two blobs those happen in parallel, doubling the overall speed. In any case, I think that's very much entering the realm of premature optimization. Downloading policy bundles from an OCI registry is, anecdotally, far from the slowest piece in EC.

PR #826, opened 2023-12-12
Give everyone a few more weeks to get a synk token

simonbaird:
The snyk requirement graduated from a warning to a violation a few days ago and a some teams are struggling to get their snyk tokens working enabled and working.

Not sure if we want to merge this, but I'll create a PR in case we decide to do it.

### DIFF
--- a/policy/release/test.rego
+++ b/policy/release/test.rego
@@ -175,7 +175,7 @@ deny contains result if {
 #   - redhat
 #   depends_on:
 #   - test.test_data_found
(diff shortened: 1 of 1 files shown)


lcarva (approved):
Approving so it can be merged right away if we decide to.

simonbaird:
Seems like everyone figured out their tokens without too much trouble. At least I haven't heard otherwise. Let's abandon.

PR #827, opened 2023-12-13
Add validation to rule data

lcarva:
This commit enhances the validation performed on rule data. It defines a JSON Schema for each value and provides a policy rules to execute them.

Ref: EC-132

### DIFF
--- a/policy/beta/packages.rego
+++ b/policy/beta/packages.rego
@@ -31,8 +31,26 @@ import data.lib.sbom
 deny contains result if {
 	some s in sbom.cyclonedx_sboms
 	some component in s.components
-	_contains(component.purl, lib.rule_data("disallowed_packages"))
+	_contains(component.purl, lib.rule_data(_rule_data_key))
 	result := lib.result_helper(rego.metadata.chain(), [component.purl])
+}
+
+# METADATA
+# title: Disallowed packages list is provided
+# description: >-
+#   Confirm the `disallowed_packages` rule data was provided, since it is required by the policy
+#   rules in this package.
+# custom:
+#   short_name: disallowed_packages_provided
+#   failure_msg: "%s"
+#   solution: Provide a list of disallowed packages in the expected format.
+#   collections:
+#   - redhat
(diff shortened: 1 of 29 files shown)


policy/release/collection/rule-data-validation.rego
```diff
@@ -0,0 +1,6 @@
+#
+# METADATA
+# title: rule-data-validation
```
zregvart, line 3: Perhaps something shorter like:

```suggestion
# title: policy-data
```

policy/beta/packages.rego
```diff
@@ -66,3 +84,60 @@ _matches_version(version, matcher) if {
 } else := false
 
 _to_semver(v) := trim_prefix(v, "v")
+
+# Verify disallowed_packages is a non-empty list of objects
```
zregvart, line 88: The list is allowed to be empty
```suggestion
# Verify disallowed_packages is an array of objects with expected properties
```

policy/beta/packages.rego
```diff
@@ -66,3 +84,60 @@ _matches_version(version, matcher) if {
...
+	pkg.format in {"semver", "semverv"}
+	some attr in ["min", "max"]
+
+	version := _to_semver(object.get(pkg, attr, ""))
+	version != ""
+
+	not semver.is_valid(version)
+
+	msg := sprintf(
+		"Item at index %d in %s does not have a valid %s semver value: %s",
```
zregvart, line 138: Perhaps we should quote the user provided value

```suggestion
		"Item at index %d in %s does not have a valid %s semver value: %q",
```

policy/beta/packages.rego
lcarva: Thanks!

policy/release/collection/rule-data-validation.rego
lcarva: Decided to go with `policy_data` since that's the case rego usually expects.

zregvart (approved):
:shipit:

PR #830, opened 2023-12-15
Add policy rules for features annotations in OLM

lcarva:
Ref: https://issues.redhat.com/browse/EC-270

### DIFF
--- a/policy/lib/rule_data.rego
+++ b/policy/lib/rule_data.rego
@@ -67,6 +67,19 @@ rule_data_defaults := {
 		"gitBlob",
 		"gitTag",
 	],
+	# Used in release/olm.rego
+	"required_olm_features_annotations": [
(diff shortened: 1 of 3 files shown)


policy/lib/rule_data.rego
```diff
@@ -67,6 +67,19 @@ rule_data_defaults := {
 		"gitBlob",
 		"gitTag",
 	],
+	# Used in release/olm.rego
+	"required_olm_features_annotations": [
+		"features.operators.openshift.io/disconnected",
```
lcarva, line 72: I opted to define these here instead of defaulting to an empty list. In either case, it is possible to overwrite these values with a data source.

I don't have a good technical reason for not adding them here, so I just went with the easy path. I'm happy to change this if others feel strongly otherwise.

zregvart (approved):

PR #841, opened 2023-12-21
Switch to using the pipeline task name instead of the taskRef name

joejstuart:
The taskRef name can be excluded if a task is defined inline, but the pipeline task name is always required.

This change only affects the message on a failed rule.

https://issues.redhat.com/browse/RHTAPBUGS-1076

### DIFF
--- a/policy/release/buildah_build_task.rego
+++ b/policy/release/buildah_build_task.rego
@@ -20,7 +20,8 @@ import data.lib
 #   the buildah task.
 # custom:
 #   short_name: buildah_task_has_dockerfile_param
-#   failure_msg: 'The DOCKERFILE param was not included in the buildah task(s): %q'
+#   failure_msg: >-
+#     The pipeline task %q does not contain the DOCKERFILE param. This is a requirement for the underlying task %q
 #   solution: >-
 #     Make sure the buildah task has a parameter named 'DOCKERFILE'.
 #   collections:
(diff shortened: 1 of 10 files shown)


policy/release/buildah_build_task.rego
```diff
@@ -33,7 +33,7 @@ deny contains result if {
 	buildah_tasks
 	some buildah_task in buildah_tasks
 	not lib.tkn.task_param(buildah_task, "DOCKERFILE")
-	name := lib.tkn.task_name(buildah_task)
+	name := lib.tkn.pipeline_task_name(buildah_task)
 	result := lib.result_helper_with_term(rego.metadata.chain(), [name], name)
```
lcarva, line 37: I just want to clarify that `pipeline_task_name` should only ever be used to provide additional context to the user as in an error message, for example. It should never be used when asserting that a particular Task was provided.

This is important because the Pipeline Task name is a property of the Pipeline definition, while a Task name is a property of the Task definition.

I agree that in this line, the error message should likely include the Pipeline Task name. But the `term` (third parameter) should continue to be the Task name. Besides the reason I mentioned above, switching the term would likely also require a change to the `include`/`exclude` options in existing policy configs.

Here, I would consider an error message that specifies both the Pipeline Task and the Task names while keeping the `term` as the Task name.

joejstuart: Good point. I updated the error message to include the pipeline task and the underlying task name.

zregvart (approved):

Issue #816, opened 2023-11-29
Move beta.packages to release.sbom_cyclonedx


lcarva:
In #732, new policy rules were added that use the newly added `ec.purl.parse` custom rego function. To avoid issues with older versions of the ec-cli that do not provide the function, they were added to a new namespace, `policy/beta`, instead of the usual `policy/release`. 

This issue is about moving those policy rules to `policy/release`, likely under the `sbom_cyclonedx` package.

Let's aim for doing this right after January 1st as that should be sufficient time to allow clients to update.

**Acceptance Criteria**

* The policy rules from `beta.packages` are moved under the `release` namespace.

zregvart:
The `ec.purl.parse` needs at least version [2269](https://github.com/enterprise-contract/ec-cli/pull/1173/commits/5211e454cb0f59d6eb1db99b794559c051234c9f) in the last 7 days the oldest version was 2340 and in the last 30 days the oldest version was 2286. I think it is now safe to do this.

PR #847, opened 2024-01-04
Move beta.packages to release.sbom_cyclonedx

zregvart:
resolves: #816


policy/release/sbom_cyclonedx.rego
```diff
@@ -68,3 +68,130 @@ deny contains result if {
...
+# description: >-
+#   Confirm the CycloneDX SBOM contains only allowed packages. By default all packages are allowed.
+#   Use the "disallowed_packages" rule data key to provide a list of disallowed packages.
+# custom:
+#   short_name: allowed
+#   failure_msg: "Package is not allowed: %s"
+#   solution: >-
+#     Update the image to not use a disallowed package.
+#   collections:
+#   - redhat
```
joejstuart, line 83: what happens if there is no sbom and the `Redhat` collection` is used?

simonbaird (approved):

policy/release/sbom_cyclonedx.rego
zregvart: The [sbom_cyclonedx.found](https://enterprisecontract.dev/docs/ec-policies/release_policy.html#sbom_cyclonedx__found) rule in the current redhat collection already fails if it is missing.

PR #848, opened 2024-01-04
if a future required task throws a warning, report the date when that task is required

joejstuart:
https://issues.redhat.com/browse/RHTAPBUGS-1079

### DIFF
--- a/policy/lib/tekton/task.rego
+++ b/policy/lib/tekton/task.rego
@@ -15,14 +15,10 @@ missing_required_tasks_data if {
 
 # The latest set of required tasks. Tasks here are not required right now
 # but will be required in the future.
-latest_required_default_tasks contains task if {
-	some task in ectime.newest(data["required-tasks"]).tasks
-}
+latest_required_default_tasks := ectime.newest(data["required-tasks"])
 
 # The set of required tasks that are required right now.
-current_required_default_tasks contains task if {
-	some task in ectime.most_current(data["required-tasks"]).tasks
-}
+current_required_default_tasks := ectime.most_current(data["required-tasks"])
 
 # tasks returns the set of tasks found in the object.
 tasks(obj) := {task |
--- a/policy/pipeline/required_tasks.rego
+++ b/policy/pipeline/required_tasks.rego
@@ -60,10 +60,10 @@ deny contains result if {
 	count(tkn.tasks(input)) > 0
 
 	# Get missing tasks by comparing with the default required task list
-	some required_task in _missing_tasks(current_required_tasks)
+	some required_task in _missing_tasks(current_required_tasks.tasks)
 
 	# Don't report an error if a task is required now, but not in the future
-	required_task in latest_required_tasks
+	required_task in latest_required_tasks.tasks
 	result := lib.result_helper_with_term(rego.metadata.chain(), [_format_missing(required_task, false)], required_task)
 }
 
@@ -74,17 +74,21 @@ deny contains result if {
 #   is not currently included in the Pipeline definition.
 # custom:
 #   short_name: missing_future_required_task
-#   failure_msg: '%s is missing and will be required in the future'
+#   failure_msg: '%s is missing and will be required on %s'
 warn contains result if {
 	count(tkn.tasks(input)) > 0
 
 	# Get missing tasks by comparing with the default required task list
-	some required_task in _missing_tasks(latest_required_tasks)
+	some required_task in _missing_tasks(latest_required_tasks.tasks)
 
 	# If the required_task is also part of the current_required_tasks, do
 	# not proceed with a warning since that's clearly a violation.
-	not required_task in current_required_tasks
-	result := lib.result_helper_with_term(rego.metadata.chain(), [_format_missing(required_task, true)], required_task)
+	not required_task in current_required_tasks.tasks
+	result := lib.result_helper_with_term(
+		rego.metadata.chain(),
+		[_format_missing(required_task, true), latest_required_tasks.effective_on],
+		required_task,
+	)
 }
 
 # METADATA
@@ -137,6 +141,8 @@ _any_missing(required, tasks) := missing if {
 
 # get the future tasks that are pipeline specific. If none exists
 # get the default list
+default latest_required_tasks := {"tasks": []}
+
 latest_required_tasks := task_data if {
 	task_data := tkn.latest_required_pipeline_tasks(input)
 } else := task_data if {
@@ -145,6 +151,8 @@ latest_required_tasks := task_data if {
 
 # get the current tasks that are pipeline specific. If none exists
 # get the default list
+default current_required_tasks := {"tasks": []}
+
 current_required_tasks := task_data if {
 	task_data := tkn.current_required_pipeline_tasks(input)
 } else := task_data if {
--- a/policy/release/tasks.rego
+++ b/policy/release/tasks.rego
@@ -92,10 +92,10 @@ deny contains result if {
 #   - tasks.pipeline_has_tasks
 #
 deny contains result if {
-	some required_task in _missing_tasks(current_required_tasks)
+	some required_task in _missing_tasks(current_required_tasks.tasks)
 
 	# Don't report an error if a task is required now, but not in the future
-	required_task in latest_required_tasks
+	required_task in latest_required_tasks.tasks
 	result := lib.result_helper_with_term(rego.metadata.chain(), [_format_missing(required_task, false)], required_task)
 }
 
@@ -119,7 +119,7 @@ warn contains result if {
 
 	# only tasks that are unacceptable
 	some unacceptable_task in bundles.unacceptable_task_bundle(tkn.tasks(att))
-	some missing_required_name in _missing_tasks(current_required_tasks)
+	some missing_required_name in _missing_tasks(current_required_tasks.tasks)
 	some unacceptable_task_name in tkn.task_names(unacceptable_task)
 
 	unacceptable_task_name == missing_required_name
@@ -156,7 +156,7 @@ warn contains result if {
 #   was not included in the PipelineRun attestation.
 # custom:
 #   short_name: future_required_tasks_found
-#   failure_msg: '%s is missing and will be required in the future'
+#   failure_msg: '%s is missing and will be required on %s'
 #   solution: >-
 #     There is a task that will be required at a future date that is missing
 #     from the build pipeline.
@@ -166,12 +166,16 @@ warn contains result if {
 #   - tasks.pipeline_has_tasks
 #
 warn contains result if {
-	some required_task in _missing_tasks(latest_required_tasks)
+	some required_task in _missing_tasks(latest_required_tasks.tasks)
 
 	# If the required_task is also part of the current_required_tasks, do
 	# not proceed with a warning since that's clearly a violation.
-	not required_task in current_required_tasks
-	result := lib.result_helper_with_term(rego.metadata.chain(), [_format_missing(required_task, true)], required_task)
+	not required_task in current_required_tasks.tasks
+	result := lib.result_helper_with_term(
+		rego.metadata.chain(),
+		[_format_missing(required_task, true), latest_required_tasks.effective_on],
+		required_task,
+	)
 }
 
 # METADATA
@@ -241,6 +245,8 @@ _any_missing(required, tasks) := missing if {
 
 # get the future tasks that are pipeline specific. If none exists
 # get the default list
+default latest_required_tasks := {"tasks": []}
+
 latest_required_tasks := task_data if {
 	some att in lib.pipelinerun_attestations
 	count(tkn.tasks(att)) > 0
@@ -251,6 +257,8 @@ latest_required_tasks := task_data if {
 
 # get current required tasks. fall back to the default list if
 # no label exists in the attestation
+default current_required_tasks := {"tasks": []}
+
 current_required_tasks := task_data if {
 	some att in lib.pipelinerun_attestations
 	count(tkn.tasks(att)) > 0
--- a/policy/release/tasks_test.rego
+++ b/policy/release/tasks_test.rego
@@ -429,12 +429,12 @@ test_future_one_of_required_tasks_missing if {
 	expected := {
 		{
 			"code": "tasks.future_required_tasks_found",
-			"msg": `One of "c1", "c2", "c3" tasks is missing and will be required in the future`,
+			"msg": `One of "c1", "c2", "c3" tasks is missing and will be required on 2099-01-02T00:00:00Z`,
 			"term": ["c1", "c2", "c3"],
 		},
 		{
 			"code": "tasks.future_required_tasks_found",
-			"msg": `One of "d1", "d3" tasks is missing and will be required in the future`,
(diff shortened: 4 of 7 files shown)


policy/release/tasks.rego
```diff
@@ -135,14 +135,30 @@ warn contains result if {
...
 warn contains result if {
-	some required_task in _missing_tasks(latest_required_tasks)
-
-	# If the required_task is also part of the current_required_tasks, do
-	# not proceed with a warning since that's clearly a violation.
-	not required_task in current_required_tasks
-	result := lib.result_helper_with_term(rego.metadata.chain(), [_format_missing(required_task, true)], required_task)
+	some required_task in _missing_tasks(tkn.get_tasks(latest_required_tasks))
+	# # If the required_task is also part of the current_required_tasks, do
+	# # not proceed with a warning since that's clearly a violation.
```
simonbaird, line 141: The extra # chars look odd here.

policy/release/tasks.rego
```diff
@@ -135,14 +135,30 @@ warn contains result if {
...
+# 	not latest_required_tasks.effective_on
+# }
+
+# deny contains result if {
+# 	not current_required_tasks.tasks
+# }
+
+# deny contains result if {
+# 	not current_required_tasks.effective_on
+# }
```
simonbaird: Is this intentionally still here? Maybe take it out unless there's a good reason for it.

policy/release/tasks_test.rego
```diff
@@ -510,7 +510,7 @@ _missing_tasks_warning(tasks) := {warning |
 	some task in tasks
 	warning := {
 		"code": "tasks.future_required_tasks_found",
-		"msg": sprintf("Task %q is missing and will be required in the future", [task]),
+		"msg": sprintf("Task %q is missing and will be required in the future on 2099-01-02T00:00:00Z", [task]),
```
simonbaird, line 513: Just an idea, perhaps we can shorten it to "will be required on 2099-01-02T00:00:00Z".

policy/lib/tekton/task.rego
```diff
@@ -21,16 +21,12 @@ latest_required_default_tasks := ectime.newest(data["required-tasks"])
...
 
 # retrieve the tasks from the required task list
-get_tasks(required_tasks) := tasks if {
-	tasks := [task | 
-		some task in required_tasks.tasks
-	]
-}
+required_tasks(required_tasks) := [task |
+	some task in required_tasks.tasks
+]
```
simonbaird, line 26: It seems more significant than just lint fixes. How about squashing this one in with the previous commit?

simonbaird (commented):
I'd like to see shorter commit subjects and some squashing of the lint fix commit as mentioned.

policy/lib/tekton/task.rego
joejstuart: yeah, the regal linter caught this. I'll squash them. I left the changes as their own commit to show they weren't part of the main change.

policy/lib/tekton/task.rego
```diff
@@ -15,14 +15,18 @@ missing_required_tasks_data if {
...
 # The set of required tasks that are required right now.
-current_required_default_tasks contains task if {
-	some task in ectime.most_current(data["required-tasks"]).tasks
-}
+current_required_default_tasks := ectime.most_current(data["required-tasks"])
+
+# retrieve the tasks from the required task list
+required_tasks(required_tasks) := [task |
+	some task in required_tasks.tasks
+]
```
zregvart, line 26: (nit) My mind sees infinite recursion here, perhaps it would help to refactor this to something like:

```suggestion
required_tasks(given) := [task |
	some task in given.tasks
]
```

policy/pipeline/required_tasks.rego
```diff
@@ -73,17 +73,22 @@ deny contains result if {
...
 	count(tkn.tasks(input)) > 0
 
 	# Get missing tasks by comparing with the default required task list
-	some required_task in _missing_tasks(latest_required_tasks)
-
-	# If the required_task is also part of the current_required_tasks, do
-	# not proceed with a warning since that's clearly a violation.
-	not required_task in current_required_tasks
-	result := lib.result_helper_with_term(rego.metadata.chain(), [_format_missing(required_task, true)], required_task)
+	# this is empty
```
zregvart, line 81: Doesn't fit with the comment above in my mind

```suggestion
```

policy/pipeline/required_tasks.rego
```diff
@@ -73,17 +73,22 @@ deny contains result if {
...
+	some required_task in _missing_tasks(tkn.required_tasks(latest_required_tasks))
+
+	# # If the required_task is also part of the current_required_tasks, do
+	# # not proceed with a warning since that's clearly a violation.
```
zregvart, line 85:
```suggestion
	# If the required_task is also part of the current_required_tasks, do
	# not proceed with a warning since that's clearly a violation.
```

zregvart (commented):
Some nitpicks, otherwise LGTM

policy/lib/tekton/task.rego
```diff
@@ -15,14 +15,18 @@ missing_required_tasks_data if {
...
+latest_required_default_tasks := ectime.newest(data["required-tasks"])
 
 # The set of required tasks that are required right now.
-current_required_default_tasks contains task if {
-	some task in ectime.most_current(data["required-tasks"]).tasks
-}
+current_required_default_tasks := ectime.most_current(data["required-tasks"])
+
+# retrieve the tasks from the required task list
+required_tasks(given) := [task |
```
lcarva, line 24: NIT: This has nothing to do with required tasks. The name and the doc are both misleading. It also makes its usage quite strange: `_missing_tasks(tkn.required_tasks(current_required_tasks))`

Why can't the above just be: `_missing_tasks(current_required_tasks.tasks)`

Still strange but maybe a bit better as it doesn't require knowing about a helper method? Also... why is `current_required_tasks` not a list of tasks already? This makes me think our understanding of the data model is wrong.

And why does `current_required_tasks` have a default value of `[]`? Shouldn't it be an object with that contains a `tasks` attribute?

joejstuart: > NIT: This has nothing to do with required tasks. The name and the doc are both misleading. It also makes its usage quite strange: _missing_tasks(tkn.required_tasks(current_required_tasks))

The name is awkward. It get the required tasks from the required task list. I initially had `get_required_tasks`, but regal didn't like the `get_`.

>Why can't the above just be: _missing_tasks(current_required_tasks.tasks)

We could remove the helper method from that statement, but the statement below required a list be returned and if the field `.tasks` isn't present it failed some of the test cases.
https://github.com/enterprise-contract/ec-policies/pull/848/files#diff-41af58b3ce642141d04cdaf73f9031a7c467e94271057bc518f8af6c369d03fdR85

>Still strange but maybe a bit better as it doesn't require knowing about a helper method? Also... why is current_required_tasks not a list of tasks already? This makes me think our understanding of the data model is wrong.

This change required retrieving the `effective_on` time from the required task list so it could be provided in a warning message. So instead of having a list of tasks it's now `[tasks: [], effective_on: string]` The helper methods are used to get the tasks or the effective_on.

>And why does current_required_tasks have a default value of []? Shouldn't it be an object with that contains a tasks attribute?

Yes, but if there's a failure `undefined` is returned and that caused an issue with the rules. Before this change `tkn.current_required_default_tasks` always returned a list.
https://github.com/enterprise-contract/ec-policies/pull/848/files#diff-01bed933c73cc61844dc5cc9a2c9a6217ff88b8dd0d953545261352355f10d3cL23
`current_required_default_tasks` now returns `{tasks: [], effective_on: string}` 

I do think this is all more complicated than it should be and I can try and see what can be done about it, but the required task policies are inherently more complicated because of the date logic.

joejstuart: > NIT: This has nothing to do with required tasks. The name and the doc are both misleading. It also makes its usage quite strange: _missing_tasks(tkn.required_tasks(current_required_tasks))

The name is awkward. It get the required tasks from the required task list. I initially had `get_required_tasks`, but regal didn't like the `get_`.

>Why can't the above just be: _missing_tasks(current_required_tasks.tasks)

We could remove the helper method from that statement, but the statement below required a list be returned and if the field `.tasks` isn't present it failed some of the test cases.
https://github.com/enterprise-contract/ec-policies/pull/848/files#diff-41af58b3ce642141d04cdaf73f9031a7c467e94271057bc518f8af6c369d03fdR85

>Still strange but maybe a bit better as it doesn't require knowing about a helper method? Also... why is current_required_tasks not a list of tasks already? This makes me think our understanding of the data model is wrong.

This change required retrieving the `effective_on` time from the required task list so it could be provided in a warning message. So instead of having a list of tasks it's now `[tasks: [], effective_on: string]` The helper methods are used to get the tasks or the effective_on.

>And why does current_required_tasks have a default value of []? Shouldn't it be an object with that contains a tasks attribute?

Yes, but if there's a failure `undefined` is returned and that caused an issue with the rules. Before this change `tkn.current_required_default_tasks` always returned a list.
https://github.com/enterprise-contract/ec-policies/pull/848/files#diff-01bed933c73cc61844dc5cc9a2c9a6217ff88b8dd0d953545261352355f10d3cL23
`current_required_default_tasks` now returns `{tasks: [], effective_on: string}` 

I do think this is all more complicated than it should be and I can try and see what can be done about it, but the required task policies are inherently more complicated because of the date logic.

lcarva: I now realize this is more than nitpicking on a function name. :sweat_smile: 

There's a problem here because the `current_required_tasks` variable, for example, has a default value of an empty array, but when it is used in the policy rules, its value is expected to be an object.

This is likely why you saw other tests failing when trying to change things. Anything relying on the default value of that variable is just not going to work anymore.

By wrapping usage of the variable into a function, i.e. `tkn.required_tasks`, we avoid this problem because that function, as it is structured, returns an empty list if all fails.

The same applies to the `latest_required_tasks` variable.

The default value for a variable should, at a bare minimum, match the type of the non-default values. I really think we should address this before merging.

I wrote a patch that addresses that concern and does not break the tests.

<details>
<summary>view patch</summary>

```diff
diff --git a/policy/lib/tekton/task.rego b/policy/lib/tekton/task.rego
index 2533824..2bdfa3d 100644
--- a/policy/lib/tekton/task.rego
+++ b/policy/lib/tekton/task.rego
@@ -20,11 +20,6 @@ latest_required_default_tasks := ectime.newest(data["required-tasks"])
 # The set of required tasks that are required right now.
 current_required_default_tasks := ectime.most_current(data["required-tasks"])
 
-# retrieve the tasks from the required task list
-required_tasks(given) := [task |
-	some task in given.tasks
-]
-
 # retrieve the effective_on from a required task list
 required_effective_on(given) := given.effective_on
 
diff --git a/policy/pipeline/required_tasks.rego b/policy/pipeline/required_tasks.rego
index 4fc4af5..411848a 100644
--- a/policy/pipeline/required_tasks.rego
+++ b/policy/pipeline/required_tasks.rego
@@ -59,10 +59,10 @@ deny contains result if {
 	count(tkn.tasks(input)) > 0
 
 	# Get missing tasks by comparing with the default required task list
-	some required_task in _missing_tasks(tkn.required_tasks(current_required_tasks))
+	some required_task in _missing_tasks(current_required_tasks.tasks)
 
 	# Don't report an error if a task is required now, but not in the future
-	required_task in tkn.required_tasks(latest_required_tasks)
+	required_task in latest_required_tasks.tasks
 	result := lib.result_helper_with_term(rego.metadata.chain(), [_format_missing(required_task, false)], required_task)
 }
 
@@ -78,11 +78,11 @@ warn contains result if {
 	count(tkn.tasks(input)) > 0
 
 	# Get missing tasks by comparing with the default required task list
-	some required_task in _missing_tasks(tkn.required_tasks(latest_required_tasks))
+	some required_task in _missing_tasks(latest_required_tasks.tasks)
 
 	# If the required_task is also part of the current_required_tasks, do
 	# not proceed with a warning since that's clearly a violation.
-	not required_task in tkn.required_tasks(current_required_tasks)
+	not required_task in current_required_tasks.tasks
 	result := lib.result_helper_with_term(
 		rego.metadata.chain(),
 		[_format_missing(required_task, true), tkn.required_effective_on(latest_required_tasks)],
@@ -134,7 +134,7 @@ _any_missing(required, tasks) := missing if {
 
 # get the future tasks that are pipeline specific. If none exists
 # get the default list
-default latest_required_tasks := []
+default latest_required_tasks := {"tasks": []}
 
 latest_required_tasks := task_data if {
 	task_data := tkn.latest_required_pipeline_tasks(input)
@@ -144,7 +144,7 @@ latest_required_tasks := task_data if {
 
 # get the current tasks that are pipeline specific. If none exists
 # get the default list
-default current_required_tasks := []
+default current_required_tasks := {"tasks": []}
 
 current_required_tasks := task_data if {
 	task_data := tkn.current_required_pipeline_tasks(input)
diff --git a/policy/release/tasks.rego b/policy/release/tasks.rego
index 61384bf..a2fe7f2 100644
--- a/policy/release/tasks.rego
+++ b/policy/release/tasks.rego
@@ -91,10 +91,10 @@ deny contains result if {
 #   - tasks.pipeline_has_tasks
 #
 deny contains result if {
-	some required_task in _missing_tasks(tkn.required_tasks(current_required_tasks))
+	some required_task in _missing_tasks(current_required_tasks.tasks)
 
 	# Don't report an error if a task is required now, but not in the future
-	required_task in tkn.required_tasks(latest_required_tasks)
+	required_task in latest_required_tasks.tasks
 	result := lib.result_helper_with_term(rego.metadata.chain(), [_format_missing(required_task, false)], required_task)
 }
 
@@ -135,11 +135,11 @@ warn contains result if {
 #   - tasks.pipeline_has_tasks
 #
 warn contains result if {
-	some required_task in _missing_tasks(tkn.required_tasks(latest_required_tasks))
+	some required_task in _missing_tasks(latest_required_tasks.tasks)
 
 	# If the required_task is also part of the current_required_tasks, do
 	# not proceed with a warning since that's clearly a violation.
-	not required_task in tkn.required_tasks(current_required_tasks)
+	not required_task in current_required_tasks.tasks
 	result := lib.result_helper_with_term(
 		rego.metadata.chain(),
 		[_format_missing(required_task, true), tkn.required_effective_on(latest_required_tasks)],
@@ -203,7 +203,7 @@ _any_missing(required, tasks) := missing if {
 
 # get the future tasks that are pipeline specific. If none exists
 # get the default list
-default latest_required_tasks := []
+default latest_required_tasks := {"tasks": []}
 
 latest_required_tasks := task_data if {
 	some att in lib.pipelinerun_attestations
@@ -215,7 +215,7 @@ latest_required_tasks := task_data if {
 
 # get current required tasks. fall back to the default list if
 # no label exists in the attestation
-default current_required_tasks := []
+default current_required_tasks := {"tasks": []}
 
 current_required_tasks := task_data if {
 	some att in lib.pipelinerun_attestations

```

</details>

joejstuart: Thanks, @lcarva. This makes things simpler.

lcarva (approved):
Very nice! :ok_hand:

lcarva:
@joejstuart, can you squash the commits before merging?

zregvart (approved):
LGTM

PR #849, opened 2024-01-05
Required task must be from acceptable bundle

zregvart:
Adds an additional check that a record must exist for a bundle the required task originates from. Meaning that the task is from an acceptable bundle.

reference: EC-254

### DIFF
--- a/policy/lib/bundles.rego
+++ b/policy/lib/bundles.rego
@@ -57,6 +57,13 @@ unacceptable_task_bundle(tasks) := {task |
 	_is_unacceptable(ref)
 }
 
+# Returns if the task uses an acceptable bundle reference
+is_acceptable_task(task) if {
+	ref := image.parse(_bundle_ref(task, _task_bundles))
+
+	_record_exists(ref)
+}
+
 _is_unacceptable(ref) if {
 	not _record_exists(ref)
 }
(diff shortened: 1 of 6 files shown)


joejstuart:
Do we need another rule that says "required task used, but not from an acceptable bundle"? In this case the user should get a violation from the acceptable bundles policy, but will the solution be obvious? And what if the acceptable bundle policy is excluded?

policy/lib/bundles.rego
```diff
@@ -57,6 +57,13 @@ unacceptable_task_bundle(tasks) := {task |
 	_is_unacceptable(ref)
 }
 
+# Returns if the task uses an acceptable bundle reference
+is_acceptable(task) if {
```
lcarva, line 61: NIT: Could we rename this to `is_acceptable_task` to make it clear that this is not an alternative to `not _is_unacceptable`?  Also, I wonder if this could be implemented like this instead:

```
is_acceptable_task(task) if {
  ref := image.parse(_bundle_ref(task, _task_bundles))
  not _is_unacceptable(ref)
}
```

`_is_unacceptable` takes into account `_newer_in_effect_version_exists` which I think should also be done here.

zregvart: I opted not to check if the task is latest as `task_ref_bundles_current` will check that already, checking that here again would cause two rule violations for out of date task bundles.

joejstuart (approved):

lcarva (approved):

lcarva:
We got two approvals here. I'm going to merge it now as I want to submit a PR that touches some of these files as well.

policy/lib/bundles.rego
zregvart: Valid point, let's examine this in https://github.com/enterprise-contract/ec-policies/issues/858

PR #850, opened 2024-01-08
Use image digest parsed from the reference

zregvart:
EC CLI will set the `input.image.ref` to the image reference, so the digest for comparison needs to be parsed from that string. Note that the digest when parsed, if not present in the image reference will be set to `""`.

### DIFF
--- a/policy/release/slsa_build_scripted_build_test.rego
+++ b/policy/release/slsa_build_scripted_build_test.rego
@@ -23,7 +23,7 @@ test_all_good if {
 		"steps": [{"entrypoint": "/bin/bash"}],
 	}]
 
-	image := {"ref": _image_digest}
+	image := {"ref": _image_ref}
 
 	task_bundles := {mock_bundle_repo: [{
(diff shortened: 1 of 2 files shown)


policy/release/slsa_build_scripted_build_test.rego
```diff
@@ -413,6 +413,8 @@ _image_digest_value := "123"
 
 _image_digest := concat(":", [_image_digest_algorithm, _image_digest_value])
 
+_image_ref := sprintf("%s@%s:%s", [_image_url, _image_digest_algorithm, _image_digest_value])
```
simonbaird, line 416: maybe:

```suggestion
_image_ref := sprintf("%s@%s", [_image_url, _image_digest])
```

simonbaird (approved):

lcarva (approved):
I had intended for this to be the initial behavior... Thanks for fixing this!

lcarva:
Merging as this is likely breaking users.

PR #856, opened 2024-01-09
Enforce pinning for additional types of Task refs

lcarva:
Ref: EC-257

This change adds a new attribute to the result returned by the
`task_ref` helper function, `pinned`. This indicates whether or not the
an immuntable reference is used.

For example, a git commit ID when using the git resolver is considered a
pinned reference, while a branch name is not.

A previous approach was to introduce a new helper method, e.g.
`is_pinned`. However, the a couple of things became obvious:

* The `task_ref` helper function already has the needed context, e.g.
  it knows that it's dealing with a git reference, so it can check if
  that's a git commit ID or a branch name.
* Being pinned is a property of the Task reference. It makes sense for
  the representation of the Task reference (what `task_ref` returns) to
  include this information.

As part of this, I also re-structured the unit tests in
`refs_test.rego`. Hopefully, this makes them easier to maintain.

This commit came out larger than I expected, but I can't think of a
reasonable way to split it up further.

### DIFF
--- a/policy/lib/refs.rego
+++ b/policy/lib/refs.rego
@@ -17,29 +17,45 @@ import future.keywords.in
 task_ref(task) := i if {
 	# Handle old-style bundle reference
 	r := _ref(task)
+	bundle := r.bundle
 	i := {
-		"bundle": r.bundle,
+		"bundle": bundle,
 		"name": _ref_name(task),
 		"kind": lower(object.get(r, "kind", "task")),
+		"pinned": _has_digest(bundle),
 	}
 } else := i if {
 	# Handle bundle-resolver reference
 	r := _ref(task)
 	r.resolver == "bundles"
+	bundle := _param(r, "bundle", "")
 	i := {
-		"bundle": _param(r, "bundle", ""),
+		"bundle": bundle,
 		"name": _ref_name(task),
 		"kind": lower(_param(r, "kind", "task")),
+		"pinned": _has_digest(bundle),
 	}
 } else := i if {
 	r := _ref(task)
 	r.resolver == "git"
+	revision := _param(r, "revision", "")
 	i := {
 		"url": _param(r, "url", ""),
-		"revision": _param(r, "revision", ""),
+		"revision": revision,
 		"pathInRepo": _param(r, "pathInRepo", ""),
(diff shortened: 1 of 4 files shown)


joejstuart (approved):

policy/lib/refs.rego
```diff
@@ -41,6 +41,15 @@ task_ref(task) := i if {
 		"name": object.get(r, "name", ""),
 		"kind": lower(object.get(r, "kind", "task")),
 	}
+} else := i if {
+	# Handle inlined Task definitions
+	_ref(task) == {}
+	i := {
+		# The Task definition itself is inlined without a name. Use a special value here to
+		# distinguish from other reference types.
+		"name": "<INLINE>",
```
zregvart, line 50: Perhaps pull the `"<INLINE>"` into a constant?

policy/release/tasks.rego
```diff
@@ -197,6 +198,30 @@ deny contains result if {
...
+#   - redhat
+#   depends_on:
+#   - tasks.pipeline_has_tasks
+#
+deny contains result if {
+	some att in lib.pipelinerun_attestations
+	some task in tkn.tasks(att)
+	ref := refs.task_ref(task)
+	not refs.is_pinned(ref)
+	result := lib.result_helper(rego.metadata.chain(), [tkn.task_name(task), tkn.pipeline_task_name(task)])
```
zregvart, line 222: Should this be:
```suggestion
	result := lib.result_helper_with_term(rego.metadata.chain(), [tkn.task_name(task), tkn.pipeline_task_name(task)], tkn.task_name(task))
```

zregvart (commented):
LGTM, couple of nitpicks...

policy/lib/refs.rego
lcarva: Good suggestion. I changed this to `<NAMELESS>` instead, but did throw that into a constant.

policy/release/tasks.rego
lcarva: Great suggestion. Done

lcarva:
I did quite a bit of rewriting here (sorry!). Please have another look :pray:

zregvart (approved):
Nice

PR #859, opened 2024-01-10
Use labels to determine task name

lcarva:


lcarva:
Putting it in draft for now. I'm gonna try doing this in refs.rego instead.

lcarva:
Ready now. I think this is a better approach as it consolidates the logic into a single module.

joejstuart (approved):

zregvart:
:+1:

PR #863, opened 2024-01-15
Consolidate pruning of stale acceptable bundles

lcarva:
The list of acceptable bundles is time based. The presence of a record is not a sufficient indicator for it being valid. As time goes on and the effective_on date of records is reached, older entries fall off the list. Using those records is akin to using a record that is not on the list at all.

This commit introduces a helper function to prune old records from the list before the list is consumed. This consolidates the decision of which records should be ignored in a single place.

This fixes the related bug by ensuring stale records are used consistently across the code base.

It also removes/modifies certain tests which were asserting incorrect behavior.

Ref: RHTAPBUGS-1099

### DIFF
--- a/policy/lib/bundles_test.rego
+++ b/policy/lib/bundles_test.rego
@@ -133,94 +133,6 @@ test_missing_required_data if {
 	lib.assert_equal(bundles.missing_task_bundles_data, true)
 }
 
-test_newer_in_effect_version_exists_not_using_tags_newest if {
-	ref := image.parse("registry.io/repository/image:tag@sha256:digest")
-	acceptable := {"registry.io/repository/image": [{
-		"digest": "sha256:digest",
-		"tag": "",
-		"effective_on": "2262-04-11T00:00:00Z",
-	}]}
-	not bundles._newer_in_effect_version_exists(ref) with data["task-bundles"] as acceptable
-}
-
-test_newer_in_effect_version_exists_not_using_tags_older if {
-	ref := image.parse("registry.io/repository/image:tag@sha256:digest")
-	acceptable := {"registry.io/repository/image": [
-		{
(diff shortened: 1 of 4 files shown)


policy/pipeline/task_bundle_test.rego
```diff
@@ -74,10 +74,7 @@ test_acceptable_bundle_up_to_date_maintained_version if {
 
 # Warn about out of date bundles that are still acceptable.
 test_acceptable_bundle_out_of_date_past if {
-	tasks := [
-		{"name": "my-task-1", "taskRef": {"bundle": "reg.com/repo@sha256:bcd"}},
-		{"name": "my-task-2", "taskRef": {"bundle": "reg.com/repo@sha256:cde"}},
```
lcarva, line 79: Given the list of `task_bundles`, `cde` is older than `bcd`. Usage of `cde` should cause a violation. Removed it here as this was clearly the wrong behavior.

policy/lib/bundles_test.rego
```diff
@@ -133,94 +133,6 @@ test_missing_required_data if {
 	lib.assert_equal(bundles.missing_task_bundles_data, true)
 }
 
-test_newer_in_effect_version_exists_not_using_tags_newest if {
```
lcarva, line 136: These tests were removed because it tests an "internal" function that no longer exists. I believe there is sufficient coverage elsewhere for the intended behavior.

simonbaird (approved):
Seems to be removing a lot of confusing logic and replacing it with a smaller amount of slightly less confusing logic. So lgtm.

zregvart (approved):

zregvart:
Awesome, best part is the removal of erroneous tests. I'm going to merge this and trigger update to infra-deployments.

PR #864, opened 2024-01-16
Add initial acceptance tests

lcarva:
These tests run most of the policy rules against a pre-recorded policy-input from the golden-container image.
    
The test utilities are very minimal. Just enough was implement to executed a single test scenario.
    
It is expected that the `acceptance` module will grow over time and will have a more formal structure.

The ec-cli dependency is updated because the previous version didn't have the `ec validate input` command.
    
Ref: EC-333


acceptance/features/release.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Golden Container Image
+
+    Scenario: Red Hat collection
+        Given there is a sample golden-container policy input
+        And there is a policy config
```
zregvart, line 5: This makes it easier to read in my mind. And allows the `"golden-container"` to be a parameter.

```suggestion
        Given sample policy input "golden-container"
          And a policy config:
```

acceptance/features/release.feature
```diff
@@ -0,0 +1,33 @@
...
+                                "sbom_cyclonedx.found",
+                                "redhat_manifests.redhat_manifests_missing",
+                                "cve.deprecated_cve_result_name"
+                            ]
+                        }
+                    }
+                ]
+            }
+            """
+        When input is validated with policy config
```
zregvart, line 31:
```suggestion
        When input is validated
```

Or like this, and then ditch the `And there is a policy config` line above:

```suggestion
        When input is validated with policy config:
```

acceptance/acceptance_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+const (
+	policyInputFilename  = "input.json"
+	policyConfigFilename = "policy.json"
+)
+
+var (
+	//go:embed samples/policy-input-golden-container.json
+	sampleGCPolicyInput string
+)
```
zregvart, line 34: This could allow us to have more than one sample:

```suggestion
//go:embed samples/*.json
var samples embed.FS
```

acceptance/acceptance_test.go
```diff
@@ -0,0 +1,275 @@
...
+	return ts, nil
+}
+
+func setTestState(ctx context.Context, ts testState) context.Context {
+	return context.WithValue(ctx, testStateKey{}, ts)
+}
+
+// guessGitRoot looks for a directory containing a .git directory. It starts from the current
+// working dirctory and walks up the directory tree.
+func guessGitRoot() (string, error) {
```
zregvart, line 235: I don't think this is needed, tests are guaranteed to run with CWD of `$GITROOT/acceptance`, so we can assume `GITROOT=..`

Makefile
```diff
@@ -261,3 +261,12 @@ update-bundles: ## Push policy bundles to quay.io and generate infra-deployments
 	@hack/update-bundles.sh
 
 #--------------------------------------------------------------------
+
+##@ Acceptance Tests
+
+.PHONY: acceptance
+acceptance: ## Run acceptance tests
+	@go build -o acceptance/bin/ec github.com/enterprise-contract/ec-cli
```
zregvart, line 269: We could be using `go run` from the acceptance test instead? One executing the test from an IDE could end up running unintended version of ec (and I just saw we're doing this from in ec-cli, though we need to, to instrument for code coverage)

acceptance/acceptance_test.go
```diff
@@ -0,0 +1,275 @@
...
+	"regexp"
+	"testing"
+
+	"github.com/cucumber/godog"
+)
+
+type (
+	inputFileName  struct{}
+	configFileName struct{}
+	testStateKey   struct{}
```
zregvart, line 23: Use one context state? i.e. place input and config file paths in the `testState`?

zregvart (approved):
Neat, merge and refactor with the ideas from the comments...

acceptance/acceptance_test.go
lcarva: Oh the parenthesis were meant to help us group the samples together. I had another in there at some point. I'll remove it for now.

acceptance/acceptance_test.go
lcarva: Nice! I didn't know that.

acceptance/features/release.feature
lcarva: https://github.com/enterprise-contract/ec-policies/pull/866

acceptance/features/release.feature
lcarva: https://github.com/enterprise-contract/ec-policies/pull/866

acceptance/acceptance_test.go
lcarva: https://github.com/enterprise-contract/ec-policies/pull/866

acceptance/acceptance_test.go
lcarva: https://github.com/enterprise-contract/ec-policies/pull/866

acceptance/acceptance_test.go
lcarva: https://github.com/enterprise-contract/ec-policies/pull/866

Makefile
lcarva: This is a bit tricky. We can get the ec command like this:

```go
import "github.com/enterprise-contract/ec-cli/cmd"
...
func validateInputWithPolicyConfig(ctx context.Context) (context.Context, error) {
	...
	ec := cmd.RootCmd

	ec.SetContext(context.Background())
	ec.SetArgs([]string{
		"validate",
		"input",
		"--file",
		ts.inputFileName,
		"--policy",
		ts.configFileName,
		"--strict=false",
	})

	var stdout bytes.Buffer
	var stderr bytes.Buffer
	ec.SetOut(&stdout)
	ec.SetErr(&stderr)
	...
```

But this means that all tests are sharing a single instance of the command. This will be a problem as soon as we have more than one scenario :disappointed: 

We could also use `NewRootCmd()` from `/cmd/root` but that creates a command with no sub-commands. We have to call call `.AddCommand(...)` to add the command structure. However! That doesn't work because the function that creates the validate input command is not exported :cry: I'll play around with it. Maybe it makes sense to export this function.

lcarva: I was trying to use the ec cli as a go dependency. It is not, at least currently, meant to be used this way. I took a different approach in: https://github.com/enterprise-contract/ec-policies/pull/867

Issue #858, opened 2024-01-10
See about `bundles.is_acceptable` checking `_is_unacceptable`


zregvart:
This is to address the comment in https://github.com/enterprise-contract/ec-policies/pull/849#discussion_r1444956109. More context there.

PR #865, opened 2024-01-16
Minor simplification of bundles.is_acceptable

lcarva:
This commit changes the implementation of bundles.rego to avoid double negatives, e.g. `not _is_unacceptable`. Instead, the logic is implemented in the positive, i.e. `_is_acceptable. This makes the other functions that rely on this a little bit easier to read.

Resolves: #858


lcarva:
The improvement here is pretty minor, and somewhat subject. Happy to drop it if this is not an obvious improvement to others.

zregvart (approved):

PR #867, opened 2024-01-17
Do not require pre-built ec binary

lcarva:


zregvart (approved):
:+1: on the other PR I missed that acceptance was it's own go module. I think this is a good approach.

PR #874, opened 2024-01-24
Also update ec-task-policy image ref

lcarva:
The infra-deployments repo now contains ECP resources that use the task policy bundle. This commit teaches the update script to handle updating those references as well.

Ref: EC-358

### DIFF
--- a/hack/update-infra-deployments.sh
+++ b/hack/update-infra-deployments.sh
@@ -28,6 +28,7 @@ TARGET_DIR="${1}"
 cd "${TARGET_DIR}" || exit 1
 
 RELEASE_POLICY_REF='quay.io/enterprise-contract/ec-release-policy:latest'
+TASK_POLICY_REF='quay.io/enterprise-contract/ec-task-policy:latest'
 
 function oci_source() {
   img="${1}"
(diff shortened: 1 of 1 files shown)


hack/update-infra-deployments.sh
```diff
@@ -53,24 +54,31 @@ function oci_source() {
...
       yq e -e \
           '(select(has("kind")) | select(.kind == "EnterpriseContractPolicy"))' \
-          $yaml_file || continue
-      # Finally, update the source references.
-      SOURCE_KEY="${1}" SOURCE_URL="${2}" yq e -i \
-          '.spec.sources[0][env(SOURCE_KEY)][0] |= env(SOURCE_URL) | .' \
-          $yaml_file
+          $yaml_file 2> /dev/null || continue
+      # Finally, update the source references. A previous iteration used yq to perform a more
+      # precise update. However, making a conditional update is non-trivial. sed is a simpler here.
```
robnester-rh, line 68: nit: "sed is a simpler option here." or "sed is simpler."

robnester-rh, line 68: nit: "sed is a simpler option here." or "sed is simpler."

robnester-rh (approved):
Looks good, only nit was wording on comments which doesn't affect functionality.

hack/update-infra-deployments.sh
lcarva: Thanks! Fixed it.

PR #875, opened 2024-01-25
Add a new kind of acceptance tests for tekton task handling

simonbaird:
Perceived problems with our current test coverage:
* Lots of mocked data in the low level tests means it's hard to be confident the mocked data represents reality, and it's hard to maintain generally
* General untidiness of the handling of the multiple formats of tekton task data (SLSA v0.2, SLSA v1.0, the different task resolvers for each, etc) makes it hard to grok and maintain.

How this might help:
* Introduces a test coverage that uses a real attestation rather than mocked data so we feel more confident the behavior being asserted is based on reality.
* Because it's higher level the test writer or review doesn't have to know low level details about the different formats to write more and better test coverage, therefore test maintenance becomes easier.
* The expected data in the feature file serves as a visible way to help readers understand what the low level code is supposed to be doing.


simonbaird:
I tidied it up. I think this could be okay to merge, wdyt?

simonbaird:
Note that the sample attestations include git resolved tasks, which will be useful for adding new code that handles those.

acceptance/acceptance_test.go
```diff
@@ -178,6 +187,86 @@ func thereShouldBeNoWarningsInTheResult(ctx context.Context) error {
...
+	if err != nil {
+		return ctx, fmt.Errorf("opaEval get test state: %w", err)
+	}
+
+	cmd := exec.Command(
+		"go",
+		"run",
+		"github.com/enterprise-contract/ec-cli",
+		"opa",
+		"eval",
```
lcarva, line 201: An argument for not using `opa eval` is that it doesn't represent what actually happens when the EC CLI runs, which is the main (only?) use for the policies in this repository.

But looking at the newly added feature file, I can definitely see how this can streamline things. I do wonder if that is better done as a unit test that uses real data :thinking:

acceptance/acceptance_test.go
```diff
@@ -178,6 +187,86 @@ func thereShouldBeNoWarningsInTheResult(ctx context.Context) error {
...
+	// Unmarshal the expected result from json so we don't need to worry
+	// about formatting diffs
+	var expectedResultData any
+	err = json.Unmarshal([]byte(expectedResult), &expectedResultData)
+	if err != nil {
+		return fmt.Errorf("parsing expected result json: %w", err)
+	}
+
+	// Marshal it back to consistently indented json
+	// Todo maybe: use github.com/yudai/gojsondiff to produce diffs
```
lcarva, line 239: Checkout `github.com/google/go-cmp/cmp`. It's what Chains uses. It gives you a lot of options, including a consistent way to format.

acceptance/acceptance_test.go
```diff
@@ -178,6 +187,86 @@ func thereShouldBeNoWarningsInTheResult(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("remarshalling expected result data: %w", err)
+	}
+
+	// Unmarshal the actual opa eval result
+	// We get something like this on stdout:
+	//   {"result":[{"expressions":[{"value": ...}]}]}
+	// I think we'll generally have just have one expression and one result,
+	// so let's pull that value up so the feature files are a little tidier
+	var parsedOutput map[string][]map[string][]map[string]any
```
lcarva, line 250: Maybe just use `var parsedOutput any` here for consistency with the expected result?

acceptance/acceptance_test.go
```diff
@@ -178,6 +187,86 @@ func thereShouldBeNoWarningsInTheResult(ctx context.Context) error {
...
+	if err := cmd.Run(); err != nil {
+		return ctx, fmt.Errorf("running ec opa eval: %w\n%s", err, stderr.String())
+	}
+
+	ts.stdout = string(stdout.Bytes())
+
+	return setTestState(ctx, ts), nil
+}
+
+func theResultShouldBe(ctx context.Context, expectedResult string) error {
```
lcarva, line 224: There are some assumptions here that the result is in JSON format, and it is from `opa eval`. I think we can remove the assumption about `opa eval` without any consequence, see my other comment further down. Maybe call this `theJSONResultShouldBe` or `theOPAEvalResultShouldBe` depending on the remaining assumptions.

acceptance/features/tekton_tasks.feature
```diff
@@ -0,0 +1,60 @@
+Feature: Lib Tekton
```
lcarva, line 1: Should this file be called `tekton_lib` instead?

lcarva (approved):
I think it's worth getting this in. But do keep in mind about approaching this via unit tests instead.

simonbaird:
> approaching this via unit tests instead

Yeah, maybe. I have working examples of this, see [this](https://github.com/enterprise-contract/ec-policies/pull/818/commits/1932bdb5d6a69fae97a8ca92f3001b8480f65cad) which pulls the records and turns them into rego. Maybe I'll dust it off and see how it looks with this kind of test.

acceptance/acceptance_test.go
```diff
@@ -78,6 +83,10 @@ func writeSampleGCPolicyInput(ctx context.Context, sampleName string) (context.C
 		content = sampleGCPolicyInput
 	case "clamav-task":
 		content = sampleClamAVTask
+	case "v02-sample-attestation":
```
zregvart, line 86: With `embed.FS`, this becomes something like:

```
	sampleData, err := sample.ReadFile(filepath.Join("sample", sampleName + ".json")
	if err != nil {
		return ctx, fmt.Errorf("reading sample %s: %w", sampleName, err)
	}

	return ctx, os.WriteFile(ts.inputFileName, sampleData, 0400)
```

simonbaird: I like this for a future PR.

simonbaird:
Still draft, but I rebased and applied some of the review suggestions.

simonbaird:
#911 is merged. Abandoning this one.

PR #881, opened 2024-01-31
Examples for representing git ref for Tasks

lcarva:


example/data/acceptable_tekton_tasks_git.yml
```diff
@@ -0,0 +1,64 @@
...
+    - commit: 81185d669f543f5cd9114377f7ee0480e15a7ac1
+      effective_on: 2024-02-10T00:00:00Z
+  https://github.com/redhat-appstudio/build-definitions.git//task/clair-scan/0.1/clair-scan.yaml:
+    - commit: 28b86c8b94c42995d98f843b0fc854c1cf2ae500
+      effective_on: 2024-02-15T00:00:00Z
+    - commit: 4efdfc8c8da9f78602b5a21bcd2b3cf5667503fc
+      effective_on: 2024-01-31T00:00:00Z
+
+# Align with how the git resolver is used.
+acceptable-task-git-refs-option3:
```
simonbaird, line 35: Not really liking this one now that we can see it. Also the revision.values being a list means we have deviated from the tekton schema anyhow.

example/data/acceptable_tekton_tasks_git.yml
```diff
@@ -0,0 +1,64 @@
...
+  https://github.com/redhat-appstudio/build-definitions.git//task/buildah/0.1/buildah.yaml:
+    - commit: 3672a457e3e89c0591369f609eba727b8e84108f
+      effective_on: 2024-02-16T00:00:00Z
+    - commit: 81185d669f543f5cd9114377f7ee0480e15a7ac1
+      effective_on: 2024-02-10T00:00:00Z
+  https://github.com/redhat-appstudio/build-definitions.git//task/clair-scan/0.1/clair-scan.yaml:
+    - commit: 28b86c8b94c42995d98f843b0fc854c1cf2ae500
+      effective_on: 2024-02-15T00:00:00Z
+    - commit: 4efdfc8c8da9f78602b5a21bcd2b3cf5667503fc
+      effective_on: 2024-01-31T00:00:00Z
```
simonbaird, line 32: If we're moving to this conceptual style for the bundles schema, maybe it's wise to be consistent with that.

lcarva: This is the reason I like this format the most.

example/data/acceptable_tekton_tasks_git.yml
```diff
@@ -0,0 +1,64 @@
...
+          effective_on: 2024-01-31T00:00:00Z
+
+# The Earth is flat
+acceptable-task-git-refs-option4:
+  - ref: https://github.com/redhat-appstudio/build-definitions.git//task/buildah/0.1/buildah.yaml?ref=<sha>
+    effective_on: 2024-02-15T00:00:00Z
+  - ref: https://github.com/redhat-appstudio/build-definitions.git//task/buildah/0.1/git-clone.yaml?ref=<othersha>
+    effective_on: 2024-03-15T00:00:00Z
+  - ref: https://github.com/redhat-appstudio/build-definitions.git//task/buildah/0.1/buildah.yaml?ref=<differentsha>
+    effective_on: 2024-03-15T00:00:00Z
```
simonbaird, line 64: Hmm, flat one does have some appeal IMO. Easy to read and understand. We can sort it at build time.

simonbaird: Could also convert it to a nested structure at run time if we decide the scanning efficiency really does matter.

lcarva: I think a flat structure like this is going to make it really hard to make any sense of the data. Imagine this list has 100 records.

If any processing of the data requires converting the data, maybe that's an indicator the data is not in the right format?

lcarva: On the other hand, a list like this could potentially make it easier to consolidate the representation of git resolver and bundle refs.

example/data/acceptable_tekton_tasks_git.yml
simonbaird: Yeah I think it's my favorite also.

example/data/acceptable_tekton_tasks_git.yml
```diff
@@ -0,0 +1,64 @@
...
+    task/clair-scan/0.1/clair-scan.yaml:
+      - commit: 28b86c8b94c42995d98f843b0fc854c1cf2ae500
+        effective_on: 2024-02-15T00:00:00Z
+      - commit: 4efdfc8c8da9f78602b5a21bcd2b3cf5667503fc
+        effective_on: 2024-01-31T00:00:00Z
+
+# Align with the future of acceptable bundles format:
+# https://github.com/enterprise-contract/ec-cli/issues/1291
+acceptable-task-git-refs-option2:
+  https://github.com/redhat-appstudio/build-definitions.git//task/buildah/0.1/buildah.yaml:
```
zregvart, line 23: Perhaps:
```suggestion
  git+https://github.com/redhat-appstudio/build-definitions.git//task/buildah/0.1/buildah.yaml:
```

Seems to be something that Chains prefers to place in the attestations

example/data/acceptable_tekton_tasks_git.yml
```diff
@@ -0,0 +1,64 @@
...
+      effective_on: 2024-02-16T00:00:00Z
+    - commit: 81185d669f543f5cd9114377f7ee0480e15a7ac1
+      effective_on: 2024-02-10T00:00:00Z
+  https://github.com/redhat-appstudio/build-definitions.git//task/clair-scan/0.1/clair-scan.yaml:
+    - commit: 28b86c8b94c42995d98f843b0fc854c1cf2ae500
+      effective_on: 2024-02-15T00:00:00Z
+    - commit: 4efdfc8c8da9f78602b5a21bcd2b3cf5667503fc
+      effective_on: 2024-01-31T00:00:00Z
+
+# Align with how the git resolver is used.
```
zregvart, line 34: If we set the base model on the Tekton bundle resolvers, i.e. we don't support the legacy bundle. Is there a way to encode both git and oci/bundle Task references in the same structure?

lcarva: I think so. There's value in, somehow, combining those two pieces of information. Added option 5 to the bottom (which is a variation of option 2).

example/data/acceptable_tekton_tasks_git.yml
```diff
@@ -0,0 +1,86 @@
...
+acceptable-task-git-refs-option4:
+  - ref: https://github.com/redhat-appstudio/build-definitions.git//task/buildah/0.1/buildah.yaml?ref=<sha>
+    effective_on: 2024-02-15T00:00:00Z
+  - ref: https://github.com/redhat-appstudio/build-definitions.git//task/buildah/0.1/git-clone.yaml?ref=<othersha>
+    effective_on: 2024-03-15T00:00:00Z
+  - ref: https://github.com/redhat-appstudio/build-definitions.git//task/buildah/0.1/buildah.yaml?ref=<differentsha>
+    effective_on: 2024-03-15T00:00:00Z
+
+# Combine bundles and git resolver information into a unified data structure
+acceptable-tasks-option5:
```
lcarva, line 67: @simonbaird, @zregvart, curious on your thoughts on this one

zregvart: This works for me. The only bit I'm not 100% on is the `id`, perhaps `ref` or `rev` instead?

lcarva:
Thank you all for the feedback. Closing this out as I believe it has served its purpose.

PR #885, opened 2024-02-02
Add trusted_task package

lcarva:
This new package allows verifying that a certain Task is trusted. It supports both the bundles and the git resolvers.

Eventually, this will supersede the attestation_task_bundle package.

Ref: EC-186

### DIFF
--- a/policy/lib/tekton/trusted.rego
+++ b/policy/lib/tekton/trusted.rego
@@ -1,0 +1,61 @@
+package lib.tkn
+
+import rego.v1
+
+import data.lib.refs
+import data.lib.time as time_lib
+
+# Returns a subset of tasks that use unpinned Task references.
+unpinned_task_references(tasks) := {task |
+	some task in tasks
+	not refs.task_ref(task).pinned
+}
+
+# Returns if the list of trusted Tasks are missing
+default missing_trusted_tasks_data := false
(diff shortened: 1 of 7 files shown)


lcarva:
Opened as a draft because I'm still working out the details, but wanted to gather some initial feedback.

policy/lib/tekton/trusted.rego
```diff
@@ -0,0 +1,85 @@
...
+	records := _trusted_tasks[ref.key]
+
+	newest_record := time_lib.newest(records)
+	newest_record.ref != ref.pinned_ref
+}
+
+# _trusted_tasks provides a safe way to access the list of trusted tasks. It prevents a policy rule
+# from incorrectly not evaluating due to missing data. It also removes stale records.
+_trusted_tasks[repo] := pruned_records if {
+	some repo, records in data.trusted_tasks
```
zregvart, line 61: Nitpick, to keep with the naming

```suggestion
_trusted_tasks[key] := pruned_records if {
	some key, records in data.trusted_tasks
```

zregvart (approved):
I think this is rather nice

lcarva:
Includes commit from https://github.com/enterprise-contract/ec-policies/pull/891.

PR #894, opened 2024-02-07
wip: Set effective_on dynamically

lcarva:
This is a proof of concept for an idea discussed in the community meeting. It leverages existing functionality in the EC CLI to demote certain violations to warnings. So far, this has been used by setting a custom annotation, `effective_on`, on the policy rule metadata. The function `lib.result_helper*` takes this value and adds it to the violation object.

In this commit, the policy rule dynamically determines what the effective_on value should be based on a given data item, instead of grabbing it from the policy rule's metadata.

This effectively means we can have time-based policy data items. This is a simplistic example. I plan on showcasing more advanced ones shortly.

### DIFF
--- a/policy/release/attestation_task_bundle.rego
+++ b/policy/release/attestation_task_bundle.rego
@@ -16,6 +16,8 @@ import rego.v1
 
 import data.lib
 import data.lib.bundles
+import data.lib.image
+import data.lib.refs
 import data.lib.tkn
 
 # METADATA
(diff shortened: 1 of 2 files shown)


lcarva:
You can test it like this:

(policy.yaml)
```yaml
---
publicKey: |-
  -----BEGIN PUBLIC KEY-----
  MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEZP/0htjhVt2y0ohjgtIIgICOtQtA
  naYJRuLprwIv6FDhZ5yFjYUEtsmoNcW7rx2KM6FOXGsCX3BNc7qhHELT+g==
  -----END PUBLIC KEY-----
sources:
  - policy:
      - github.com/enterprise-contract/ec-policies//policy/lib
      - github.com/lcarva/ec-policies//policy/release?ref=poc-dynamic-effective-on
    ruleData:
      disallowed_inherited_labels:
        - name: description
        - name: io.k8s.description
        - name: io.k8s.display-name
        - name: io.openshift.tags
        - name: summary
          effective_on: 2024-01-01T00:00:00Z
        - name: name
          effective_on: 2024-05-01T00:00:00Z
        - name: com.redhat.component
          effective_on: 2024-05-01T00:00:00Z
    config:
      include:
        - labels

```

```
ec validate image --policy policy.yaml \
   --output yaml --show-successes --ignore-rekor \
  --image quay.io/redhat-user-workloads/rhtas-tenant/trillian-1-0-gamma/trillian-logsigner-1-5@sha256:015f430f8f966d5cf83cdaad5ba5ca87dd75a83e64abd229c00ddd370f3c393b 
```

lcarva:
Added an example for using this pattern for determining acceptable bundles. We end up with less rules, and no date manipulation at all in rego.

It does require that the list of acceptable bundles contains the `expires_on` attribute. I was able to do that with https://github.com/enterprise-contract/ec-cli/commit/4a09043197194ebc3a6297d1548a3ca203e5a35e.

policy/release/attestation_task_bundle.rego
```diff
@@ -78,48 +80,55 @@ warn contains result if {
...
+	records := [record |
+		some record in data["task-bundles"][ref.repo]
+		record.tag == ref.tag
+		record.digest == ref.digest
+	]
+	count(records) > 0
+} else := [{"expires_on": _missing_record}]
+
+# TODO: This has a slightly different implementation than the one from the labels package. Good
+# enough for a POC but we should try to avoid such things.
```
zregvart, line 120: Not seeing any issues with the code, is your concern in having two implementations? I'd be okay with that, both require slightly different things and a bit of duplication here should be okay.

zregvart (approved):
When do we get to merge this :)?

policy/release/attestation_task_bundle.rego
lcarva: Yeah. It's probably ok.

lcarva:
> When do we get to merge this :)?

I think we should be good to go for the labels stuff. It should be backwards compatible. I'll make a proper PR.

For the acceptable bundles, we need `expires_on` in the data. So this is blocked on a change in the ec-cli, https://github.com/enterprise-contract/ec-cli/issues/1332.

lcarva:
Create https://github.com/enterprise-contract/ec-policies/pull/896 for adding time-based support to the label checks. I'm gonna go ahead and close out this POC as I believe it has served its purpose. Thanks all!

PR #897, opened 2024-02-10
Fix update-bundles.sh script

lcarva:
The conditional branch to skip re-pushing a bundle required the `ENSURE_UNIQUE` environment variable to be set to `"1"`. A quick find across the EC repos didn't find any occurrences of this env var being set, at all.

The `ENSURE_UNIQUE` seems like the remnants of a previous iteration of this script that is not longer used.

This commit removes support for `ENSURE_UNIQUE`. It also moves registry query to its own line instead of inlininig it in the if-statement for readability purposes.


hack/update-bundles.sh
```diff
@@ -80,7 +72,11 @@ for b in $BUNDLES; do
     skopeo_cp_args='--dest-tls-verify=false --src-tls-verify=false'
   fi
 
-  if [ "$(skopeo list-tags ${skopeo_args} "docker://${push_repo}" | jq 'any(.Tags[]; . == "'"${tag}"'")')" == "true" ] && [ "$ENSURE_UNIQUE" == "1" ]; then
+  tag_found="$(
+    skopeo list-tags --tls-verify=false docker://localhost:5000/ec2/ec-task-policy |
```
zregvart, line 76: Typo:
```suggestion
    skopeo list-tags "docker://${push_repo}" |
```

zregvart (changes requested):
Found a typo

zregvart (approved):

PR #902, opened 2024-02-14
Warn/fail for Task expiry

zregvart:
Task definition can now set the
`build.appstudio.redhat.com/expires-on` annotation with the date in RFC3339 format (shorthand supported) that will emit a violation if that date has passed, or a warning until.

Ref: https://issues.redhat.com/browse/EC-353


policy/release/tasks.rego
```diff
@@ -227,6 +227,78 @@ deny contains result if {
...
+
+	annotations := tkn.task_annotations(task)
+
+	deprecated_on := annotations[_expires_on_annotation]
+	deprecated_on_rfc3339 := sprintf("%s%s", [
+		deprecated_on,
+		substring("0000-00-00T00:00:00Z", count(deprecated_on), 20 - count(deprecated_on)),
+	])
+
+	time.parse_rfc3339_ns(deprecated_on_rfc3339) <= lib.time.effective_current_time_ns
```
lcarva, line 293: Let's avoid doing any sort of date processing in rego.

I think here we can combine both the warn and deny policy rules into a single deny rule. Then, simply set the `effective_on` in the result based on the value from the deprecated label.

policy/release/tasks.rego
```diff
@@ -339,3 +411,5 @@ _format_missing(o, opt) := desc if {
 	opt
 	msg := sprintf("Task %q", [o])
 } else := sprintf("Required task %q", [o])
+
+_expires_on_annotation := "build.appstudio.redhat.com/expires-on"
```
lcarva, line 415: We should probably include an explanation in the description of the policy rule(s) about how EC determines a Task is expired/deprecated.

policy/release/tasks.rego
simonbaird: +1

policy/release/tasks.rego
simonbaird: Something like this maybe as a comment:
```
// If the annotation is present at all, then we consider the task deprecated
// If the value is a date in the past, then we consider the task disallowed
```

But I wonder how a task author would find out about it. Do we have any docs aimed at task authors?

zregvart: I don't see any use of setting the annotation without specifying the date.

Since the value must always be set for an annotation, I don't want to allow setting different kinds of values, e.g. support `build.appstudio.redhat.com/expires-on: true` (boolean) or `build.appstudio.redhat.com/expires-on: ""` (string), only `build.appstudio.redhat.com/expires-on: 2023-04-05` (date).

zregvart:
Please have another look. Suggestions on the annotation values (title, description, code) are especially welcome...

policy/release/tasks.rego
```diff
@@ -227,6 +227,44 @@ deny contains result if {
...
+#   - tasks.pipeline_has_tasks
+#
+deny contains result if {
+	some att in lib.pipelinerun_attestations
+	some task in tkn.tasks(att)
+
+	annotations := tkn.task_annotations(task)
+
+	deprecated_on := annotations[_expires_on_annotation]
+	deprecated_on_rfc3339 := sprintf("%s%s", [
```
lcarva, line 253: If we care about the format of the value (which I think is a good idea), can we just use `time.parse_rfc3339_ns` instead?

The RFC3339 allows different variations (not all have 20 characters). Maybe it's a good idea to just stick with it?

(This is probably contradictory to what I said earlier that we shouldn't do any time processing in rego. I do think verifying the value is in the format we expect is something rego should do, e.g. we do it for most of the rule data values.)

policy/release/tasks.rego
```diff
@@ -227,6 +227,44 @@ deny contains result if {
...
+		deprecated_on,
+		substring("0000-00-00T00:00:00Z", count(deprecated_on), 20 - count(deprecated_on)),
+	])
+
+	result := object.union(
```
lcarva, line 258: Nice and clean

zregvart:
Reworded the task towards `unsupported`, added a Task rule to check for the date format of the annotation.

policy/release/tasks.rego
lcarva: Then that's gonna fail to be parsed the by the ec-cli: https://github.com/enterprise-contract/ec-cli/blob/e86012ff8925632abe7ef41d6a11e084f5ad09ac/internal/evaluator/conftest_evaluator.go#L136

Date without a timezone is just as ambiguous as time without a timezone. I think we should just stick to well know timedate format...

zregvart: `effective_on` here is in RFC3339 format, date without time and timezone is in UTC, because that's what we defined it as

lcarva: This is all still in development. Very easy to adjust our assumption to the realities we encounter :shrug:

lcarva (approved):

lcarva:
Let's be sure to add the new policy package from the task namespace in build-definitions: https://github.com/redhat-appstudio/build-definitions/blob/main/policies/all-tasks.yaml

lcarva (approved):

PR #911, opened 2024-02-22
Introduce tests using realistic attestation data

simonbaird:
Alternative to #875 . Which do we like better..?

See commit messages for more details.


zregvart:
This is perhaps a bit neater than #875. An interesting followup could be that we automate PR creation against hacks/recordings and here for actual production workload. Say for the golden container. We could add different kinds of Tekton Pipelies there, but also include attestations from different build systems (GitHub Actions, Jenkins...).
I like that this keeps it within Rego and doesn't require running additional infrastructure while it is still representable of real attestations. Makes for (a bit) quicker turnarounds.

simonbaird:
There was a little more work needed to get the lint checker to pass.

zregvart (approved):

PR #915, opened 2024-02-27
Tolerate git URL with a slash at the end

zregvart:
This adds an additional case in the `_ref` function to handle the SCM URLs, e.g. `https://github.com/org/repository/`. This is needed as the attestation will contain something like
`git+https://github.com/org/repository.git`


mbestavros (approved):
LGTM

simonbaird (approved):

PR #917, opened 2024-03-01
Add source_image package

lcarva:
This change adds a new policy package, source_image. It leverages the newly added `ec.oci.image_manifest` rego function to verify the source image reference found in the SLSA Provenance is valid.

Because it relies on a newly added custom rego function, the package is added to the `beta` namespace. At some point in the future, it will be moved to the `release` namespace. The benefit of adding this package now is that others can more easily verify its behavior.

Ref: [EC-235](https://issues.redhat.com/browse/EC-235)


policy/beta/source_image.rego
```diff
@@ -0,0 +1,56 @@
...
+	error := sprintf("Source image has no layers %q", [img])
+}
+
+# _source_images is a set of image references. Each corresponding to the
+# SOURCE_IMAGE_URL@SOURCE_IMAGE_DIGEST parameter of a source-build Task.
+_source_images contains img if {
+	some att in lib.pipelinerun_attestations
+	some task in tkn.tasks(att)
+
+	"source-build" in tkn.task_names(task)
```
zregvart, line 51: Should this be a helper function in `policy/lib/tekton/task.rego`, like the `build_tasks` or `git_clone_tasks`?

zregvart (approved):

simonbaird (approved):

zregvart (approved):

PR #944, opened 2024-03-28
Fetch SBOM from OCI blob

zregvart:
This is using `ec.oci.blob` function to fetch the SBOM from an OCI blob, expecting a JSON string to be present, which will be unmarshaled and stored in the `data.lib.sbom.cyclonedx_sboms` array.

This all will happen only if the `beta` source directory is included in the policy sources.

Reference: EC-499

### DIFF
--- a/policy/lib/sbom_test.rego
+++ b/policy/lib/sbom_test.rego
@@ -15,12 +15,35 @@ test_cyclonedx_sboms if {
 			"predicateType": "https://example.org/boom",
 			"predicate": "not an sbom",
 		}},
+		{"statement": {"predicate": {
+			"buildType": lib.tekton_pipeline_run,
+			"buildConfig": {"tasks": [{"results": [
+				{
+					"name": "IMAGE_DIGEST",
+					"type": "string",
+					"value": "sha256:f0cacc1a",
+				},
+				{
+					"name": "IMAGE_URL",
+					"type": "string",
+					"value": "registry.io/repository/image@sha256:baadf00d",
+				},
+				{
+					"name": "SBOM_BLOB_URL",
+					"type": "string",
+					"value": "registry.io/repository/image@sha256:f0cacc1a",
+				},
+			]}]},
+		}}},
 	]
 	image := {"files": {
 		"root/buildinfo/content_manifests/sbom-cyclonedx.json": "sbom from image",
 		"root/foo": "not an sbom",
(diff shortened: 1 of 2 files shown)


zregvart:
In draft as I don't know how to cover the [else line](https://github.com/enterprise-contract/ec-policies/pull/944/files#diff-c246490dce29a9415d6325293912f9de93a429af7c04f9eba981a7e3b64f28b5R26) and I need to figure out the testing which involves `ec.oci.blob` function.

policy/lib/sbom_test.rego
```diff
@@ -24,3 +24,11 @@ test_cyclonedx_sboms if {
 	lib.assert_equal(sbom.cyclonedx_sboms, expected) with input.attestations as attestations
 		with input.image as image
 }
+
+test_opt_in_oci_sbom if {
+	lib.assert_equal(sbom._cyclonedx_sboms_from_oci, [])
+	lib.assert_equal(sbom.cyclonedx_sboms, [])
+
+	lib.assert_equal(sbom._cyclonedx_sboms_from_oci, [{}]) with sbom.oci._cyclonedx_sboms as [{}]
```
lcarva, line 32: Add this to cover the `else` statement:

```rego
lib.assert_equal(sbom._cyclonedx_sboms_from_oci, []) with sbom.oci as {}
```

policy/beta/sbom/sbom_oci.rego
```diff
@@ -0,0 +1,15 @@
...
+import data.lib
+import data.lib.tkn
+import rego.v1
+
+_cyclonedx_sboms := [sbom |
+	some attestation in lib.pipelinerun_attestations
+	some task in tkn.build_tasks(attestation)
+
+	blob_ref := tkn.task_result(task, "SBOM_BLOB_URL")
+	blob := ec.oci.blob(blob_ref)
```
lcarva, line 12: I think this function has been around for long enough that we can probably bring it out of beta?

zregvart: I wasn't too sure when we added it, but it was last year in October, so yeah, I'll move it to the release...

policy/lib/sbom_test.rego
zregvart: TIL

lcarva (approved):

PR #945, opened 2024-03-29
Use Trusted task terminology

zregvart:
Reference: EC-390

### DIFF
--- a/policy/release/attestation_task_bundle.rego
+++ b/policy/release/attestation_task_bundle.rego
@@ -106,7 +106,7 @@ warn contains result if {
 #   For each Task in the SLSA Provenance attestation, check if the Tekton Bundle used is
 #   a trusted task.
 # custom:
-#   short_name: task_ref_bundles_acceptable
+#   short_name: task_ref_bundles_trusted
 #   failure_msg: Pipeline task '%s' uses an untrusted task bundle '%s'
 #   solution: >-
 #     For each Task in the SLSA Provenance attestation, check if the Tekton Bundle used is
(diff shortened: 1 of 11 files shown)


zregvart:
Make note of the second commit:

[Mass renaming of *acceptable to *trusted](https://github.com/enterprise-contract/ec-policies/pull/945/commits/0ae90b8ae498b6ccfef10a2f38ecc3324546c889)

Attention: this also changes rule short codes, which I think for these
rules should be okay.

policy/release/attestation_task_bundle.rego
```diff
@@ -106,7 +106,7 @@ warn contains result if {
 #   For each Task in the SLSA Provenance attestation, check if the Tekton Bundle used is
 #   a trusted task.
 # custom:
-#   short_name: task_ref_bundles_acceptable
+#   short_name: task_ref_bundles_trusted
```
lcarva, line 109: I know of at least one user that excludes `attestation_task_bundle.task_ref_bundles_acceptable`. We could simply add the new name now to the policies in [here](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/tree/main/config/OP/EnterpriseContractPolicy).

lcarva (commented):
I think we should make this change, but let's see if we can first update the internal policies we are aware of.

lcarva (approved):

zregvart:
Tracking rollout, there was an issue in the pipeline, just making sure the change has been deployed

zregvart:
Validated in audit logs the new policy was applied.

PR #952, opened 2024-04-07
Add policy check to verify if the digest was scanned

yashvardhannanavati:
For tests emitting IMAGES_PROCESSED result, verify the input image digest has been processed.

Resolves: CVP-4067

### DIFF
--- a/policy/release/test.rego
+++ b/policy/release/test.rego
@@ -10,6 +10,7 @@ package policy.release.test
 import rego.v1
 
 import data.lib
+import data.lib.image
 
 # METADATA
 # title: Test data found in task results
@@ -220,6 +221,34 @@ warn contains result if {
 deny contains result if {
 	some error in _rule_data_errors
 	result := lib.result_helper(rego.metadata.chain(), [error])
+}
+
+# METADATA
+# title: Image digest is present in IMAGES_PROCESSED result
+# description: >-
+#   Ensure that task producing the IMAGES_PROCESSED result contains the
+#   digests of the built image.
+# custom:
+#   short_name: test_all_images
+#   failure_msg: Test '%s' did not process image with digest '%s'.
+#   solution: >-
+#     Found an image not processed by a task. Make sure that the task
+#     processes and includes the image digest of all images in the
+#     `IMAGES_PROCESSED` result.
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	img := image.parse(input.image.ref)
+	img_digest := img.digest
+
+	some task in lib.images_processed_results_from_tests
+	not img_digest in object.get(task.value, ["image", "digests"], [])
+	result := lib.result_helper_with_term(
+		rego.metadata.chain(),
+		[task.name, img_digest],
+		task.name,
+	)
 }
 
 # Collect all tests that have resulted with one of the given
--- a/policy/release/test_test.rego
+++ b/policy/release/test_test.rego
@@ -435,6 +435,37 @@ test_wrong_attestation_type if {
 	lib.assert_empty(test.deny) with input.attestations as [tr, tr_slsav1]
 }
 
+test_all_image_processed if {
+	# regal ignore:line-length
+	digests_processed := {"image": {"digests": ["sha256:4e388ab32b10dc8dbc7e28144f552830adc74787c1e2c0824032078a79f227fb"]}}
+	pipeline_run := lib_test.att_mock_helper_ref(lib.task_test_image_result_name, digests_processed, "success_23", _bundle)
+	attestations := [
+		pipeline_run,
+		lib_test.att_mock_helper_ref(lib.task_test_result_name, {"result": "SUCCESS"}, "errored_1", _bundle),
+	]
+
+	lib.assert_empty(test.deny) with input.attestations as attestations
+		with input.image.ref as _bundle
+}
+
+test_all_images_not_processed if {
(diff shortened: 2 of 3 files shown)


yashvardhannanavati:
@lcarva @simonbaird @zregvart Could you please help review this?

policy/release/test.rego
```diff
@@ -222,6 +223,40 @@ deny contains result if {
...
 }
 
+# METADATA
+# title: Verify image disgest is present in IMAGES_PROCESSED result
+# description: >-
+#   Validates that each task with the IMAGES_PROCESSED result contains the
+#   digest of the image beind processed.
+# custom:
+#   short_name: image_digest_processed
+#   failure_msg: Test '%s' is missing '%s' from IMAGES_PROCESSED result.
```
zregvart, line 233: Perhaps
```suggestion
#   failure_msg: Task %q did not process image with digest %q.
```

policy/release/test.rego
```diff
@@ -222,6 +223,40 @@ deny contains result if {
...
+#   solution: >-
+#     There is a task with result `IMAGES_PROCESSED` which did not scan
+#     a digest. Check that all tasks with `IMAGES_PROCESSED` result has
+#     processed every digest of the input image(s).
```
zregvart, line 237: Perhaps
```suggestion
#     Found an image not processed by a task. Make sure that the task
#     processes and includes the image digest of all images in the 
#     `IMAGES_PROCESSED` result.
```

policy/release/test.rego
```diff
@@ -222,6 +223,40 @@ deny contains result if {
...
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	input_image := image.parse(input.image.ref)
+	input_image_digest := input_image.digest
+	unprocessed_digest_task := [unprocessed_task |
+		some x_result in lib.images_processed_results_from_tests
+		result_map := lib.unmarshal(x_result.value)
```
zregvart, line 246: `lib.results_named` should already unmarshal (if valid JSON), so `.value` should be an object here.

policy/release/test.rego
```diff
@@ -222,6 +223,40 @@ deny contains result if {
...
+		unprocessed_task := x_result.name
+	]
+
+	count(unprocessed_digest_task) > 0
+	some task in unprocessed_digest_task
+	result := lib.result_helper_with_term(
+		rego.metadata.chain(),
+		[task, input_image_digest],
+		task,
+	)
```
zregvart, line 257: I think this can be simpler, from the top of my head:

```suggestion
	img := image.parse(input.image.ref)
	digest := input_image.digest
	
	some task in lib.images_processed_results_from_tests
	not digest in object.get(task.value, ["image", "digests"], [])

	result := lib.result_helper_with_term(
		rego.metadata.chain(),
		[task.name, digest],
		task,
	)
```

zregvart (commented):
Some suggestions, I think on the wording others might have better suggestions

policy/release/test.rego
yashvardhannanavati: Makes sense. This indeed seems like a cleaner approach. I'll change it

policy/release/test.rego
yashvardhannanavati: I did try this before but the result string then contains `\"` around the task name and digest. So I decided to use %s with single quotes instead. Please let me know what you think

zregvart: I think that should be okay, we already use %q [elswhere](https://github.com/enterprise-contract/ec-policies/blob/8287b80e8398b275d435089b0cfb0058bbfd054c/policy/release/sbom_spdx.rego#L78)

policy/release/test_test.rego
```diff
@@ -435,6 +435,47 @@ test_wrong_attestation_type if {
 	lib.assert_empty(test.deny) with input.attestations as [tr, tr_slsav1]
 }
 
+test_all_image_processed if {
+	# regal ignore:line-length
+	digests_processed := {"image": {"digests": ["sha256:4e388ab32b10dc8dbc7e28144f552830adc74787c1e2c0824032078a79f227fb"]}}
+	images_processed_result := {
```
zregvart: This variable is unused ([checks caught that)](https://github.com/enterprise-contract/ec-policies/pull/952/checks#step:6:29), similarly for the test below

policy/release/test_test.rego
```diff
@@ -435,6 +435,47 @@ test_wrong_attestation_type if {
...
+		"name": lib.task_test_image_result_name,
+		"type": "string",
+		"value": json.marshal(digests_processed),
+	}
+	pipeline_run := lib_test.att_mock_helper_ref(lib.task_test_image_result_name, digests_processed, "success_23", _bundle)
```
zregvart, line 441: Perhaps a nicer task name would be more helpful here and below

```suggestion
	pipeline_run := lib_test.att_mock_helper_ref(lib.task_test_image_result_name, digests_processed, "task-image-check", _bundle)
```

zregvart (commented):
Make sure to run `make fmt` to format the code as well

policy/release/test.rego
yashvardhannanavati: so is `'%s'` : https://github.com/enterprise-contract/ec-policies/blob/main/policy/release/test.rego#L66

zregvart (approved):

PR #958, opened 2024-04-10
Add license header lint

zregvart:
Copied over from ec-cli repository.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,20 @@
 ---
+# Copyright The Enterprise Contract Contributors
+#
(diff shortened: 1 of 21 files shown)


.github/dependabot.yml
```diff
@@ -1,3 +1,19 @@
+# Copyright The Enterprise Contract Contributors
```
lcarva, line 1: This should go after the `---` line. Otherwise, this file is interpreted as two yaml documents which may not always work as intended.

zregvart: Perhaps, I'm sure we don't want to have a license header per-document though...

lcarva (approved):

PR #964, opened 2024-04-15
Revert "Add policy check to verify if the digest was scanned"

yashvardhannanavati:
This reverts commit 901d9e1eab0dbdd3d8842870dfd4ba1b18cea76a. 

Revert this temporarily until ec-cli PR #1486 has been merged.


lcarva (approved):
This is causing an issue because the `IMAGES_PROCESSED` results store the digests of the Image Manifests. However, the digest of the image is the digest of the Image Index. We need a change in the ec-cli, https://github.com/enterprise-contract/ec-cli/pull/1486, in order for this to work as expected.

robnester-rh (commented):
LGTM

PR #965, opened 2024-04-16
Generate the documentation via templates

zregvart:
Rather than generating the documentation via Antora extension at website build time, this generates the documentation using golang templates at build time.

Reference: EC-513


docs/.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
lcarva, line 2: Maybe remove these comments since that's no valid JSON?

antora/docs/modules/ROOT/pages/release_policy.adoc
```diff
@@ -0,0 +1,1707 @@
...
+
+The Enterprise Contract CLI now places the attestation data in a different location. This check fails if the expected new format is not found.
+
+*Solution*: Use a newer version of the Enterprise Contract CLI.
+
+* Rule type: [rule-type-indicator failure]#FAILURE#
+* FAILURE message: `Deprecated policy attestation format found`
+* Code: `attestation_type.deprecated_policy_attestation_format`
+* Effective from: `2023-08-31T00:00:00Z`
+* https://github.com/enterprise-contract/ec-policies/blob/main/policy/release/attestation_type.rego#L72[Source, window="_blank"]
```
lcarva, line 232: I see what you mean about this easily getting out of date. It's not tied to a particular commit ID so even when things are working properly, there's a period of time where this will inevitably point to the wrong place. I think that's ok. This is a nice-to-have feature IMO. We can try to come up with something clever in the future if needed.

lcarva (approved):

docs/.vscode/launch.json
zregvart: Did not want to commit that...

antora/docs/modules/ROOT/pages/release_policy.adoc
zregvart: Tweaked it a bit, might be okay now

zregvart:
Preview is here: https://ecd3e852.test-alm.pages.dev/docs/ec-policies/release_policy
I'll probably merge this today.

PR #968, opened 2024-04-19
Support reading base images from SBOM

lcarva:
Ref: EC-538


zregvart (approved):
LGTM, linter doesn't agree...

PR #969, opened 2024-04-19
Ignore linting on .git directory

lcarva:


simonbaird (approved):

simonbaird:
I think `$(LICENSE_IGNORE)` is not set anywhere, which is a copy/paste bug from a long time ago. Could consider cleaning that up here, either removing it or using it for the `.git/**`

lcarva:
> I think `$(LICENSE_IGNORE)` is not set anywhere, which is a copy/paste bug from a long time ago. Could consider cleaning that up here, either removing it or using it for the `.git/**`
> 
> Like so: https://github.com/enterprise-contract/ec-cli/blob/main/Makefile#L138 .

Nice! That's much better. Updated

PR #975, opened 2024-04-22
Add policy check to verify if the digest was scanned

yashvardhannanavati:
This reverts commit 7ca728010b10f4dd58a4be12cf606ab4900b693b.


yashvardhannanavati:
can we please re-merge this after https://github.com/enterprise-contract/ec-cli/pull/1486 is released?

policy/release/test.rego
```diff
@@ -227,6 +228,34 @@ deny contains result if {
...
+# custom:
+#   short_name: test_all_images
+#   failure_msg: Test '%s' did not process image with digest '%s'.
+#   solution: >-
+#     Found an image not processed by a task. Make sure that the task
+#     processes and includes the image digest of all images in the
+#     `IMAGES_PROCESSED` result.
+#   collections:
+#   - redhat
+#
```
zregvart, line 245: Let's put an effective_on, so the rule is a warning at first and after this date a violation
```suggestion
#   effective_on: 2024-05-29T00:00:00Z
#
```

yashvardhannanavati: Good idea! Done

zregvart (approved):

zregvart:
@enterprise-contract/devs I think this should be okay to merge, comments?

lcarva:
It has an effective_on date for a month for now. I think it's ok to merge this.

PR #977, opened 2024-04-26
Clarifying required-tasks usage

joejstuart:
Document the pipeline-required-tasks and required-tasks keys and usage.

https://issues.redhat.com/browse/EC-71

### DIFF
--- a/policy/release/tasks.rego
+++ b/policy/release/tasks.rego
@@ -2,16 +2,26 @@
 # METADATA
 # title: Tasks
 # description: >-
-#   Enterprise Contract expects that a set of tasks were included
-#   in the pipeline build for each image to be released.
-#   This package includes a set of rules to verify that the expected
-#   tasks ran in the pipeline when the image was built.
-#   Required tasks are listed by the names given to them within the Tekton
-#   Bundle image. Optionally invocation parameter of a Task can be also
-#   mandated by including the name and the value in square brackets following
-#   the name of the task. For example: ``name[PARAM=val]``. Only single parameter
-#   is supported, to assert multiple parameters repeat the required task
-#   definition for each parameter seperately.
+#   Enterprise Contract expects that a set of tasks were
+#   included in the pipeline build for each image to be
+#   released. This package includes a set of rules to verify
+#   that the expected tasks ran in the pipeline when the image
+#   was built.
+#   Required tasks for a pipeline are specified in a configuration
+#   file provided at runtime. This file features two primary
+#   sections: pipeline-required-tasks and required-tasks.
+#   The pipeline-required-tasks section lists tasks specific
+#   to particular pipelines, while required-tasks details a
+#   default or baseline set of tasks. If your pipeline corresponds
+#   to an entry under pipeline-required-tasks, those tasks will be prioritized;
+#   otherwise, the system will default to the tasks listed under
+#   required-tasks. Required tasks are listed by the names given to
+#   them within the Tekton Bundle image. Optionally invocation parameter
+#   of a Task can be also mandated by including the name and the value
+#   in square brackets following the name of the task. For example: name[PARAM=val].
+#   Only single parameter is supported, to assert multiple parameters repeat the
+#   required task definition for each parameter seperately.
+#
 #
 package policy.release.tasks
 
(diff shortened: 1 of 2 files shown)


zregvart:
@joejstuart make sure to run `make generate-docs` to update the (now) generated documentation

policy/release/tasks.rego
```diff
@@ -2,16 +2,26 @@
...
-#   mandated by including the name and the value in square brackets following
-#   the name of the task. For example: ``name[PARAM=val]``. Only single parameter
-#   is supported, to assert multiple parameters repeat the required task
-#   definition for each parameter seperately.
+#   Enterprise Contract expects that a set of tasks were
+#   included in the pipeline build for each image to be
+#   released. This package includes a set of rules to verify
+#   that the expected tasks ran in the pipeline when the image
+#   was built.
+#   Required tasks for a pipeline are specified in a configuration
```
lcarva, line 10: I believe "data source" is more accurate than configuration file.

policy/release/tasks.rego
```diff
@@ -2,16 +2,26 @@
...
+#   file provided at runtime. This file features two primary
+#   sections: pipeline-required-tasks and required-tasks.
```
lcarva, line 12: Not sections, but rule data keys?

policy/release/tasks.rego
```diff
@@ -2,16 +2,26 @@
...
+#   The pipeline-required-tasks section lists tasks specific
+#   to particular pipelines, while required-tasks details a
+#   default or baseline set of tasks. If your pipeline corresponds
+#   to an entry under pipeline-required-tasks, those tasks will be prioritized;
+#   otherwise, the system will default to the tasks listed under
+#   required-tasks. Required tasks are listed by the names given to
+#   them within the Tekton Bundle image. Optionally invocation parameter
```
lcarva, line 19: The Task name is given by its definition, regardless if it comes from a Tekton Bundle, git repo, etc.

policy/release/tasks.rego
```diff
@@ -2,16 +2,26 @@
...
+#   included in the pipeline build for each image to be
+#   released. This package includes a set of rules to verify
+#   that the expected tasks ran in the pipeline when the image
+#   was built.
+#   Required tasks for a pipeline are specified in a configuration
+#   file provided at runtime. This file features two primary
+#   sections: pipeline-required-tasks and required-tasks.
+#   The pipeline-required-tasks section lists tasks specific
+#   to particular pipelines, while required-tasks details a
+#   default or baseline set of tasks. If your pipeline corresponds
```
lcarva, line 15: > pipeline corresponds to an entry under pipeline-required-tasks

Can you elaborate on what this means?

joejstuart: I tried to elaborate more just before that to give more context. Let me know if that helps. https://github.com/enterprise-contract/ec-policies/pull/977/commits/c70865d4371d1ed3a3b53741cf6cba3b58bdbb17

simonbaird: Would this be better fixed by moving the `required-tasks` list to `pipeline-required-tasks.default` ?

simonbaird: We could deprecate and keep the original required-tasks lists forever, but it would make the explanation much easier, right?

joejstuart: I like the idea. The only downside is what if a pipeline name is `default`? We'd have to protect that somehow.

simonbaird: :shrug: `__default__` then?

lcarva: Renaming the rule data key is outside the scope of EC-71. But is an important point. Their current names don't make sense. I also think that we may have a better way to do this now. For example, it is now possible to specify an [or-list](https://github.com/release-engineering/rhtap-ec-policy/blob/c9be04198008e84c49f8575b69a0a6af2a7d78eb/data/required_tasks.yml#L20) for each required Task. This should help consolidate most of the entries under `pipeline-required-tasks`, leaving fbc as its own thing. We may take this further and simply have one list for fbc and one list of anything else like we do for deprecated labels (see [fbc](https://github.com/release-engineering/rhtap-ec-policy/blob/c9be04198008e84c49f8575b69a0a6af2a7d78eb/data/rule_data.yml#L105) vs [everything else](https://github.com/release-engineering/rhtap-ec-policy/blob/c9be04198008e84c49f8575b69a0a6af2a7d78eb/data/rule_data.yml#L49)).

lcarva: The updated docs look great. Thank you!

lcarva (approved):

PR #992, opened 2024-05-09
Fix some misformatted trusting tasks docs

simonbaird:
Ref: https://issues.redhat.com/browse/EC-621


simonbaird:
Looks good according to GitHub's renderer: https://github.com/enterprise-contract/ec-policies/blob/3d9a7bed1ed08263005cee818a033074f0261ced/antora/docs/modules/ROOT/pages/trusting_tasks.adoc .

Let's merge.

PR #993, opened 2024-05-10
Prevent release with quay expires-after set

simonbaird:
The quay.expires-after label is used by quay.io to indicate the image doesn't need to be kept indefinitely. The goal is to prevent unintentionally releasing images with that label is set.

There's some dependency on the extra rule data name/value chosen in EC-618 so this is WIP for now.

Ref: https://issues.redhat.com/browse/KFLUXBUGS-1223

### DIFF
--- a/policy/release/quay_expiration.rego
+++ b/policy/release/quay_expiration.rego
@@ -1,0 +1,56 @@
+#
+# METADATA
+# title: Quay expiration
+# description: >-
+#   Policies to prevent releasing an image to quay that has a quay
+#   expiration date. In Konflux images with an expiration date are
+#   produced by "on-pr" build pipelines, i.e. pre-merge CI builds,
+#   so this is intended to prevent accidentally releasing a CI build.
+#
+package policy.release.quay_expiration
+
+import rego.v1
+
+import data.lib
+
+# METADATA
+# title: Expires label
+# description: >-
+#   Check the image metadata for the presence of a "quay.expires-after"
+#   label. If it's present then produce a violation. This check is enforced
+#   only for a "release" pipeline, as determined by the value of the
+#   `pipeline_intention` rule data.
+# custom:
+#   short_name: expires_label
+#   failure_msg: The image has a 'quay.expires-after' label set to '%s'
+#   solution: >-
+#     Make sure the image is built without setting the "quay.expires-after" label. This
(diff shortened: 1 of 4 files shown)


simonbaird:
I didn't test it with a real image yet.

policy/lib/rule_data.rego
```diff
@@ -80,6 +80,8 @@ rule_data_defaults := {
 		"features.operators.openshift.io/token-auth-azure",
 		"features.operators.openshift.io/token-auth-gcp",
 	],
+	# In a Konflux release pipeline this will be set to "release"
+	"pipeline_intention": "",
```
simonbaird, line 84: Could change based on https://issues.redhat.com/browse/EC-618

simonbaird: This is part of #999 now, and actually it's not necessary for this.

policy/release/quay_expires_after.rego
```diff
@@ -0,0 +1,47 @@
...
+#     Make sure the image is built without setting the "quay.expires-after" label
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	# This is where we can access the image labels
+	some label_name, label_value in input.image.config.Labels
+
+	# This only applies if we're intending to release the image
+	lib.rule_data("pipeline_intention") == "release"
```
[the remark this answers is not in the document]
simonbaird: Since we settled on the key name and value we intend to use, this is no longer blocked.

zregvart (approved):

zregvart:
Be wary both this and #999 modify the same docs so one of them most likely would need to be rebased on top of the other.

policy/release/quay_expiration.rego
```diff
@@ -0,0 +1,55 @@
+#
+# METADATA
+# title: Labels
```
lcarva, line 3: This seems a bit ambiguous with the already existing [labels](https://enterprisecontract.dev/docs/ec-policies/release_policy.html#labels_package) package.

lcarva (approved):

policy/release/quay_expiration.rego
simonbaird: Oops, copy paste snafu I think. Thanks!

simonbaird:
A rebase plus minor tweaks.

simonbaird:
Let's merge.

PR #999, opened 2024-05-14
Ensure schedule checks apply only to releases

simonbaird:
Make it so the "no Friday deploys" rule (for example) is used only for EC checks in release pipelines, and not for EC checks in regular pre-merge or post-merge CI pipelines.

Whether we're in a release pipeline is determined by checking the value of the new `pipeline_intention` rule data key.

See also:
* https://github.com/enterprise-contract/ec-cli/pull/1604
* https://github.com/konflux-ci/release-service-catalog/pull/409

Notes:
* Set a null default for the pipeline_intention rule data value since the standard default of an empty list seems less useful.
* This should allow the revert here to be unreverted: https://github.com/release-engineering/rhtap-ec-policy/pull/12

Ref: https://issues.redhat.com/browse/EC-293

### DIFF
--- a/policy/release/schedule.rego
+++ b/policy/release/schedule.rego
@@ -14,7 +14,9 @@ import data.lib
 # title: Weekday Restriction
 # description: >-
 #   Check if the current weekday is allowed based on the rule data value from the key
-#   `disallowed_weekdays`. By default, the list is empty in which case *any* weekday is allowed.
+#   `disallowed_weekdays`. By default, the list is empty in which case *any* weekday is
+#   allowed. This check is enforced only for a "release" pipeline, as determined by
+#   the value of the `pipeline_intention` rule data.
 # custom:
 #   short_name: weekday_restriction
 #   failure_msg: '%s is a disallowed weekday: %s'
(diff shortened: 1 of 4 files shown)


simonbaird:
The branch name is a bit wrong here, but never mind.

policy/release/schedule.rego
```diff
@@ -64,10 +66,19 @@ deny contains result if {
...
 	some error in _rule_data_errors
 	result := lib.result_helper(rego.metadata.chain(), [error])
 }
 
+# We want these checks to apply only if we're doing a release. Detect that by checking
+# the `pipeline_intention` value which is set to "release" for Konflux release pipelines.
+default _schedule_restrictions_apply := false
+
+_schedule_restrictions_apply if {
+	lib.rule_data("pipeline_intention") == "release"
```
lcarva, line 79: It would be great to include something in the docs that indicate that these rules only apply if `pipeline_intention` is set to `release`.

simonbaird: Thanks, good idea.

simonbaird: Done in latest revision.

PR #1014, opened 2024-05-30
Add trusted_artifacts task policy package

lcarva:
These policy rules are meant to assist Task authors in conforming to the conventions expected by Trusted Artifacts. Mainly, it is important to follow the pattern of using the `_ARTIFACT` suffix for any Trusted Artifact parameter and result.

This PR also includes a small refactoring so we can re-use the helper functions previously defined in the `step_image_registries` package.

NOTE: These are not added to the `redhat` collection because, currently, collections are only supported on the policy rules in the `release` namespace.

Ref: EC-371

### DIFF
--- a/antora/docs/modules/ROOT/pages/task_policy.adoc
+++ b/antora/docs/modules/ROOT/pages/task_policy.adoc
@@ -22,7 +22,7 @@ Confirm the `allowed_step_image_registry_prefixes` rule data was provided, since
 * Rule type: [rule-type-indicator failure]#FAILURE#
 * FAILURE message: `%s`
 * Code: `step_image_registries.step_image_registry_prefix_list_provided`
-* https://github.com/enterprise-contract/ec-policies/blob/{page-origin-refhash}/policy/task/step_image_registries.rego#L42[Source, window="_blank"]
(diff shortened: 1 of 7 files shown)


robnester-rh:
I'm a little confused perhaps, but there's a lot of mention of _prefix_ but `_ARTIFACT` is only shown as a _suffix_ in examples?

antora/docs/modules/ROOT/pages/task_policy.adoc
```diff
@@ -81,3 +81,31 @@ Confirm the task definition has the kind "Task".
...
+* Package name: `trusted_artifacts`
+* Package full path: `policy.task.trusted_artifacts`
+
+[#trusted_artifacts__parameter]
+=== link:#trusted_artifacts__parameter[Parameter]
+
+Trusted Artifact parameters follow the expected naming convention.
+
+* Rule type: [rule-type-indicator failure]#FAILURE#
+* FAILURE message: `The parameter %q of the Task %q does not use the _ARTIFACT prefix`
```
robnester-rh, line 99: Should this be:
```
* FAILURE message: `The parameter %q of the Task %q does not use the _ARTIFACT suffix?`
```

robnester-rh (approved):
LGTM

lcarva:
Merging. I'm gonna enable this in the build-definitions repo next.

PR #1019, opened 2024-06-03
Add trusted.trusted_parameters policy rule

lcarva:
This policy rule is responsible for identifying when a trusted builder Task uses the result from an untrusted Task.

The parameters inspected are scoped to values that look like a digest.This is because, at least with the SLSA Provenance 0.2 format, we cannot easily tell where a paramter came from (Pipeline parameter, default Task Parameter, hard-coded, from another Task, etc). The new policy rule makes an assumption that if a parameter includes a digest, that digest must be accounted for within the Pipeline.

This pull request also includes two other minor commits. One to add supporting helper functions, and the other to tidy up one of the existing test cases.

Fixes #1019
Ref: EC-658

### DIFF
--- a/policy/release/lib/attestations.rego
+++ b/policy/release/lib/attestations.rego
@@ -138,3 +138,28 @@ task_succeeded(name) if {
 	task := task_in_pipelinerun(name)
 	task.status == "Succeeded"
 }
+
+# param_values expands the value into a list of values as needed. This is useful when handling
+# parameters that could be of type string or an array of strings.
+param_values(value) := {value} if {
+	is_string(value)
+} else := values if {
+	is_array(value)
+	values := {v | some v in value}
+} else := values if {
+	is_object(value)
+	values := {v | some k, v in value}
+}
+
+# result_values expands the value of the given result into a list of values. This is useful when
+# handling results that could be of type string, array of strings, or an object.
+result_values(result) := value if {
+	result.type == "string"
+	value := {result.value}
+} else := value if {
+	result.type == "array"
+	value := {v | some v in result.value}
+} else := value if {
+	result.type == "object"
+	value := {v | some k, v in result.value}
+}
--- a/policy/release/lib/attestations.rego
+++ b/policy/release/lib/attestations.rego
@@ -148,7 +148,7 @@ param_values(value) := {value} if {
 	values := {v | some v in value}
 } else := values if {
 	is_object(value)
-	values := {v | some k, v in value}
+	values := {v | some v in value}
 }
 
 # result_values expands the value of the given result into a list of values. This is useful when
@@ -161,5 +161,5 @@ result_values(result) := value if {
 	value := {v | some v in result.value}
 } else := value if {
 	result.type == "object"
-	value := {v | some k, v in result.value}
+	value := {v | some v in result.value}
 }
--- a/policy/release/trusted_task.rego
+++ b/policy/release/trusted_task.rego
@@ -16,6 +16,8 @@ import data.lib.tkn
 
 _supported_ta_uris_reg := {"oci:.*@sha256:[0-9a-f]{64}"}
 
+_digest_patterns := {`sha256:[0-9a-f]{64}`}
+
 # METADATA
 # title: Task references are pinned
 # description: >-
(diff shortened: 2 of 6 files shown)


policy/release/trusted_task.rego
```diff
@@ -145,6 +147,36 @@ deny contains result if {
 	result := lib.result_helper(rego.metadata.chain(), [])
 }
 
+# METADATA
+# title: Trusted parameters
+# description: >-
+#   Confirm certain parameters provided to each builder Task have come from trusted Tasks.
+# custom:
+#   short_name: trusted_parameters
```
lcarva, line 155: I considered an alternative to introducing a new policy rule. We could instead extend how the `_artifact_chain` is computed to take into account this logic. I was half-way through the implementation when I realized that this would exclude the check when Trusted Artifacts are not in use.  Maybe it's something we move towards in the future, but I do like the benefit of applying this rule in call cases.

lcarva:
Need to debug the acceptance test. I can't seem to reproduce it locally :disappointed:

policy/release/lib/attestations.rego
```diff
@@ -138,3 +138,28 @@ task_succeeded(name) if {
...
+# param_values expands the value into a list of values as needed. This is useful when handling
+# parameters that could be of type string or an array of strings.
+param_values(value) := {value} if {
+	is_string(value)
+} else := values if {
+	is_array(value)
+	values := {v | some v in value}
+} else := values if {
+	is_object(value)
+	values := {v | some k, v in value}
```
zregvart, line 151: nit
```suggestion
	values := {v | some v in value}
```

policy/release/lib/attestations.rego
```diff
@@ -138,3 +138,28 @@ task_succeeded(name) if {
...
+# handling results that could be of type string, array of strings, or an object.
+result_values(result) := value if {
+	result.type == "string"
+	value := {result.value}
+} else := value if {
+	result.type == "array"
+	value := {v | some v in result.value}
+} else := value if {
+	result.type == "object"
+	value := {v | some k, v in result.value}
```
zregvart, line 164: nit
```suggestion
	value := {v | some v in result.value}
```

policy/release/trusted_task.rego
```diff
@@ -145,6 +147,37 @@ deny contains result if {
 	result := lib.result_helper(rego.metadata.chain(), [])
 }
 
+# METADATA
+# title: Trusted parameters
+# description: >-
+#   Confirm certain parameters provided to each builder Task have come from trusted Tasks.
```
zregvart, line 153: I think we want to check every Task on the chain not just the builder Tasks. Perhaps one can manipulate the prefetch-dependencies Task via a parameter from an untrusted Task? Not really via digests as it currently stands, but still...

policy/release/trusted_task.rego
zregvart: Yeah I think this is a generic constraint, we don't want parameters from untrusted tasks to go to any trusted task when not using trusted artifacts; or to any trusted tasks on the artifact chain when using trusted artifacts. Not sure if we need two variants of this rule for those two cases, perhaps...

zregvart (approved):
With the caveat that this handles digests currently I think this is good to go, perhaps we need to ponder a bit if there are cases when we want to check the non-digest parameters as well in a followup to this...

policy/release/trusted_task.rego
lcarva: Let me ponder on this for a bit. I think if someone is using Trusted Artifacts, it's ok if they have certain Tasks that do this as long as it doesn't affect the builder Tasks.

policy/release/lib/attestations.rego
lcarva: I didn't this worked. Thanks!

lcarva:
@zregvart, not sure if you want to have another look before merging.

zregvart (approved):
:rocket:

policy/release/trusted_task.rego
zregvart: Here is an example, I add an untrusted Task `pick-proxy` which will emit a `HTTP_PROXY` result, i.e. a HTTP proxy that will be used for the git-clone or prefetch-dependencies Task, now in good faith this Task was added to choose a correct proxy, say depending on where the build is running. If that Task is compromised then the sources or the dependencies can be influenced. That's why I think we should look at the whole chain. This is in the end similar to how the non Trusted Artifacts pipeline is examined where we say all Tasks need to be trusted.

lcarva: It's a valid point. I just don't think we have enough information in the SLSA Provenance to do this. We need the Pipeline definition. I'll file a follow up spike for this.

lcarva: Filed https://issues.redhat.com/browse/EC-670

PR #1020, opened 2024-06-04
Deprecated CVE result name is ok if new is present

zregvart:
If the scanning task emits both the new result name (`SCAN_OUTPUT`) and the deprecated one `CLAIR_SCAN_RESULT` the `deprecated_cve_result_name` warning should not be emitted.

Also the second warning about the same issue with the code `deprecated_unpatched_cve_result_name` is removed to reduce the noise.


lcarva (approved):

simonbaird (approved):
Nice.

PR #1024, opened 2024-06-05
Allow image URLs and digests to contain \n

lcarva:
The type-hinting results that Chains expects, e.g. IMAGE_URL, sometimes contain a trailing slash. Chains allows for this behavior by always trimming the spaces around such values. The same should be done by the policy rules.

These changes also use the policy input extracted from the latest golden-container image in the acceptance tests. This now contains a Task that builds the source container image which highlights the bug fixed in this PR.


joejstuart (approved):
Thanks!

PR #1028, opened 2024-06-06
Add trusted_artifacts.workspaces rule to task ns

lcarva:
This commit adds a new policy rule, `trusted_artifacts.workspaces`, under the `task` namespace. Its purpose is to disallow general purpose workspaces from being used by Tasks that implement the Trusted Artifacts pattern.

Ref: EC-658

### DIFF
--- a/policy/task/trusted_artifacts.rego
+++ b/policy/task/trusted_artifacts.rego
@@ -38,6 +38,26 @@ deny contains result if {
 	result := lib.result_helper(rego.metadata.chain(), [result_name, k8s.name_version(input)])
 }
 
+# METADATA
+# title: Workspace
+# description: >-
+#   Tasks that implement the Trusted Artifacts pattern should not allow general purpose workspaces
+#   to share data. Instead, data should be passed around via Trusted Artifacts. Workspaces used for
+#   other purposes, e.g. provide auth credentials, are allowed. Use the rule data key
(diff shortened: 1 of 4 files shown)


policy/task/trusted_artifacts.rego
```diff
@@ -59,3 +79,14 @@ _ta_results contains result_name if {
 _has_ta_suffix(name) if endswith(name, "_ARTIFACT")
 
 _is_ta_step(step) if contains(step.image, "trusted-artifacts")
+
+_uses_trusted_artifacts(task) if {
```
simonbaird, line 83: I'm wondering how bulletproof this detection method is. Could it produce false positives? Maybe add some commentary here to explain the assumptions being made and if/how it might reach the wrong conclusion.

simonbaird: Oh, I realized we're checking a task definitions not attestation after writing that. So it's probably no big deal. Even so some comments would be useful.

simonbaird (approved):

policy/task/trusted_artifacts.rego
lcarva: Good idea. Added a comment explaining the approach.

lcarva:
This was meant to reference EC-258 instead. My bad.

PR #1041, opened 2024-06-13
Support using arbitrary & specific ec version

simonbaird:
Optionally specify a git ref that will be used as the ec version for running the test suite using make.

This lets me confirm that v0.1-alpha is not compatible with the current latest policies and v0.2 is compatible[1].

```
EC_REF=release-v0.1-alpha make ec-version quiet-test #=> fails
EC_REF=release-v0.2 make ec-version quiet-test #=> passes
EC_REF=release-v0.4 make ec-version quiet-test #=> passes
```

You could use a specific git sha instead of a branch if you want to.

Ref: [EC-695](https://issues.redhat.com/browse/EC-695)

[1] ...probably. The test suite passes at least.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,16 @@ POLICY_DIR=./policy
 # Use go run so we use the exact pinned versions from the mod file.
 # Use ec for the opa and conftest commands so that our custom rego
 # functions are available.
-EC=go run github.com/enterprise-contract/ec-cli
+ifndef EC_REF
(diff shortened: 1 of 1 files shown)


lcarva (approved):
Very nice! :taco:

zregvart (approved):

simonbaird:
Rebased and fixed the commit signing. Will merge when green.

PR #1030, opened 2024-06-10
Disallow dependencies based on attributes in SBOM

zregvart:
Given rule data set to:

```
dissalowed_attributes:
  - name: attribute
    value: value
  - name: another
```

Will disallow any dependency that has properties "attribute" with value value" or "another" (without a value) set.

Reference: https://issues.redhat.com/browse/EC-638

PR #1046, opened 2024-06-20
Fix typo in rule data attributes key

simonbaird:
A small fixup for PR #1030

Ref: https://issues.redhat.com/browse/EC-638


robnester-rh (approved):
LGTM

PR #1047, opened 2024-06-21
Merge rule_data.trusted_tasks into trusted_tasks

simonbaird:
Ernesto (reasonably IMO) assumed that this would work and I can't think of a good reason that it shouldn't, especially since it's such a small change.

It's true that we're moving towards the "trusted artifacts", aka "-oci-ta" tasks which mean we can permit custom tasks without requiring they they are trusted. But still, it's reasonable for teams to want to ensure their custom tasks are trusted, so I like the idea of make it more convenient.

Ref: [EC-709](https://issues.redhat.com/browse/EC-709)

### DIFF
--- a/antora/docs/modules/ROOT/pages/release_policy.adoc
+++ b/antora/docs/modules/ROOT/pages/release_policy.adoc
@@ -1632,7 +1632,7 @@ Each test result is expected to have a `results` key. Verify that the `results` 
 [#trusted_task_package]
 == link:#trusted_task_package[Trusted Task checks]
 
-This package is used to verify all the Tekton Tasks involved in building the image are trusted. Trust is established by comparing the Task references found in the SLSA Provenance with the pre-defined list of trusted Tasks. The list is customized via the `trusted_tasks` rule data key.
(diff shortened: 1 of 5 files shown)


robnester-rh (approved):
LGTM

zregvart (changes requested):
We could do with docs on this, perhaps [here](https://enterprisecontract.dev/docs/ec-policies/release_policy.html#trusted_task__trusted). And we want to schema check the value as it is relatively complex

simonbaird:
I'll do some more work on this based on Zoran's suggestions.

zregvart (approved):

zregvart (approved):
We can also do the schema check and the docs in a followup...

simonbaird:
I did update the documentation a little, see the second commit.

Schema validation seems like a good idea, and we are doing similar elsewhere (`git grep json.match_schema`), but I think it could be done in a future PR rather than hold up this one.

I'll pull this out of draft.

simonbaird:
Let's merge.

PR #1059, opened 2024-07-02
Implement label checks using builtins

zregvart:
Now image manifest and image config is fetched using the `ec.oci.image_manifest` and `ec.oci.blob` builtin EC CLI functions. This allows the policy to handle failures to fetch either as policy violations.

Reference: EC-454


zregvart:
There is a slight implementation difference of note here: `org.opencontainers.image.base.digest` is not optional, I thought it would be better to have a deterministic policy, i.e. look at the exact parent image rather than at some (perhaps newer) parent image.

policy/release/labels.rego
```diff
@@ -44,6 +45,47 @@ warn contains result if {
 	)
 }
 
+# METADATA
+# title: Inaccessible image manifest
+# description: >-
+#   The image manfest is not accessible.
```
simonbaird, line 51:
```suggestion
#   The image manifest is not accessible.
```

policy/release/labels.rego
```diff
@@ -146,15 +190,90 @@ deny contains result if {
...
+#     Check the provided authentication configuration and the credentials
+#     within it.
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	manifest := ec.oci.image_manifest(input.image.ref)
+
+	name := manifest.annotations["org.opencontainers.image.base.name"]
+	digest := manifest.annotations["org.opencontainers.image.base.digest"]
```
simonbaird, line 210: If these annotations are not present, do we expect we already produced a "required_labels" violation? That would make sense to me, but could consider a comment to mention that.

policy/release/labels.rego
```diff
@@ -146,15 +190,90 @@ deny contains result if {
...
+#     within it.
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	manifest := ec.oci.image_manifest(input.image.ref)
+
+	parent_name := manifest.annotations["org.opencontainers.image.base.name"]
+	parent_digest := manifest.annotations["org.opencontainers.image.base.digest"]
+	parent_ref = sprintf("%s@%s", [parent_name, parent_digest])
```
simonbaird, line 236: Could consider moving these into a parent_ref_from_manifest (or something) helper rule to reduce repetition.

simonbaird (approved):

policy/release/labels.rego
zregvart: :+1: found a nice refactor that can be done here

policy/release/labels.rego
zregvart: `required_labels` should be independent of the parent labels correlation checks, if these are not present that means the image has no (recorded) parent

PR #1060, opened 2024-07-04
Add disallowed package external references rule

ralphbean:
The purpose of this is to ban packages that have an externalReference in their cyclonedx sbom that match some regex in the policy rule data.

This is meant to be useful in particular in conjunction with the [oci-copy](https://github.com/konflux-ci/build-definitions/tree/main/task/oci-copy) task. That task permits copying arbitrary binary content for the purposes of distributing it via an OCI registry. It is not safe in general. Policies should generally forbid it, and not treat it as a trusted task.

In scenarios where it is deemed permissible, the task should be trusted, but only if this rule is also in place which will put limits on what artifacts can be redistributed.

Ref: https://issues.redhat.com/browse/EC-676

### DIFF
--- a/policy/release/sbom_cyclonedx.rego
+++ b/policy/release/sbom_cyclonedx.rego
@@ -119,6 +119,36 @@ deny contains result if {
 	result := lib.result_helper(rego.metadata.chain(), [component.purl, property.name, msg])
 }
 
+# METADATA
+# title: Disallowed package external references
+# description: >-
+#   Confirm the CycloneDX SBOM contains only packages without disallowed
+#   external references. By default all external references are allowed. Use the
+#   "disallowed_external_references" rule data key to provide a list of type-pattern pairs
+#   that forbid the use of an external reference of the given type where the reference url
+#   matches the given pattern.
+# custom:
+#   short_name: disallowed_package_external_references
+#   failure_msg: Package %s has reference %q of type %q which is disallowed%s
+#   solution: Update the image to not use a package with a disallowed external reference.
+#   collections:
+#   - redhat
+#   - policy_data
+#   effective_on: 2024-07-31T00:00:00Z
+deny contains result if {
+	some s in sbom.cyclonedx_sboms
+	some component in s.components
+	some reference in component.externalReferences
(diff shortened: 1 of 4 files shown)


policy/release/sbom_cyclonedx.rego
```diff
@@ -119,6 +119,36 @@ deny contains result if {
...
+deny contains result if {
+	some s in sbom.cyclonedx_sboms
+	some component in s.components
+	some reference in component.externalReferences
+	some disallowed in lib.rule_data(_rule_data_external_references_key)
+
+	reference.type == disallowed.type
+	regex.match(object.get(disallowed, "url", ""), object.get(reference, "url", ""))
+
+	msg := regex.replace(object.get(reference, "value", ""), `(.+)`, ` to "$1"`)
```
zregvart, line 147: Would be nice to have the `msg` generation covered in a test.

zregvart (approved):

policy/release/sbom_cyclonedx.rego
ralphbean: Thanks for pointing this out. I think `value` is never defined there.

I updated it in [1bc7ca9](https://github.com/enterprise-contract/ec-policies/pull/1060/commits/1bc7ca9d2c59d665f38435a52ea7716af8ad7500) to be more useful.

zregvart (approved):

chmeliik:
A bit late to the conversation, but is a denylist the way to go?

>  That task permits copying arbitrary binary content

> only if this rule is also in place which will put limits on what artifacts can be redistributed.

If we ban a few urls, it will still be mostly arbitrary content

zregvart:
Yes, we usually do allow-lists as a preference, it is not a hard stance I think

@chmeliik would you prefer an allow-list instead? WDYT @ralphbean?

chmeliik:
> Yes, we usually do allow-lists as a preference, it is not a hard stance I think
> 
> @chmeliik would you prefer an allow-list instead? WDYT @ralphbean?

Yeah, that would make more sense to me

PR #1061, opened 2024-07-08
Script to update examples 

zregvart:
Updates the `example/data/trusted_tekton_tasks.yml` and `acceptance/samples/policy-input-golden-container.json` files. Added the updated files and aligned the golang version in `.tool-versions` file to the `go.mod`.


hack/refresh-examples.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/env bash
+# Copyright 2022 Red Hat, Inc.
```
lcarva, line 2: Can we just use `Copyright The Enterprise Contract Contributors` like we do in ec-cli?

lcarva (approved):
This is great! Can you also update the docs here: https://github.com/enterprise-contract/ec-policies/blob/main/acceptance/samples/README.md

hack/refresh-examples.sh
zregvart: Tots, not sure where I got this from...

zregvart: Actually all scripts in `hacks` are doing Red Hat, Inc... I'll do that in a separate PR...

lcarva (approved):

PR #1065, opened 2024-07-08
feat: add an explicit allowlist rule for sbom external references

ralphbean:
I decided that I needed this feature once I learned that `re2` doesn't support negative lookarounds. I intended to use the "disallowed" external packages feature with a negative regex that forbade any external reference that didn't match a small number of permitted urls. That's just not possible with golang regular expressions.

With this, I add a new rule that by default permits every package, but when rule data is provided, it will deny every artifact with packages in its sbom that do not appear explicitly in its allow list of external references.


zregvart (approved):
LGTM
cc @chmeliik looks like this is what you had in mind

ralphbean:
Thanks Zoran!

Issue #1027, opened 2024-06-05
buildah_build_taks does not handle buildah variants


lcarva:
The policy rules in the `buildah_build_task` package [only look for](https://github.com/enterprise-contract/ec-policies/blob/9bb5e5cc1443d1eafb626e459bb4ec17d71112a3/policy/release/buildah_build_task.rego#L70) the Task name `"buildah"`. However, there are other bulidah variant Tasks that should be handled, e.g. buildah-remote. See all Tasks [here](https://github.com/konflux-ci/build-definitions/tree/main/task).

We may also want to consider re-evaluating how those policy rules are implemented. Maybe we looks at all the [build_tasks](https://github.com/enterprise-contract/ec-policies/blob/9bb5e5cc1443d1eafb626e459bb4ec17d71112a3/policy/lib/tekton/task.rego#L166)? But then some may not have a `DOCKERFILE` parameter. Maybe that's ok?

**Acceptance Criteria** 
* All buildah variants are handled by the policy rules in the `buildah_build_task` package.

PR #1069, opened 2024-07-11
Redefine DOCKERFILE parameter check

zregvart:
The policy no longer requires that the buildah Task is named "buildah", but rather looks at the Tasks that build images and selects those that have the `DOCKERFILE` parameter instead. With that the rule `buildah_task_has_dockerfile_param` doesn't make much sense any more and it has been removed.

Resolves: #1027

### DIFF
--- a/policy/release/buildah_build_task.rego
+++ b/policy/release/buildah_build_task.rego
@@ -9,33 +9,7 @@ package policy.release.buildah_build_task
 import rego.v1
 
 import data.lib
-
-# METADATA
-# title: Buildah task has Dockerfile param defined
-# description: >-
-#   Verify that a DOCKERFILE parameter was provided to
-#   the buildah task.
-# custom:
-#   short_name: buildah_task_has_dockerfile_param
-#   failure_msg: >-
-#     The pipeline task %q does not contain the DOCKERFILE param. This is a requirement for the underlying task %q
-#   solution: >-
-#     Make sure the buildah task has a parameter named 'DOCKERFILE'.
-#   collections:
-#   - redhat
-#   depends_on:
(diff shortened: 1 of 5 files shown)


policy/release/buildah_build_task.rego
```diff
@@ -9,33 +9,7 @@ package policy.release.buildah_build_task
 import rego.v1
 
 import data.lib
-
-# METADATA
-# title: Buildah task has Dockerfile param defined
```
lcarva, line 14: This one was first introduced to safe guard against a builder Task not including the `DOCKERFILE` parameter. For example, if the parameter is renamed, `buildah_uses_local_dockerfile` is no-op. Unlikely, but it could happen.

Maybe we want to introduce a new policy rule under the `task` directory to verify this, and enable it [here](https://github.com/konflux-ci/build-definitions/tree/main/policies). 

WDYT?

lcarva (approved):

policy/release/buildah_build_task.rego
zregvart: Not sure, not all image-building tasks need a Dockerfile, e.g. the s2i ones. I think we'd need to classify which Tasks need a `DOCKERFILE` parameter and which do not, and I choose the simplest possible classifications: those that do have the `DOCKERFILE` parameter. Making it very trivial and pragmatic.
I'm don't know how strict is the hermetic build requirement, I'm expecting folk to run only hermetic releases in the future, so a Dockerfile from remote should not be possible, and this rule might not be needed at all...

lcarva: WFM!

PR #1067, opened 2024-07-09
Verify all operator CSV pull specs will be valid on release

yashvardhannanavati:
This commit verifies the validity of the pull specs in the operator bundle CSV by checking if they are planned to be released as a part of the input snapshot or are already accessible on the registries they reference.

resolves: CVP-4148

PR #1073, opened 2024-07-15
Expect pinned references in input snapshot for operators

yashvardhannanavati:
Instead of trying to resolve the input components, expect pinned references.

### DIFF
--- a/policy/release/olm.rego
+++ b/policy/release/olm.rego
@@ -123,6 +123,36 @@ deny contains result if {
 }
 
 # METADATA
+# title: Unpinned images in input snapshot
+# description: >-
+#   Check the input snapshot for the presence of unpinned image references.
+#   Unpinned image pull references are references to images
+#   that do not contain a digest -- uniquely identifying the version of
+#   the image being pulled.
+# custom:
+#   short_name: unpinned_snapshot_references
+#   failure_msg: The %q image reference is not pinned in the input snapshot.
+#   solution: >-
+#     Update the input snapshot replacing the unpinned image reference with pinned image
+#     reference. Pinned image reference contains the image digest.
+#   collections:
+#   - redhat
+#   effective_on: 2024-08-15T00:00:00Z
+#
+deny contains result if {
+	_release_restrictions_apply
+
+	input_image = image.parse(input.image.ref)
+	components := input.snapshot.components
+	some component in components
+	parsed_image := image.parse(component.containerImage)
+	parsed_image.repo == input_image.repo
+	parsed_image.digest == "" # unpinned image references have no digest
+
+	result := lib.result_helper_with_term(rego.metadata.chain(), [image.str(parsed_image)], image.str(parsed_image))
+}
+
+# METADATA
 # title: Unable to access images in the input snapshot
 # description: >-
 #   Check the input snapshot and make sure all the images are accessible.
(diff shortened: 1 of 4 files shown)


yashvardhannanavati:
follow-up for https://github.com/enterprise-contract/ec-policies/pull/1067

policy/release/olm.rego
```diff
@@ -122,6 +122,34 @@ deny contains result if {
 	result := lib.result_helper(rego.metadata.chain(), [error])
 }
 
+# METADATA
+# title: Unpinned images in input snapshot
+# description: >-
+#   Check the input snapshot for the presence of unpinned image references.
```
lcarva, line 128: ec-cli should always resolve image references to a digest. Is that not happening?

policy/release/olm.rego
```diff
@@ -122,6 +122,34 @@ deny contains result if {
...
+#     reference. Pinned image reference contains the image digest.
+#   collections:
+#   - redhat
+#   effective_on: 2024-08-01T00:00:00Z
+#
+deny contains result if {
+	_release_restrictions_apply
+
+	components := input.snapshot.components
+	some component in components
```
lcarva, line 146: If the application snapshot has 10 components, and two of them do not use a pinned reference, there will be 20 violations. Each component will report two violations. This is confusing.

Can we filter on components that match `input.image`? Maybe on the repo name?

policy/release/olm.rego
```diff
@@ -122,6 +122,34 @@ deny contains result if {
 	result := lib.result_helper(rego.metadata.chain(), [error])
 }
 
+# METADATA
+# title: Unpinned images in input snapshot
```
lcarva, line 126: Since ec-cli always resolves image references to a digest, this is more like a validation on the ApplicationSnapshot itself, is that correct? Maybe it belongs on a different package?

policy/release/olm.rego
```diff
@@ -122,6 +122,34 @@ deny contains result if {
...
+#   the image being pulled.
+# custom:
+#   short_name: unpinned_snapshot_references
+#   failure_msg: The %q image reference is not pinned in the input snapshot.
+#   solution: >-
+#     Update the input snapshot replacing the unpinned image reference with pinned image
+#     reference. Pinned image reference contains the image digest.
+#   collections:
+#   - redhat
+#   effective_on: 2024-08-01T00:00:00Z
```
lcarva, line 140: Let's bump this to 30 days out?

policy/release/olm.rego
yashvardhannanavati: nice catch! Done

policy/release/olm.rego
yashvardhannanavati: well, it looks like ec-cli does resolve the input component but not the entire snapshot..unless I am missing something.

Also, we need this check to run for operator bundles because that's the one using snapshot components later to verify if it matches with relatedImages. Is there any other package more suitable for this?

lcarva: > well, it looks like ec-cli does resolve the input component but not the entire snapshot..unless I am missing something.

Yes. That's what I meant to by my statement above. Maybe not as clear :smile: 

> Also, we need this check to run for operator bundles because that's the one using snapshot components later to verify if it matches with relatedImages. Is there any other package more suitable for this?

Understood. I'm just thinking that a policy that verifies ApplicationSnapshots always contain pinned image references is useful in all cases, not just for OLM.

Let's keep it here for now. We can refactor later.

lcarva (approved):

PR #1074, opened 2024-07-18
Set correct license headers

zregvart:


zregvart:
I'm going to merge this as it is proper thing to do.

lcarva:
Thanks!

PR #1090, opened 2024-07-30
Consider finding counts in addition to status

zregvart:
The status held in the `result` field of the test output is what was solely considered, so when a test resulted in ERROR (`"result": "ERROR"`) but in addition to that also had failures, e.g. `"result": "ERROR", "failures": 3` the failures were not reported.

This changes the logic so that the individual finding counts are reported as well as the outcome held in `result` field.

Reference: https://issues.redhat.com/browse/EC-620

### DIFF
--- a/policy/release/test.rego
+++ b/policy/release/test.rego
@@ -32,7 +32,7 @@ import data.lib.image
 #   - test.test_data_found
 #
 warn contains result if {
-	some test in resulted_in(lib.rule_data("failed_tests_results"))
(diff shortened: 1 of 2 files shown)


zregvart:
Not sure if this is in sync with the rule data parameterization, something to consider when reviewing...

lcarva (approved):

lcarva:
FYI: https://github.com/konflux-ci/architecture/blob/38810437c9a8f0a4952d1c9f81aaccb403de4809/ADR/0030-tekton-results-naming-convention.md#tekton-result-format-for-test_output

PR #1091, opened 2024-07-30
Add some checks for RHTAP Jenkins attestations

simonbaird:
The motivation is to have some real checks in the "@rhtap-jenkins" collection so I can use them for the RHTAP Jenkins sample templates development.

I didn't give a lot of thought to the checks themselves - the invocation id was chosen somewhat arbitrarily - or how we'll organize things when/if we have many more rules for RHTAP Jenkins builds, but let's just get something working to start with and iterate as required.

Ref: [https://issues.redhat.com/browse/EC-746](https://issues.redhat.com/browse/EC-746)


policy/release/rhtap_jenkins_checks.rego
```diff
@@ -0,0 +1,66 @@
+#
```
lcarva, line 1: Consider `rhtap_jenkins.rego` for the filename and `rhtap_jenkins` for the package name. `checks` is somewhat redundant since every rego file is performing checks.

lcarva (approved):

PR #1092, opened 2024-07-30
Verify operator CSV images are from allowed registries

nmars:
This commit adds a rule that verifies that images from operator bundle CSVs are hosted on approved registries. 

resolves: CVP-4149

### DIFF
--- a/policy/release/olm.rego
+++ b/policy/release/olm.rego
@@ -78,7 +78,7 @@ deny contains result if {
 #
 deny contains result if {
 	some manifest in _csv_manifests
-	some annotation in lib.rule_data(_rule_data_key)
+	some annotation in lib.rule_data("required_olm_features_annotations")
 	value := object.get(manifest.metadata.annotations, annotation, "")
 	not value in {"true", "false"}
 	result := lib.result_helper_with_term(rego.metadata.chain(), [annotation], annotation)
@@ -215,6 +215,39 @@ deny contains result if {
 	result := lib.result_helper_with_term(rego.metadata.chain(), [image.str(unmatched_image.ref)], image.str(unmatched_image.ref))
 }
 
+# METADATA
+# title: Images referenced by OLM bundle are from allowed registries
+# description: >-
+#   Each image referenced by the OLM bundle should match an entry in the list of prefixes
+#   defined by the rule data key `allowed_registry_prefixes` in your policy configuration.
+# custom:
+#   short_name: allowed_registries
+#   failure_msg: The %q CSV image reference is not from an allowed registry.
+#   solution: >-
+#     Use image from an allowed registry, or modify your
+#     xref:ec-cli:ROOT:configuration.adoc#_data_sources[policy configuration] to include additional registry prefixes.
+#   collections:
+#   - redhat
+#   effective_on: 2024-09-01T00:00:00Z
+#
+deny contains result if {
+	# The presence of expected rule_data verified in _rule_data_errors
+	allowed_registry_prefixes := lib.rule_data("allowed_registry_prefixes")
+
+	# Parse manifests from snapshot
+	some csv_manifest in _csv_manifests
+
+	# Parse image references from each manifest
+	all_csv_images := all_image_ref(csv_manifest)
+
+	some img in all_csv_images
+	not _image_registry_allowed(img.ref.repo, allowed_registry_prefixes)
+
+	img_str := image.str(img.ref)
+
+	result := lib.result_helper_with_term(rego.metadata.chain(), [img_str], img.ref.repo)
+}
+
 _name(o) := n if {
 	n := o.name
 } else := "unnamed"
@@ -335,11 +368,12 @@ _csv_manifests contains manifest if {
 	manifest.kind == "ClusterServiceVersion"
 }
 
-# Verify allowed_registry_prefixes is a non-empty list of strings
+# Verify allowed_registry_prefixes & required_olm_features_annotations are non-empty list of strings
 _rule_data_errors contains msg if {
 	# match_schema expects either a marshaled JSON resource (String) or an Object. It doesn't
 	# handle an Array directly.
-	value := json.marshal(lib.rule_data(_rule_data_key))
+	some rule_data_key in _rule_data_keys
+	value := json.marshal(lib.rule_data(rule_data_key))
 	some violation in json.match_schema(
 		value,
 		{
@@ -350,10 +384,13 @@ _rule_data_errors contains msg if {
 			"minItems": 1,
 		},
 	)[1]
-	msg := sprintf("Rule data %s has unexpected format: %s", [_rule_data_key, violation.error])
+	msg := sprintf("Rule data %s has unexpected format: %s", [rule_data_key, violation.error])
 }
 
-_rule_data_key := "required_olm_features_annotations"
+_rule_data_keys := [
+	"required_olm_features_annotations",
+	"allowed_registry_prefixes",
+]
 
 _subscriptions_errors contains msg if {
 	some manifest in _csv_manifests
@@ -391,4 +428,10 @@ default _release_restrictions_apply := false
 
 _release_restrictions_apply if {
 	lib.rule_data("pipeline_intention") == "release"
+}
+
+# Used by allowed_registries
+_image_registry_allowed(image_repo, allowed_prefixes) if {
+	some allowed_prefix in allowed_prefixes
+	startswith(image_repo, allowed_prefix)
 }
--- a/policy/release/olm_test.rego
+++ b/policy/release/olm_test.rego
@@ -114,11 +114,13 @@ test_all_image_ref if {
 test_all_good if {
 	lib.assert_empty(olm.deny) with input.image.files as {"manifests/csv.yaml": manifest}
 		with input.image.config.Labels as {olm.manifestv1: "manifests/"}
+		with data.rule_data.allowed_registry_prefixes as ["registry.io"]
 }
 
 test_all_good_custom_dir if {
 	lib.assert_empty(olm.deny) with input.image.files as {"other/csv.yaml": manifest}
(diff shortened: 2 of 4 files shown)


policy/release/olm.rego
```diff
@@ -215,6 +215,49 @@ deny contains result if {
...
+		some img in all_csv_images
+		img_ref = img.ref
+	}
+
+	disallowed_registry_image_refs := [img_ref |
+		some img_ref in unique_image_refs
+		not _image_registry_allowed(img_ref.repo, allowed_registry_prefixes)
+	]
+
+	some disallowed_reg_image in disallowed_registry_image_refs
```
lcarva, line 255: You could combine this line and the two blocks above:

```rego
all_csv_images := all_image_ref(csv_manifest)

some img in all_csv_images
not _image_registry_allowed(img.ref.repo, allowed_registry_prefixes)

img_str := image.str(img.ref)

result := lib.result_helper_with_term(rego.metadata.chain(), [img_str], img_str)
```

policy/release/olm.rego
```diff
@@ -215,6 +215,49 @@ deny contains result if {
...
+
+	# regal ignore:line-length
+	result := lib.result_helper_with_term(rego.metadata.chain(), [image.str(disallowed_reg_image)], image.str(disallowed_reg_image))
```
lcarva, line 258: I think for the term, the last parameter, we probably want to omit any tags and digests. Only include the repo, e.g. quay.io/foo/bar. This could then be used to create a certain policy exclusion.

wdyt?

policy/release/olm_test.rego
```diff
@@ -373,3 +373,34 @@ test_unmapped_references_none_found if {
...
+			"msg": "The \"registry.io/repository/image2@sha256:tea\" CSV image reference is not from an allowed registry.",
+			"term": "registry.io/repository/image2@sha256:tea",
+		},
+	}
+
+	# This expects failure as registry.io is not a member of allowed_registry_prefixes
+	lib.assert_equal_results(olm.deny, expected) with data.rule_data.pipeline_intention as "release"
+		with data.rule_data.allowed_registry_prefixes as ["registry.access.redhat.com", "registry.redhat.io"]
+		with input.image.config.Labels as {olm.manifestv1: "manifests/"}
+		with input.image.files as {"manifests/csv.yaml": manifest}
```
lcarva, line 405: Did you mean to create/use a `manifest` that includes references to the non-allowed registry `registry.io`?

lcarva (changes requested):

policy/release/olm.rego
```diff
@@ -215,6 +215,49 @@ deny contains result if {
...
+#   failure_msg: The %q CSV image reference is not from an allowed registry.
+#   solution: >-
+#     The image must be hosted on an allowed registry.
+#   collections:
+#   - redhat
+#   effective_on: 2024-08-15T00:00:00Z
+#
+deny contains result if {
+	# Restrict to release pipeline for now
+	_release_restrictions_apply
```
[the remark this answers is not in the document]
nmars: I think it may make sense, but I did this to be consistent with other work in the same Epic. I suggest we keep it at release for now and un-restrict this and other OLM checks in a future task.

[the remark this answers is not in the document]
lcarva: ack. Keep in mind that finding out about a violation at release time it's quite terrible. We would like to avoid this as much as possible. We should have a strong reason to make it release-only, not the other way around.

policy/release/olm.rego
```diff
@@ -335,7 +378,7 @@ _csv_manifests contains manifest if {
 	manifest.kind == "ClusterServiceVersion"
 }
 
-# Verify allowed_registry_prefixes is a non-empty list of strings
+# Verify required_olm_features_annotations is a non-empty list of strings
 _rule_data_errors contains msg if {
```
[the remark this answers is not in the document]
nmars: Newbie question... Is it more conventional to make a copy of the [entire block](https://github.com/enterprise-contract/ec-policies/blob/main/policy/release/olm.rego#L339-L354) to check `allowed_registry_prefixes` or, modify that block and pass it a list of _rule_data_keys consisting of `allowed_registry_prefixes` and `required_olm_features_annotations` ?

policy/release/olm.rego
nmars: I thought that would mess up my expected output in `test_unallowed_registries`  by `_image_registry_allowed` evaluating multiple entries of the same image ref (specified at different paths in the CSV), but it worked!  :thinking:

policy/release/olm.rego
nmars: Good idea... it would make copy/paste easier

policy/release/olm_test.rego
nmars: Yes, I wanted `olm.allowed_registries` to detect a non-allowed registry in this test, so I re-used the same manifest and excluded `registry.io` from `allowed_registry_prefixes`.

policy/release/olm.rego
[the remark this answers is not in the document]
lcarva: I think in this case, a list of rule data keys makes more sense since both must conform to the same schema.

Also, we should make a change [here](https://github.com/release-engineering/rhtap-ec-policy/blob/main/data/rule_data.yml) to define the new rule data.

policy/release/olm.rego
lcarva: Rego does a good job at de-duplicating things. The `deny` block is actually building up a set. So if you emit the same exact violation twice, it gets de-duplicated to just one.

nmars: Thanks for the info!

policy/release/olm.rego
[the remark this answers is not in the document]
nmars: I see that `required_olm_features_annotations` are missing from rhtap, but `allowed_registry_prefixes` [are there](https://github.com/release-engineering/rhtap-ec-policy/blob/74f112c8c00a03b079dc98f2ccb4ca127dcbd819/data/rule_data.yml#L5).

Also, looking at this more, since the time when the check for the features annotations were added [last December](https://github.com/enterprise-contract/ec-policies/pull/830) three of those were made optional:

```
features.operators.openshift.io/cnf
features.operators.openshift.io/cni
features.operators.openshift.io/csi
```

I can open a new separate ticket to update the rego checks for this, if one doesn't exist.

policy/release/olm.rego
```diff
@@ -215,6 +215,42 @@ deny contains result if {
...
+#   defined by the rule data key `allowed_registry_prefixes` in your policy configuration.
+# custom:
+#   short_name: allowed_registries
+#   failure_msg: The %q CSV image reference is not from an allowed registry.
+#   solution: >-
+#     Use image from an allowed registry, or modify your
+#     xref:ec-cli:ROOT:configuration.adoc#_data_sources[policy configuration] to include additional registry prefixes.
+#   collections:
+#   - redhat
+#   effective_on: 2024-0901T00:00:00Z
```
zregvart, line 231: Typo
```suggestion
#   effective_on: 2024-09-01T00:00:00Z
```

zregvart (approved):
LGTM, one typo

policy/release/olm.rego
nmars: Nice catch, thanks

policy/release/olm.rego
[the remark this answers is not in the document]
nmars: @lcarva you make a good point. After consulting a couple folks, I will remove the restriction.

policy/release/olm.rego
```diff
@@ -215,6 +215,49 @@ deny contains result if {
...
+# description: >-
+#   Images referenced by the OLM bundle are from allowed registries
+# custom:
+#   short_name: allowed_registries
+#   failure_msg: The %q CSV image reference is not from an allowed registry.
+#   solution: >-
+#     The image must be hosted on an allowed registry.
+#   collections:
+#   - redhat
+#   effective_on: 2024-08-15T00:00:00Z
```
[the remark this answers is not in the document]
nmars: Updated to 1 September

policy/release/olm.rego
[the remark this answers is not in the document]
nmars: Removed the restriction in the latest push

policy/release/olm.rego
```diff
@@ -215,6 +215,49 @@ deny contains result if {
...
+	allowed_registry_prefixes := lib.rule_data("allowed_registry_prefixes")
+
+	# Parse manifests from snapshot
+	some csv_manifest in _csv_manifests
+
+	# Parse image references from each manifest
+	all_csv_images := all_image_ref(csv_manifest)
+
+	# Filter out paths for set of unique image references
+	unique_image_refs := {img_ref |
```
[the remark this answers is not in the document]
nmars: Done

policy/release/olm.rego
nmars: Suggestion implemented

policy/release/olm.rego
nmars: Fixed. Also fixed in `antora/docs/modules/ROOT/pages/release_policy.adoc`

policy/release/olm.rego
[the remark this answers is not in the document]
nmars: I [opened a PR](https://github.com/release-engineering/rhtap-ec-policy/pull/34) for the missing features annotations rule data

policy/release/olm_test.rego
nmars: Should I do something a bit different here @lcarva?

lcarva (approved):

lcarva:
@nmars, one more thing. We require signed commits in this repo.

nmars:
@lcarva signed commit added.

lcarva (approved):

lcarva:
Thanks for the PR!

PR #1096, opened 2024-07-31
Disallow certain PLATFORM patterns

lcarva:
Ref: EC-726


lcarva:
Marking this as a draft until I confirm that this approach is expected.

lcarva:
Moving it out of draft since I've confirmed that this is the expected approach.

zregvart (approved):

PR #1116, opened 2024-08-27
Verify Task definitions contain a certain result

lcarva:
Ref: EC-794


policy/task/results.rego
```diff
@@ -0,0 +1,78 @@
...
+	some error in _rule_data_errors
+	result := lib.result_helper(rego.metadata.chain(), [error])
+}
+
+errors contains err if {
+	some required in lib.rule_data(_rule_data_key)
+	input.metadata.name == required.task
+	found := [result |
+		some result in input.spec.results
+		result.name == required.result
```
simonbaird, line 49: Probably an unlikely edge case, but should we check that a value key is present as well?

simonbaird (approved):

policy/task/results_test.rego
```diff
@@ -0,0 +1,87 @@
...
+	lib.assert_equal_results(results.deny, expected) with input as _mock_task
+		with data.rule_data as d
+}
+
+_mock_task := {
+	"apiVersion": "tekton.dev/v1",
+	"kind": "Task",
+	"metadata": {"name": "bacon"},
+	"spec": {"results": [
+		{"name": "GRILLED"},
```
simonbaird, line 82: The missing value here made me think of it..

policy/task/results.rego
lcarva: This is verifying a Task definition. There won't be a value. The result would be present in the status of a TaskRun.

policy/task/results_test.rego
simonbaird: (Disregard)

PR #1118, opened 2024-08-29
Add effective_on support to disallowed_attributes

lcarva:
Also, use the same go version in all modules. Failure to do so causes issues when generating docs.


robnester-rh (approved):
LGTM

PR #1124, opened 2024-09-03
Rework package path logic in result helper

simonbaird:
Disregarding the readme fix, the first commit is a small change to fix the reported bug, while the second commit is a more holistic refactor, imaging the existence of custom rego with any arbitrary package naming conventions.

Ref: https://issues.redhat.com/browse/EC-831

### DIFF
--- a/policy/lib/result_helper.rego
+++ b/policy/lib/result_helper.rego
@@ -24,22 +24,8 @@ _basic_result(chain, failure_sprintf_params) := {
 
 _code(chain) := code if {
 	rule_path := chain[0].path
+	pkg_name := _pkg_name(rule_path)
 
-	# rule_path examples:
-	# ["data", "some_package", "deny"]
-	# ["data", "some_package_namespace", "some_package", "deny"]
-	# ["data", "namespace", "another_namespace", "some_package", "deny"]
-	#
-	# Our convention in the ec-policies is something like ["data", "policy",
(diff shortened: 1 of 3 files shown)


zregvart (approved):
LGTM, do the one Regal ignore needed and I think we can merge this.

simonbaird:
I make a few little unit test tweaks. Thanks for the reviews. Will merge when it goes green.

policy/lib/result_helper.rego
```diff
@@ -23,13 +23,59 @@ _basic_result(chain, failure_sprintf_params) := {
...
+	# Seems to not work if I keep assigning to a single var, so
+	# that's why the many different pN vars.
+
+	# Strip off the first element which is always "data"
+	p1 := _left_strip_elements(["data"], rule_path)
+
+	# Strip off policy.release or policy.pipeline to match what ec-cli
+	# does. (There are some edge cases where the behavior is not exactly
+	# the same, but I think this version is better.)
+	p2 := _left_strip_elements(["policy", "release"], p1)
```
lcarva, line 55: I think we should try to come up with a generic approach where we don't need to rely on supporting a specific set of namespaces. Could be done later. Could you file an issue for that?

simonbaird: Filed https://issues.redhat.com/browse/EC-864 .

lcarva: Thanks!

PR #1126, opened 2024-09-04
Add helper for extracting a repo id from a purl

simonbaird:
Not sure how this will fit in to the higher level rule, but let's take a bottom-up approach and figure it out as we go.

Ref: https://issues.redhat.com/browse/EC-847

### DIFF
--- a/policy/lib/purl.rego
+++ b/policy/lib/purl.rego
@@ -1,0 +1,16 @@
+package lib.purl
+
+import rego.v1
+
+# Extract a repository id from a purl. These are often
+# called "repoids" so we'll follow that convention.
+repoid(purl) := _named_qualifier("repository_id", purl)
+
+# (Qualifiers are similar to url params)
+_named_qualifier(key, purl) := result if {
+	ec.purl.is_valid(purl)
+	parsed_purl := ec.purl.parse(purl)
+	some qualifier in parsed_purl.qualifiers
+	qualifier.key == key
+	result := qualifier.value
+}
--- a/policy/lib/purl_test.rego
+++ b/policy/lib/purl_test.rego
@@ -1,0 +1,35 @@
+package lib.purl_test
+
+import rego.v1
+
+import data.lib
+import data.lib.purl
+
+# regal ignore:line-length
+_good := "pkg:rpm/rhel/bash@5.1.8-9.el9?arch=aarch64&upstream=bash-5.1.8-9.el9.src.rpm&distro=rhel-9.4&repository_id=some-repo-id"
+
+_bad := [
+	# regal ignore:line-length
+	"pkg:rpm/rhel/bash@5.1.8-9.el9?arch=aarch64&upstream=bash-5.1.8-9.el9.src.rpm&distro=rhel-9.4&repostory_id=some-repo-id",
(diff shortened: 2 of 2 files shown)


simonbaird:
Leaving as draft since I might want to change it later.

zregvart (approved):

policy/lib/purl.rego
```diff
@@ -0,0 +1,26 @@
...
+
+import rego.v1
+
+# Extract the repoid
+repoid(purl) := _named_qualifier("repository_id", purl)
+
+#
+# Extract a named quantifier from a purl.
+# (Quantifiers are similar to url params.)
+# Let's fall back to returning null if the quantifier is not present.
```
lcarva, line 11: I think this is a bit unusual in rego. Not returning a value is usually what is done in such cases.

policy/lib/purl.rego
```diff
@@ -0,0 +1,26 @@
...
+#
+_named_qualifier(key, purl) := result if {
+	# Sanity check I guess
```
lcarva, line 14: The comments in this rule don't seem very useful :smile_cat:

policy/lib/purl.rego
```diff
@@ -0,0 +1,26 @@
...
+	ec.purl.is_valid(purl)
```
lcarva, line 15: Not sure if this is actually needed. Won't `parse` fail anyways if it is not a valid purl?

policy/lib/purl.rego
```diff
@@ -0,0 +1,26 @@
+package lib.purl
+
+import rego.v1
+
+# Extract the repoid
+repoid(purl) := _named_qualifier("repository_id", purl)
```
lcarva, line 6: I know it sounds a bit counter-intuitive, but I find that in rego an abundance of helper methods makes the code harder to maintain, especially when they are spread out across different packages. The tooling for navigating the code base is just not as sophisticated as the tooling for other languages like Go, python, etc.

I would suggest writing the policy rules first, using local helpers (same file) where needed, and only then decide if anything should be abstracted out into a lib package.

policy/lib/purl.rego
simonbaird: Yeah, I think you're right. It feels more comfortable to return null, but also less rego-ish.

policy/lib/purl.rego
simonbaird: The difference is whether or not ec-cli produces an error with log.Errorf, (if https://github.com/enterprise-contract/ec-cli/pull/1923 is merged.) My thinking is I'd rather not produce an error, so that's why I'm doing is_valid first.

policy/lib/purl.rego
simonbaird: Yeah, that's reasonable. This code could probably live in the package where the denies are defined. I'll leave it in draft while considering that.

simonbaird:
I cleaned it up based on the suggestions, specifically:

* Don't return null if there's no repoid
* Less pointless commentary
* Explanation for the purl.parse test 
* Explanation for the non-standard rule name

simonbaird:
New PR is coming soon. It will incorporate a version of this probably.

PR #1127, opened 2024-09-05
Tweaks to fetching CycloneDX SBOM

### LINKED ISSUE #1115: Use ec.oci.image_files to fetch the SBOM from the image, if needed.
Now that we have a custom rego function to fetch files from the image, let's teach the SBOM policy rules to use it.

Today, the ec-cli will fetch certain files from the image being validated if the [vendor label matches a certain value](https://github.com/enterprise-contract/ec-cli/blob/90bf60bfe0a43f7582c0c7e1d275adb2c759da39/internal/fetchers/oci/files/rhmanifest.go#L41). This is a giant hack :sweat_smile:. Because of this, we need to be careful to avoid download the image twice (one in rego and one in the ec-cli).

The relevant bits in rego are [here](https://github.com/enterprise-contract/ec-policies/blob/3629b60ae1c0a5217098b8628cc57c2095647408/policy/lib/sbom.rego#L1).

Let's take the following approach:

1. Tweak `_cyclonedx_sboms_from_image` to first look for the SBOM at `input.image.files`. If not, and if the vendor label equals `Red Hat, Inc.`, then use `oc.oci.image_files` to fetch it from the image.
2. Only execute `_cyclonedx_sboms_from_image` if `_cyclonedx_sboms_from_oci` and `_cyclonedx_sboms_from_attestations` do not produce a result.

The outcome should be that the SBOM from the image should only be used if there isn't another place to get it, and the rego only fetches it if it hasn't already been fetched by the ec-cli.

**Acceptance Criteria**

* Rego policy rules do not rely on the ec-cli to fetch the SBOM from the image.
* Rego policy rules only fetch the SBOM from within the image if they cannot find it in any other place.

lcarva:
This commit changes the mechanism of fetching CycloneDX SBOMs. It now only takes into account the SBOM from within the image if necessary. It also has additional logic to either use the prefetched SBOM from the image, or fetch it directly from rego.

This change allows us to stop fetching the SBOM from the image from the EC CLI. Paired with those changes, this commit will have a significant performance boost, most notably to large images.

Resolves #1115


policy/lib/sbom_test.rego
```diff
@@ -36,14 +36,44 @@ test_cyclonedx_sboms if {
...
 		with ec.oci.blob as mock_ec_oci_blob
+		with ec.oci.image_files as mock_ec_oci_image_files
 }
 
 mock_ec_oci_blob("registry.io/repository/image@sha256:f0cacc1a") := `{"sbom": "from oci blob"}`
+
+mock_ec_oci_image_files(
+	"registry.io/repository/image@sha256:284e3029",
+	["root/buildinfo/content_manifests/sbom-cyclonedx.json"],
+) := {sbom._sbom_image_path: {"sbom": "from lvie image"}}
```
simonbaird, line 70: I didn't grok "lvie" here. Maybe doesn't matter much but I'm curious.

simonbaird (approved):

policy/lib/sbom_test.rego
zregvart: "live"?

zregvart (approved):

PR #1132, opened 2024-09-09
Update description to only include sha1 entries

joejstuart:
The materials section includes other entries besides sha1 digests. The description leads you to believe the policy rule includes those.

https://issues.redhat.com/browse/EC-828


lcarva (approved):

lcarva:
Probably need to generate docs: `make generate-docs`

PR #1133, opened 2024-09-10
Specific version support for required Task results

zregvart:
Often new Task versions will change the emitted results, this allows for an additional `version` qualifier to support that use case.


jsztuka (approved):
Thanks for doing this in such a short notice.

lcarva (approved):

PR #1135, opened 2024-09-11
Confirm all rpms in sboms have a known repo id

simonbaird:
Note that we can't enable this now, since the sboms that Konflux creates don't yet have repo ids, so I didn't add the one relevant check to the `redhat` collection in this PR.

Ref: [EC-742](https://issues.redhat.com/browse/EC-742)


simonbaird:
I might tinker with the annotations a little more, but I think it's ready for review.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
...
+
+data_problems contains reason if {
+	is_array(known_repo_ids)
+	non_strings := [r |
+		some r in known_repo_ids
+		not is_string(r)
+	]
+	count(non_strings) > 0
+	reason := "The 'known_rpm_repositories' data is not a list of strings"
+}
```
zregvart, line 80: Usually we do this via a JSON schema check, no objection to doing it this way, just an observation.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
...
+
+# Keep the raw and parsed purl in a little struct-like
+all_parsed_rpm_purls contains p if {
+	some purl in all_rpm_purls
+	ec.purl.is_valid(purl)
+	p := {
+		"raw": purl,
+		"parsed": ec.purl.parse(purl),
+	}
```
zregvart, line 133: Perhaps, since from `parsed` only the `repository_id` is used? I think the `is_valid` can be skipped and we can rely on parsed not being nil, perhaps saves a bit of cpu time...

```suggestion
	parsed := ec.purl.parse(purl)
	parsed
	p := {
		"raw": purl,
		"repository_id": purl_repo_id(parsed),
	}
```

zregvart (commented):
LGTM, one test case that could be added is for checking that the list appears in the error message, in addition to testing `truncated_list_to_string`, but that's just nitpicking...

policy/release/repo_ids.rego
simonbaird: The one difference is we spit out an `fmt.Errorf` when an attempted parse fails.

My reasoning is that I'd prefer not to produce an error in the output.

Not sure how important that is, but that's what I was thinking about.  ([This](https://github.com/enterprise-contract/ec-cli/pull/1923/commits/77732242cc8c533ae5c4163c0d0a3bbc1b8da2db) is related, since we used to also produce that error when calling is_valid.)

zregvart: Yeah, that makes sense, let's keep `ec.purl.is_valid`. Not sure if it's worth it, the second part is that we only the `repository_id` in the returned struct, perhaps a tiny bit of cpu/memory might be saved?

simonbaird: Oh yeah good point.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
...
+	non_strings := [r |
+		some r in known_repo_ids
+		not is_string(r)
+	]
+	count(non_strings) > 0
+	reason := "The 'known_rpm_repositories' data is not a list of strings"
+}
+
+# See https://github.com/release-engineering/rhtap-ec-policy/tree/main/data
+# This list of repo_ids is not under the 'rule_data' key, so we don't use lib.rule_data.
```
lcarva, line 83: I see a couple of issues with this approach:

1. This policy rule should not care about what is defined in https://github.com/release-engineering/rhtap-ec-policy/tree/main/data. That repo knows about this one, not the other way around. Consider removing all references to that repo from this PR.
2. We should absolutely use `lib.rule_data` here. Otherwise, the existing overwrite mechanisms won't work, e.g. `ruleData` in the ECP. Arguably, it's unlikely that someone will overwrite the data in that way, but we don't have to restrict it if someone really wants to. It would be surprising to a user if it didn't work that way.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
...
+#   repository id.
+#
+package policy.release.repo_ids
+
+import rego.v1
+
+import data.lib
+
+# METADATA
+# title: Known repo id list provided
```
lcarva, line 15: This can probably be merged with `rpm_repo_ids_valid`.

policy/release/repo_ids.rego
lcarva: +1 to using a schema. You can also define that the list must be non-empty which should simplify the checks here.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
...
+	count(bad_items) > 0
+
+	reason := sprintf(
+		"Some rpm components in the SBOM have purls that could not be parsed: %s",
+		[truncated_list_to_string(bad_items)],
+	)
+}
+
+repo_id_problems contains reason if {
+	bad_items := all_parsed_rpm_purls - all_parsed_purls_with_repo_ids
```
lcarva, line 97: I think we should emit a single violation for each bad item. Having the bad item as a term on the violation. This would allow someone to be very specific in an exclude, for example.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
...
+
+# Extract a named qualifier
+purl_qualifier(key, parsed_purl) := result if {
+	some qualifier in parsed_purl.qualifiers
+	qualifier.key == key
+	result := qualifier.value
+}
+
+# Avoid including thousands of bad purls in the violation reason
+max_bad_purls := 10
```
lcarva, line 159: This seems oddly specific. Let all the failures show up so users can tackle them as needed.

policy/release/repo_ids.rego
simonbaird: Do you mean lib.rule_data should be able to find things in the top level data scope?

policy/release/repo_ids.rego
simonbaird: :thinking:

policy/release/repo_ids.rego
simonbaird: Are you worried about showing (say) 5000 violations?

policy/release/repo_ids.rego
lcarva: Oh! I see your point now. That's a bug on rhtap-ec-policy. That should be under `rule_data`. I'll make a fix.

policy/release/repo_ids.rego
lcarva: Because every purl is malformed? That seems like a bigger problem.

policy/release/repo_ids.rego
lcarva: https://github.com/release-engineering/rhtap-ec-policy/pull/39

policy/release/repo_ids.rego
simonbaird: Done in latest revision.

policy/release/repo_ids.rego
simonbaird: Fixed in latest revision.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
+#
+# METADATA
+# title: Repo IDs
+# description: >-
+#   Checks that all RPMs listed in the SBOM have a valid and known
+#   repository id.
+#
+package policy.release.repo_ids
```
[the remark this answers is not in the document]
simonbaird: Revised in latest version. See if you like it.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
...
+
+import rego.v1
+
+import data.lib
+
+# METADATA
+# title: Known repo id list provided
+# description: >-
+#   A list of valid and known repository ids should be available in the data.
+# custom:
```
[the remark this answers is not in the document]
simonbaird: Still todo, but I think it needs to be 30 days after the syft update that will produce the new style purls rather than 30 days from now.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
...
+#   short_name: rpm_repo_ids_list_provided
+#   failure_msg: 'Valid repo id list not provided: %s'
+#   solution: >-
+#     Include a data source that provides a list of known repository ids, for example
+#     https://github.com/release-engineering/rhtap-ec-policy/blob/main/data/known_rpm_repositories.yml
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	some problem in data_problems
```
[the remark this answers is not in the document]
simonbaird: Might revisit this again - not sure I considered every angle.

policy/release/repo_ids.rego
simonbaird: Fixed.

policy/release/repo_ids.rego
simonbaird: Still open to persuasion, but I think I preferred them separate.

policy/release/repo_ids.rego
simonbaird: I tried it, will demo at standup for consideration.

policy/release/repo_ids.rego
simonbaird: My thinking is: if you do have this problem, it's better if we don't add a 9000 screen EC task log on top of problem pile. :grin:

simonbaird:
Still draft, but nearly there.

policy/release/repo_ids.rego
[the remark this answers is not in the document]
simonbaird: Good now.

simonbaird:
The latest version has a different way of limiting the number of violations, as discussed post standup today.

policy/release/repo_ids.rego
simonbaird: A new approach in latest revision: there is one violation per SBOM component, but the number of violations is capped at a certain limit.

policy/release/repo_ids.rego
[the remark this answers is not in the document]
simonbaird: My plan is to make another PR for that.

policy/release/repo_ids.rego
```diff
@@ -0,0 +1,170 @@
...
+# description: >-
+#   The list of rpm packages in the SBOM will be inspected. A violation will be produced if any
+#   of them do not specify a "repository_id" from the list of known accepted repository ids.
+# custom:
+#   short_name: rpm_repo_ids_valid
+#   failure_msg: 'RPM repo id check failed: %s'
+#   solution: >-
+#     Ensure every rpm is from a known and accepted repository and that the data in the SBOM
+#     correctly records that.
+#   # Todo: Until the sbom generation is upated, this will always fail,
```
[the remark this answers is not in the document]
simonbaird: Will do in a future PR.

simonbaird:
Moved out of draft.

simonbaird:
Things not in this PR:
* Adding the new rule to the 'redhat' collection. I'm planning to do that later while adding an effective_on date in the future.
* Using the 'item' mechanism to be able to create fine grained exclude directives. This is a good idea, but I think it can be done in a new PR and in a new story.

zregvart (approved):
Would have loved to have seen term on a per package basis, but this is okay for now and that can be added in a followup

PR #1142, opened 2024-09-18
Use PURL's subpath when matching disallowed pkgs

lcarva:
There are certain repositories that defined multiple packages. These packages may have different attributes, e.g. different licenses. As such, it is desirable to disallow certain packages from those repos while still allowing others.

A good example of this is the github.com/hashicorp/consul repo. The top-level license is quite restritive, but the one under the `/api` subpath is not.

This commit changes the sbom_cyclonedx.allowed policy rule to take into account the subpath attribute. The subpaths defined in the items of the disallowed_packages list must be an exact match to the subpath of components defined in the SBOM.

To more easily handle the case where all the subpaths must be disallowed, the special `"*"` subpath value can be used.

Ref: EC-870

### DIFF
--- a/policy/release/sbom_cyclonedx.rego
+++ b/policy/release/sbom_cyclonedx.rego
@@ -54,7 +54,9 @@ deny contains result if {
 # title: Allowed
 # description: >-
 #   Confirm the CycloneDX SBOM contains only allowed packages. By default all packages are allowed.
-#   Use the "disallowed_packages" rule data key to provide a list of disallowed packages.
+#   Use the "disallowed_packages" rule data key to provide a list of disallowed packages. The
+#   subpath attribute of the PURL must be an exact match. Use the special value of "*" as a matcher
+#   for all subpaths, e.g. `pkg:golang/my-pakcage#*`.
 # custom:
 #   short_name: allowed
 #   failure_msg: "Package is not allowed: %s"
@@ -191,6 +193,7 @@ _contains(needle, haystack) if {
 	needle_purl.type == hay_purl.type
 	needle_purl.namespace == hay_purl.namespace
 	needle_purl.name == hay_purl.name
+	hay_purl.subpath in {"*", needle_purl.subpath}
 	_matches_version(needle_purl.version, hay)
 } else := false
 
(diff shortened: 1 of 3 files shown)


lcarva:
Corresponding data changes: https://github.com/release-engineering/rhtap-ec-policy/pull/40

simonbaird:
Looking at https://github.com/release-engineering/rhtap-ec-policy/pull/40/files I wonder if we could effectively infer a "#*" if there is no subpath specified, just to make it tidier, and to avoid the need for the wildcard entirely.

That's just an idea, not a blocker. I think this is a quite reasonable solution.

policy/release/sbom_cyclonedx.rego
```diff
@@ -54,7 +54,9 @@ deny contains result if {
 # title: Allowed
 # description: >-
 #   Confirm the CycloneDX SBOM contains only allowed packages. By default all packages are allowed.
-#   Use the "disallowed_packages" rule data key to provide a list of disallowed packages.
+#   Use the "disallowed_packages" rule data key to provide a list of disallowed packages. The
+#   subpath attribute of the PURL must be an exact match. Use the special value of "*" as a matcher
+#   for all subpaths, e.g. `pkg:golang/my-pakcage#*`.
```
simonbaird, line 59:
```suggestion
#   for all subpaths, e.g. `pkg:golang/my-package#*`.
```
I think?

policy/release/sbom_cyclonedx.rego
```diff
@@ -191,6 +193,7 @@ _contains(needle, haystack) if {
 	needle_purl.type == hay_purl.type
 	needle_purl.namespace == hay_purl.namespace
 	needle_purl.name == hay_purl.name
+	hay_purl.subpath in {"*", needle_purl.subpath}
```
simonbaird, line 196: Not sure if it's quite that easy, but:

```suggestion
	hay_purl.subpath in {"", needle_purl.subpath}
```

simonbaird:
```
$ ec opa eval 'ec.purl.parse("pkg:golang/github.com/hashicorp/terraform").subpath' --format values
[
  ""
]
```

lcarva:
Closing this in favor of https://github.com/enterprise-contract/ec-policies/pull/1143

PR #1143, opened 2024-09-18
Allow subpaths to be excluded in disallowed pkgs

lcarva:
There are certain repositories that defined multiple packages. These packages may have different attributes, e.g. different licenses. As such, it is desirable to disallow certain packages from those repos while still allowing others.

A good example of this is the github.com/hashicorp/consul repo. The top-level license is quite restritive, but the one under the `/api` subpath is not.

This commit changes the sbom_cyclonedx.allowed policy rule to take into account the subpath attribute. The objects in the list disallowed_packages array can now provide a list of `exclude_subpaths` in order to ignore certain subpaths.

Ref: EC-870


lcarva:
Usage of the new rule data attribute can only be done when folks are using a version of ec-policies that includes this commit. Otherwise, schema validation of the rule data will fail.

policy/release/sbom_cyclonedx_test.rego
```diff
@@ -198,6 +198,29 @@ test_not_allowed_with_max if {
...
+		"purl": "pkg:golang/github.com/hashicorp/consul",
+		"format": "semverv",
+		"min": "v1.29.2",
+		"exclude_subpaths": [
+			"api",
+			"sdk",
+		],
+	}]
+
+	# Unkonwn subpath matches
```
simonbaird, line 212:
```suggestion
	# Unknown subpath matches
```

policy/release/sbom_cyclonedx_test.rego
```diff
@@ -198,6 +198,29 @@ test_not_allowed_with_max if {
 	assert_allowed("pkg:golang/k8s.io/client-go@v99.99.99", disallowed_packages)
 }
 
+test_not_allowed_with_subpaths if {
+	disallowed_packages := [{
+		"purl": "pkg:golang/github.com/hashicorp/consul",
+		"format": "semverv",
+		"min": "v1.29.2",
+		"exclude_subpaths": [
```
simonbaird, line 206: To consider:

```suggestion
		"subpath_exceptions": [
```

lcarva: Good idea. I think it reads better

lcarva:
@zregvart I like the idea. I'll get the PR updated.

lcarva:
@zregvart updated. Let me know what you think.

policy/release/sbom_cyclonedx.rego
```diff
@@ -192,8 +192,18 @@ _contains(needle, haystack) if {
 	needle_purl.namespace == hay_purl.namespace
 	needle_purl.name == hay_purl.name
 	_matches_version(needle_purl.version, hay)
+
+	not _excluded(needle_purl, object.get(hay, "exceptions", []))
 } else := false
 
+_excluded(purl, exceptions) if {
```
zregvart, line 199: I :heart: this pattern, we can add `_excluded` for any further pattern/key we have, nice

zregvart (approved):

PR #1144, opened 2024-09-18
Mention module purls in rpm repo id check

simonbaird:
Explain in comments that we do intend to match both rpms and modules. Also add some test coverage to confirm and illustrate it.

Ref: https://issues.redhat.com/browse/EC-848

### DIFF
--- a/policy/release/rpm_repos.rego
+++ b/policy/release/rpm_repos.rego
@@ -105,9 +105,15 @@ all_rpm_purls contains purl if {
 	some sbom in _all_sboms
 	some component in sbom.components
 	purl := component.purl
+	_is_rpmish(purl)
+}
 
-	# I'm assuming this is faster than parsing it and checking the type
-	startswith(purl, "pkg:rpm")
+# Match rpms and modules
+# (Use a string match instead of parsing it and checking the type)
+_is_rpmish(purl) if {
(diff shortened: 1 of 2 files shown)


policy/release/rpm_repos.rego
```diff
@@ -106,7 +106,9 @@ all_rpm_purls contains purl if {
 	some component in sbom.components
 	purl := component.purl
 
-	# I'm assuming this is faster than parsing it and checking the type
+	# I'm assuming this is faster than parsing it and checking the type.
+	# It would be more correct to match exactly "pkg:rpm/" and "pkg:rpmmod/"
```
lcarva, line 110: Easy enough to do something like this if you want to aim for correctness:

```rego
_is_rpmish(purl) if {
  startswith(purl, "pkg:rpm/")
} else if {
  startswith(purl, "pkg:rpmmod/")
}
```

simonbaird: Hmm yeah, that's a little tidier that what I had. Let's do it.

lcarva (approved):

policy/release/rpm_repos.rego
simonbaird: I added it, works nicely. No other changes, so I'm taking your approval as still good.

PR #1145, opened 2024-09-19
Exclude non-OCI refs from list of OLM image refs

lcarva:
Ref: EC-874

### DIFF
--- a/policy/lib/image.rego
+++ b/policy/lib/image.rego
@@ -23,6 +23,8 @@ parse(ref) := d if {
 		),
 	)
 
+	not contains(repo, "://")
+
 	d := {
 		"digest": digest,
 		"repo": repo,
--- a/policy/release/olm.rego
+++ b/policy/release/olm.rego
@@ -348,6 +348,7 @@ all_image_ref(manifest) := [e |
 		],
 	]
 	some i in imgs
+
 	e := {"ref": i.ref, "path": i.path}
 ]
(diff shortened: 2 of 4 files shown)


zregvart (commented):
Perhaps this belongs to `image.parse`

policy/release/olm.rego
```diff
@@ -348,6 +348,7 @@ all_image_ref(manifest) := [e |
 		],
 	]
 	some i in imgs
+	not contains(i.ref.repo, "://")
```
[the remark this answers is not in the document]
lcarva: Yeah, definitely. I'll update.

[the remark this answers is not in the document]
lcarva: Done. Let me know what you think.

zregvart (approved):
Nice :+1:

policy/lib/image.rego
```diff
@@ -23,6 +23,8 @@ parse(ref) := d if {
 		),
 	)
 
+	not contains(repo, "://")
```
simonbaird, line 26: I'm assuming `oci://` is not a thing in this context, right? Or `docker://`.

zregvart: I haven't seen that being used outside of skopeo, so I'm not expecting it here

lcarva: Agree.

PR #1152, opened 2024-09-24
Add terms to rpm_repos.ids_known

zregvart:
The term is the whole purl found to be in violation. In addition the main difference is that the rule doesn't emit +1 violation when the list is truncated, a notice is added to the message about remaining similar violations instead.

Reference: EC-877

Also, adds `rpm_repos.rule_data_provided` to the `policy_data` collection.

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -28,6 +28,8 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"slices"
+	"strings"
 	"testing"
 
 	"github.com/cucumber/godog"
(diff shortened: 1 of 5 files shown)


simonbaird (approved):

lcarva:
Does a term work as expected in ec-cli if it contains `:`?

zregvart:
> Does a term work as expected in ec-cli if it contains `:`?

Seems to, I've added an acceptance test in [`f44493e` (#1152)](https://github.com/enterprise-contract/ec-policies/pull/1152/commits/f44493eb0f8273a816c45c52f06c19973b2a15c7), not super happy with it it might be a burden in the future or stop testing the case it wants to test if we refactor. I didn't come up with anything nicer than that...

lcarva (approved):

PR #1154, opened 2024-09-26
Handle missing component `purl` property in SBOM

zregvart:
Seems that we can't rely on all components containing the `purl` property so when the failure message is generated for the a component that does not have the `purl` component the rule execution stops and a violation that should have been reported is not reported.

Empirically it seems that we can use the `bom-ref` property in those cases, and if both `purl` and `bom-ref` properties are not present we can resort to having a dummy `<unknown>` string in the failure message.


kosciCZ:
Looking at the [CycloneDX spec](https://cyclonedx.org/docs/1.6/json/#components_items_name), one of the few required properties of a component is `name`, so that might be better fallback than `<unknown>`

zregvart:
Thanks @kosciCZ, really good feedback. I've changed the rule to use `component.name` with no fallback.

policy/release/sbom_cyclonedx.rego
```diff
@@ -116,8 +116,9 @@ deny contains result if {
 
 	msg := regex.replace(object.get(property, "value", ""), `(.+)`, ` to "$1"`)
 
+	id := object.get(component, "purl", component.name)
```
lcarva, line 119: I see now that `name` and `type` are the only [required properties](https://github.com/enterprise-contract/ec-policies/blob/a9c2e4c0a708e16b00a2e71d4cc3f1d182ede5fb/policy/release/sbom_cyclonedx_schema.rego#L384-L386).

lcarva (approved):

PR #1156, opened 2024-09-26
Make the task expiry message customizable

chmeliik:
Note: this came from an internal Slack conversation about how to distinguish between "task is deprecated because we only maintain the newer version" and "task is discontinued". Somewhat related to https://github.com/konflux-ci/build-definitions/pull/1471#discussion_r1776637740

Along with the expires-on annotation, Task authors can optionally define the expiry-message annotation to override the default expiry message. The default is still "Upgrade to a newer version of the Task."

This required moving the 'solution' part of the rule into the 'failure_msg' instead, because 'solution' does not support templating.

### DIFF
--- a/policy/release/tasks.rego
+++ b/policy/release/tasks.rego
@@ -246,9 +246,7 @@ deny contains result if {
 # custom:
 #   short_name: unsupported
 #   failure_msg: >-
-#     Task %q is used by pipeline task %q is or will be unsupported as of %s.
-#   solution: >-
-#     Upgrade to a newer version of the Task.
+#     Task %q is used by pipeline task %q is or will be unsupported as of %s. %s
 #   collections:
 #   - redhat
 #   depends_on:
@@ -261,11 +259,16 @@ deny contains result if {
 	annotations := tkn.task_annotations(task)
 
 	expires_on := annotations[_expires_on_annotation]
+	expiry_message := object.get(
+		annotations,
+		_expiry_msg_annotation,
+		"Upgrade to a newer version of the Task.",
+	)
 
 	result := object.union(
 		lib.result_helper_with_term(
 			rego.metadata.chain(),
-			[tkn.task_name(task), tkn.pipeline_task_name(task), expires_on],
+			[tkn.task_name(task), tkn.pipeline_task_name(task), expires_on, expiry_message],
 			tkn.task_name(task),
 		),
 		{"effective_on": expires_on},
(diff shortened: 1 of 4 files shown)


policy/release/tasks.rego
```diff
@@ -246,9 +246,7 @@ deny contains result if {
 # custom:
 #   short_name: unsupported
 #   failure_msg: >-
-#     Task %q is used by pipeline task %q is or will be unsupported as of %s.
-#   solution: >-
-#     Upgrade to a newer version of the Task.
```
chmeliik, line 251: Not sure how important it is for the solution to be in `solution` rather than in `failure_msg`. I can look into making the solution template-able, but I would need some help with that

policy/release/tasks_test.rego
```diff
@@ -639,7 +639,27 @@ test_expired_slsa_v1 if {
...
+		with data["pipeline-required-tasks"] as {"generic": []}
+		with data["task-bundles"] as _trusted_tasks
+}
+
+test_expired_with_custom_message if {
+	attestation := _slsav1_attestations_with_tasks({}, [object.union(
+		_task("task"),
+		{"invocation": {"environment": {"annotations": {
+			tasks._expires_on_annotation: "2000-01-01T00:00:00Z",
+			tasks._expiry_msg_annotation: "The Task has been discontinued.",
```
chmeliik, line 656: I noticed the other test case is kinda quadrupled - `[slsav0_2, slsav1] X [already expired, expires in the future]`

It didn't seem necessary to cover the expiry message for all 4 cases, but LMK if that's incorrect

policy/release/tasks.rego
lcarva: I think this is fine.

policy/release/tasks.rego
```diff
@@ -385,3 +388,5 @@ _format_missing(o, opt) := desc if {
 } else := sprintf("Required task %q", [o])
 
 _expires_on_annotation := "build.appstudio.redhat.com/expires-on"
+
+_expiry_msg_annotation := "build.appstudio.redhat.com/expiry-message"
```
lcarva, line 392: It would be great to add some info on this annotation [here](https://github.com/enterprise-contract/ec-policies/edit/main/antora/docs/modules/ROOT/pages/tasks.adoc).

lcarva (approved):

lcarva:
Some linting failures. Add the comment `# regal ignore:line-length` above the lines flagged as too long.

policy/release/tasks.rego
lcarva: And I reviewed it too :facepalm:

chmeliik:
Now it's saying there are uncommitted changes, but I'm not sure what they are

lcarva:
> Now it's saying there are uncommitted changes, but I'm not sure what they are

`make generate-docs`

PR #1162, opened 2024-10-01
SPDX SBOM support

joejstuart:
This change brings the support for SPDX to the same level as CyclondeDX.
- Support for schema checks
- Retrieving SPDX sbom from images, attestations and from the taskRun result url.

It also create a new policy in `sbom.rego` meant for more general sbom related checks, does any sbom exists, etc.

https://issues.redhat.com/browse/EC-800


zregvart (approved):
I did a quick read, LGTM, might want to squash though

joejstuart:
I'm going to mark this as draft until I can get the ec configs in konflux-release-data squared away.

lcarva:
> I'm going to mark this as draft until I can get the ec configs in konflux-release-data squared away.

What do you mean?

joejstuart:
> > I'm going to mark this as draft until I can get the ec configs in konflux-release-data squared away.
> 
> What do you mean?

There's a handful of ec configs in konflux-release-data with an exclude for `sbom_cyclonedx.found`, since I moved that rule to `sbom.found` they'll start throwing violations. I made an MR adding both rules to the exclude list. Then once this change is propagated, I'll remove the `sbom_cyclonedx.found` rule from those configs.

PR #1166, opened 2024-10-01
Limit sbom rpm repo id checks to cachi2 content

simonbaird:
Rather than require all rpm components found in the sbom to have a repo id, we now require only the rpm components that were created by cachi2.

The motivation: We know that Syft does not create the repo id data when it produces sbom components for rpms. Because we can get syft created rpm components in our sboms, they would always fail the check if they were included.

Some day syft might support adding the repo id to the rpm purls it creates, but at this stage it's only cachi2 that adds them.

Ref: https://issues.redhat.com/browse/EC-901


zregvart:
Should there be a warning that non-cachi2 found packages without repository id were found?

simonbaird:
> Should there be a warning that non-cachi2 found packages without repository id were found?

I don't think so because there's nothing that can be done about it.

simonbaird:
> > Should there be a warning that non-cachi2 found packages without repository id were found?
> 
> I don't think so because there's nothing that can be done about it.

..Actually I'm not so sure about this, but either way I expect it could be considered in a future PR.

PR #1167, opened 2024-10-03
Specify correct branch in antora.yml

simonbaird:
We don't have a master branch. Likely a copy/paste snafu made in 93b0e925ed36e7f8fa64249c007d7f08c95910bc.

The impact is minor but let's fix it up anyhow.

Drive-by fix while working on...

Ref: https://issues.redhat.com/browse/EC-903

### DIFF
--- a/antora/docs/antora.yml
+++ b/antora/docs/antora.yml
@@ -17,6 +17,6 @@
 
 name: ec-policies
 title: Enterprise Contract Policies
-version: master
+version: main
 nav:
 - modules/ROOT/nav.adoc


lcarva (approved):

simonbaird:
Hmm, actually I think this does impact some urls. Instead of the first it's now the second.
* https://enterprisecontract.dev/docs/ec-policies/release_policy.html
* https://enterprisecontract.dev/docs/ec-policies/main/release_policy.html

Let me see how/if a redirect can be done so we don't break links.

simonbaird:
Replaced by #1169 probably.

PR #1172, opened 2024-10-04
Specify "unversion" link in ec-cli component

simonbaird:
Trying to avoid this error when building the main website antora docs:

  target of xref not found: main@ec-cli::index.adoc

See https://docs.antora.org/antora/latest/component-with-no-version/#reference-an-unversioned-resource

Ref: https://issues.redhat.com/browse/EC-86


simonbaird:
I think this fixes the antora build, so I'm gonna merge it as soon as it goes green.

lcarva (approved):

PR #1173, opened 2024-10-04
Fixup for last PR: Actually remove the version

simonbaird:
I think this is better since there should be no need to specify a version at all.

Ref: https://issues.redhat.com/browse/EC-86


simonbaird:
I clicked merge just a little too fast on #1172.

simonbaird:
I'll merge at 4:59:59 :grin:

PR #1187, opened 2024-10-10
Bundle CycloneDX schemas

### LINKED ISSUE #1185: Test failure when offline
A test seems to be failing when run without internet connection:

```
policy/release/sbom_cyclonedx_test.rego:
data.policy.release.sbom_cyclonedx_test.test_not_valid: FAIL (5.320726ms)

  Assert equal failure:
    Left value:  {{"code": "sbom_cyclonedx.valid", "msg": "CycloneDX SBOM at index 0 is not valid: components: Invalid type. Expected: array, given: string"}}
    Right value: set()
````

lcarva:
The CycloneDX schema references two other schemas. These are fetched dynamically when the schema is first compiled in rego. This is not ideal because it requires accessing an external server. This commit takes those additional schemas and bundles[1] them making it possible to validate a CycloneDX SBOM completely offline.

[1] https://json-schema.org/understanding-json-schema/structuring#bundling

Fixes #1185


joejstuart:
Do we need to worry about this with `spdx`?

lcarva:
> Do we need to worry about this with `spdx`?

No because there are no refs there. 

We should probably have something to prevent this from happening again in the future. I was looking for a command to create the bundle automatically but couldn't find something ready to use. I just created the bundle by hand. I'll file an issue for that next week.  

Probably low priority as schema changes are rare events, including adding support for new versions.

zregvart (approved):

zregvart:
> We should probably have something to prevent this from happening again in the future.

I think we should run tests without network, see #1188

PR #1188, opened 2024-10-11
Run tests offline on Linux

zregvart:
Uses `unshare` to run the Rego tests on Linux without network access. This makes sure that the tests are not depending on outside service and that all network access is mocked.


zregvart:
Depends on #1187

lcarva (approved):
Nice!

PR #1192, opened 2024-10-14
Make sure `ec` is built with network

zregvart:
In #1188 tests were made to run without network. This works if the local golang module cache already contains the required dependencies, i.e can be built offline. When a dependency is not available from the cache, due to the offline nature of the namespace the build fails.

This circumvents that by invoking the equivalent of `ec opa version` so that the `ec` command is build outside the offlined namespace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,10 +24,13 @@ OPA=$(EC) opa
 CONFTEST=EC_EXPERIMENTAL=1 $(EC)
 TKN=go run github.com/tektoncd/cli/cmd/tkn
 TEST_CMD_DEFAULT=$(OPA) test $(TEST_FILES) $(TEST_FILTER)
+# if unshare is available we isolate the process to run without network access,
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -27,7 +27,7 @@ TEST_CMD_DEFAULT=$(OPA) test $(TEST_FILES) $(TEST_FILTER)
 ifeq ($(shell command -v unshare),)
   TEST_CMD=$(TEST_CMD_DEFAULT)
 else
-  TEST_CMD=unshare -r -n $(TEST_CMD_DEFAULT)
+  TEST_CMD=$(OPA) version > /dev/null && unshare -r -n $(TEST_CMD_DEFAULT)
```
lcarva, line 30: It would be great to add a comment to help us remember in the future.

lcarva (approved):

PR #1198, opened 2024-10-17
Flatten package names

lcarva:
This commit changes the package name of all the policy rules to a single level, e.g. `release.cve` becomes `cve`. This better aligns with how these packages are used today. It moves the EC ecosystem closer to using package names as they are defined without having to rely on dynamically alterations to the package name.

All package shortening code is removed from this repository. The package name used is now an exact match to the value provided to the `package` keyword in the rego files. (Note that package shortening still exists in the EC CLI, but that is expected to be removed soon as well.)

The docs generation code is modified to take into account the file path of the policies in addition to its annotations. This is needed to distinguish between the policies in different directories.

The "Package full name" section from the docs is removed since it is now exactly the same as the value of "Package name".

Although policy rules under different directories are never used together, the tests in this repo load them all at once. As such, the package names have to be unique across the different directories. The one case where this caused a conflict was with
`policy/build_task/labels` and `policy/release/labels`. To mitigate this, `policy/build_task/labels` is renamed to
`policy/build_task/build_labels`. This is a breaking change but given that this policy is not widely used, this is ok.

Ref: EC-864

### DIFF
--- a/policy/build_task/build_labels/build_labels.rego
+++ b/policy/build_task/build_labels/build_labels.rego
@@ -5,7 +5,7 @@
 #   Policies to verify that a Tekton build task definition has the
 #   required build type label.
 #
-package build_task.labels
+package build_labels
 
 import rego.v1
 
--- a/antora/docs/modules/ROOT/pages/build_task_policy.adoc
+++ b/antora/docs/modules/ROOT/pages/build_task_policy.adoc
@@ -4,30 +4,29 @@
 
 These rules are applied to Tekton build task definitions.
 
-[#labels_package]
-== link:#labels_package[Tekton task build type label checks]
+[#build_labels_package]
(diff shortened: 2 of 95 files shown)


policy/build_task/build_labels/build_labels.rego
```diff
@@ -5,7 +5,7 @@
 #   Policies to verify that a Tekton build task definition has the
 #   required build type label.
 #
-package build_task.labels
+package build_labels
```
lcarva, line 8: As mentioned in the description, this is a breaking change. That should be ok as, likely, the only place that uses it is [here](https://github.com/konflux-ci/build-definitions/blob/main/policies/build-tasks.yaml), and that references the bundle, not the package itself.

zregvart (approved):
:+1:

Issue #1157, opened 2024-09-26
Downgrade `additionalProperties` violations to warnings


lcarva:
When validating rule data, we often set `additionalProperties` to `false`, see [example](https://github.com/enterprise-contract/ec-policies/blob/1c21485ac210058932cea9e8bc66588c9d4a9340/policy/release/sbom_cyclonedx.rego#L257). The value of doing this is that it allows us to catch typos on non-required properties.

However, when we want to add a new property, we must carefully synchronize changes. After all, older versions of ec-policies will  emit a violation for such property.

Let's introduce a mechanism to bring the best of both sides. Instead of emitting a violation when an unknown property is used, let's emit a warning.

Because this pattern will have to me used in multiple places within this repo, it's a good idea to create some sort of helper rule for this.

To avoid duplicating policy rules (one for schema violations and one for schema warnings), let's leverage https://github.com/enterprise-contract/ec-cli/issues/2025 for this work.

**Acceptance Criteria**

* When unknown properties are used for rule data items, a warning is emitted instead of a violation.

PR #1204, opened 2024-10-23
Add lib.json package with JSON schema validation

zregvart:
Consolidates the repeated logic of JSON schema validation in the `json.validate_schema` function. Namely, support for arrays -- by marshaling to a string first.

Also categorizes the schema violations by severity (`failure` vs `warning`).

This is taken advantage of when unexpected additional properties are encountered in the provided document, which are categorized as a `warning` instead of `failure`.

Resolves: #1157

Second commit makes the use of this new helper in the existing rules.


lcarva (approved):
Nice!

Thanks for taking care of this :pray:

PR #1212, opened 2024-11-05
Add allowed package sources policy

kosciCZ:
This adds a new `allowed_package_sources` policy that can gate content fetched by cachi2. This rule take a lot from the `allowed_external_references`, but makes it specific to content fetched by cachi2, because it is not uncommon for dependencies fetched by other means to define externalReferences of type `distribution`. Cachi2 currently produces the mentioned external reference only for packages fetched with the generic fetcher ([docs WIP](https://github.com/containerbuildsystem/cachi2/pull/718)) that always produces a `pkg:generic` purl, but it is possible it will be extended to other package managers supported by cachi2 as well. 

The rule data for this policy are structured as a list of regex patterns for a given purl type. Examples in tests.

Finally, this is my first contribution to EC (and rego) so please excuse any antipatterns I might have invented, feedback very much welcomed. 

Resolves ISV-5342

### DIFF
--- a/policy/lib/sbom/sbom.rego
+++ b/policy/lib/sbom/sbom.rego
@@ -255,6 +255,38 @@ rule_data_errors contains error if {
 	}
 }
 
+# Verify allowed_package_sources is array of purl/regex list pairs
+rule_data_errors contains error if {
+	some e in j.validate_schema(
+		lib.rule_data(rule_data_allowed_package_sources_key),
+		{
+			"$schema": "http://json-schema.org/draft-07/schema#",
+			"type": "array",
+			"items": {
+				"type": "object",
+				"properties": {
+					"type": {"type": "string"},
+					"patterns": {
+						"type": "array",
+						"items": {
+							"type": "string",
+							"format": "regex",
+						},
+					},
+				},
+				"required": ["type", "patterns"],
+				"additionalProperties": false,
+			},
+		},
+	)
+
+	error := {
+		# regal ignore:line-length
+		"message": sprintf("Rule data %s has unexpected format: %s", [rule_data_allowed_package_sources_key, e.message]),
+		"severity": e.severity,
+	}
+}
+
 _sbom_cyclonedx_image_path := "root/buildinfo/content_manifests/sbom-cyclonedx.json"
 
 _sbom_spdx_image_path := "root/buildinfo/content_manifests/sbom-spdx.json"
@@ -266,3 +298,5 @@ rule_data_attributes_key := "disallowed_attributes"
 rule_data_allowed_external_references_key := "allowed_external_references"
 
 rule_data_disallowed_external_references_key := "disallowed_external_references"
+
+rule_data_allowed_package_sources_key := "allowed_package_sources"
--- a/policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
+++ b/policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
@@ -148,6 +148,61 @@ deny contains result if {
 	result := lib.result_helper_with_term(rego.metadata.chain(), [id, reference.url, reference.type, msg], id)
 }
 
+# METADATA
+# title: Allowed package sources
+# description: >-
+#   For each of the components fetched by Cachi2 which define externalReferences of type
+#   distribution, verify they are allowed based on the allowed_package_sources rule data
+#   key. By default, allowed_package_sources is empty, which means no components with such
+#   references are allowed.
+# custom:
+#   short_name: allowed_package_sources
+#   failure_msg: Package %s fetched by cachi2 was sourced from %q which is not allowed
+#   solution: Update the image to not use a package from a disallowed source.
+#   collections:
+#   - redhat
+#   - policy_data
+#   effective_on: 2024-12-15T00:00:00Z
+deny contains result if {
+	some s in sbom.cyclonedx_sboms
+	some component in s.components
+
+	# only look at components that define an externalReferences of type `distribution`
+	some reference in component.externalReferences
+	reference.type == "distribution"
+
+	# only look at components fetched by cachi2
+	some properties in component.properties
+	properties.name == "cachi2:found_by"
+	properties.value == "cachi2"
+
+	purl := component.purl
+	parsed_purl := ec.purl.parse(purl)
+
+	# patterns are either those defined by the rule for a given purl type, or empty by default
+	allowed_data := lib.rule_data(sbom.rule_data_allowed_package_sources_key)
+	patterns := _get_purl_allowed_patterns(parsed_purl.type, allowed_data)
+	distribution_url := object.get(reference, "url", "")
+
+	# only progress past this point if no matches were found
+	not _url_matches_any_pattern(distribution_url, patterns)
+
+	result := lib.result_helper_with_term(rego.metadata.chain(), [purl, distribution_url], purl)
+}
+
+# get allowed pattens for given purl type, or empty list if not defined
+_get_purl_allowed_patterns(purl_type, allowed_rule_data) := patterns if {
+	some allowed in allowed_rule_data
+	purl_type == allowed.type
+	patterns := allowed.patterns
+} else := []
+
+# see if any pattern matches given url
+_url_matches_any_pattern(url, patterns) if {
+	some pattern in patterns
+	regex.match(pattern, url)
+}
+
 # _with_effective_on annotates the result with the item's effective_on attribute. If the item does
 # not have the attribute, result is returned unmodified.
 _with_effective_on(result, item) := new_result if {
--- a/antora/docs/modules/ROOT/pages/release_policy.adoc
+++ b/antora/docs/modules/ROOT/pages/release_policy.adoc
@@ -75,6 +75,7 @@ Rules included:
 * xref:release_policy.adoc#rpm_repos__rule_data_provided[RPM Repos: Known repo id list provided]
 * xref:release_policy.adoc#rpm_signature__rule_data_provided[RPM Signature: Rule data provided]
 * xref:release_policy.adoc#sbom_cyclonedx__allowed_package_external_references[SBOM CycloneDX: Allowed package external references]
+* xref:release_policy.adoc#sbom_cyclonedx__allowed_package_sources[SBOM CycloneDX: Allowed package sources]
 * xref:release_policy.adoc#sbom_cyclonedx__disallowed_package_attributes[SBOM CycloneDX: Disallowed package attributes]
 * xref:release_policy.adoc#sbom_cyclonedx__disallowed_package_external_references[SBOM CycloneDX: Disallowed package external references]
 * xref:release_policy.adoc#sbom__disallowed_packages_provided[SBOM: Disallowed packages list is provided]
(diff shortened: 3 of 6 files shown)


kosciCZ:
@ralphbean could I get your eyes on this as general concept as well as the `effective_on` date?

policy/lib/sbom/sbom.rego
```diff
@@ -266,3 +266,5 @@ rule_data_attributes_key := "disallowed_attributes"
 rule_data_allowed_external_references_key := "allowed_external_references"
 
 rule_data_disallowed_external_references_key := "disallowed_external_references"
+
+rule_data_allowed_package_sources_key := "allowed_package_sources"
```
lcarva, line 270: See the `rule_data_errors` rules above. Let's add a new one that verifies the consistency of this new rule data key. This helps us determine if the rule data provided has any problems.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,63 @@ deny contains result if {
 	result := lib.result_helper_with_term(rego.metadata.chain(), [id, reference.url, reference.type, msg], id)
 }
 
+# METADATA
+# title: Allowed package sources
+# description: >-
+#   Confirm the CycloneDX SBOM packages fetched by Cachi2 that define externalReferences
```
lcarva, line 154: This is a bit hard to read. Let's try to make it nicer since it's user-facing, consider:

> For each of the components fetched by Cachi2 which define `externalReferences` of type `distribution`, verify they are allowed based on the `allowed_package_sources` rule data key. By default, `allowed_package_sources` is empty which means no components with such references are allowed.

(We should call them components, not packages, as per the CycloneDX spec.)

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,63 @@ deny contains result if {
...
+	some reference in component.externalReferences
+	some properties in component.properties
+
+	parsed_purl := ec.purl.parse(component.purl)
+
+	# only look at components fetched by cachi2 that define an externalReferences
+	# of type `distribution`
+	properties.name == "cachi2:found_by"
+	properties.value == "cachi2"
+	reference.type == "distribution"
```
lcarva, line 178: This should probably go right after `some reference in ...`. There's no need to iterate over properties if the reference type has an unexpected value.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,63 @@ deny contains result if {
...
+#   - redhat
+#   - policy_data
+#   effective_on: 2024-11-30T00:00:00Z
+deny contains result if {
+	some s in sbom.cyclonedx_sboms
+	some component in s.components
+	some reference in component.externalReferences
+	some properties in component.properties
+
+	parsed_purl := ec.purl.parse(component.purl)
```
lcarva, line 172: Let's do this after the filtering below to avoid having to parse PURLs unnecessarily.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,63 @@ deny contains result if {
...
+	properties.value == "cachi2"
+	reference.type == "distribution"
+
+	# patterns are either those defined by the rule for a give purl type, or empty
+	# by default
+	allowed_data := lib.rule_data(sbom.rule_data_allowed_package_sources_key)
+	patterns := _get_purl_allowed_patterns(parsed_purl.type, allowed_data)
+
+	distribution_url := object.get(reference, "url", "")
+	matches := _url_matches_any_pattern(distribution_url, patterns)
```
lcarva, line 186: Let's remove the `else` from `_url_matches_any_pattern` and change this line to:

```rego
not _url_matches_any_pattern(distribution_url, patterns)
```

which means we can also remove `matches == false`

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,63 @@ deny contains result if {
...
+
+	# only progress past this point if no matches were found
+	matches == false
+
+	id := component.purl
```
lcarva, line 191: We learned the other day that `purl` is actually an optional attribute. I think it's ok to do this here because you have already accessed earlier on in the rule, and presumably if there isn't a purl defined, this rule doesn't make sense for that component.

To make things safer(?) for the future, consider `purl := component.purl` earlier in rule and just use `purl` throughout.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,63 @@ deny contains result if {
...
+#   given purl type. By default no sources are allowed. Use the "allowed_package_sources"
+#   rule data key to provide a list of patterns for each purl type.
+# custom:
+#   short_name: allowed_package_sources
+#   failure_msg: Package %s fetched by cachi2 was sourced from %q which is not allowed
+#   solution: Update the image to not use a package from a disallowed source.
+#   collections:
+#   - redhat
+#   - policy_data
+#   effective_on: 2024-11-30T00:00:00Z
```
lcarva, line 165: Let's set this to a month from now.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,63 @@ deny contains result if {
...
+deny contains result if {
+	some s in sbom.cyclonedx_sboms
+	some component in s.components
+	some reference in component.externalReferences
+	some properties in component.properties
+
+	parsed_purl := ec.purl.parse(component.purl)
+
+	# only look at components fetched by cachi2 that define an externalReferences
```
lcarva, line 174: I do wonder if this rule can be a bit more data driven. For example, what if we allowed the rule data to be something like this:

```yaml
- type: maven
  properties:
    - name: cachi2:found_by
      value: cachi2
  reference:
    type: distribution
  allowed_references:
    - ".*apache.org.*"
    - ".*example.com.*"
```

Or some variation of that. WDYT?

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,63 @@ deny contains result if {
 	result := lib.result_helper_with_term(rego.metadata.chain(), [id, reference.url, reference.type, msg], id)
 }
 
+# METADATA
```
lcarva, line 151: We also need corresponding changes in `sbom_spdx.rego` :sob:

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: (Also I think we should settle this first before I get to some of the other comments)

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: Fixed.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: Agree, this rule won't work unless the purl is present. Implemented the suggestion in the second part of the comment.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: Nice, fixed!

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: Fixed.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: Fixed.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: Fixed.

policy/lib/sbom/sbom.rego
kosciCZ: Holding out on this until the the comment about rule data format is settled.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: Cachi2 currently does not produce an SPDX SBOM, so there's not yet anything to latch onto (as in to the externalReference of type distribution in CycloneDX). Holding out on this until the the comment about rule data format is settled.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,61 @@ deny contains result if {
...
+
+	# only look at components fetched by cachi2
+	some properties in component.properties
+	properties.name == "cachi2:found_by"
+	properties.value == "cachi2"
+
+	purl := component.purl
+	parsed_purl := ec.purl.parse(purl)
+
+	# patterns are either those defined by the rule for a give purl type, or empty by default
```
simonbaird, line 182:
```suggestion
	# patterns are either those defined by the rule for a given purl type, or empty by default
```

simonbaird:
It would be nice to add some explanation to the commit message, and also include a Jira reference, e.g.:

```
Ref: https://issues.redhat.com/browse/ISV-5342
```

(I know it's explained in the PR, but having it in the git history is preferable IMO.)

Nice PR btw!

policy/lib/sbom/sbom.rego
simonbaird: Since the rule is specific for cachi2, should it maybe be "cachi2_allowed_sources"?

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -148,6 +148,61 @@ deny contains result if {
 	result := lib.result_helper_with_term(rego.metadata.chain(), [id, reference.url, reference.type, msg], id)
 }
 
+# METADATA
+# title: Allowed package sources
+# description: >-
+#   For each of the components fetched by Cachi2 which define externalReferences of type
+#   distribution, verify they are allowed based on the allowed_package_sources rule data
+#   key. By default, allowed_package_sources is empty which means no components with such
```
simonbaird, line 156:
```suggestion
#   key. By default, allowed_package_sources is empty, which means no components with such
```

:shrug:

policy/lib/sbom/sbom.rego
kosciCZ: Well, a change in the name of cachi2 is imminent, so I'd like to avoid making this too specific. ([ref](https://github.com/containerbuildsystem/cachi2/discussions/711?sort=old))

What do you think is the right course of action?

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: Applied

kosciCZ:
> It would be nice to add some explanation to the commit message, and also include a Jira reference, e.g.:

Commit message amended to better describe the change and reference a JIRA ticket.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: @lcarva @simonbaird any thoughts on this?

lcarva: Let's keep it as is. We can always evolve it later on if the need becomes obvious.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
lcarva: Ack. We can do this in a follow up PR. No need to block this change. I just want to make sure it doesn't get forgotten. Maybe a follow up JIRA issue?

policy/lib/sbom/sbom.rego
kosciCZ: Rule validation + tests added.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
kosciCZ: I created [EC-1022](https://issues.redhat.com/browse/EC-1022) in your backlog, hope that's okay.

lcarva (approved):

PR #1215, opened 2024-11-06
RHTAP multi-ci initial MVP

simonbaird:
Draft WIP. This is not nice, since it's basically copy/pasting a bunch of code, not once but twice.

I'm sure there's a better way to do it - please share your ideas!

Ref: https://issues.redhat.com/browse/EC-990
Ref: https://issues.redhat.com/browse/EC-991


simonbaird:
To unblock RHTAP testing, I'm thinking to merge this as-is and improve it later.

To that end I removed the invocation id check which is jenkins specific from the two new packages.

It leaves on a "attestation of expected build type found" check, but I think that is a reasonable MVP, and it will allow the RHTAP gitops pipeline to pass.

PR #1220, opened 2024-11-12
Check accessibility of Step images

zregvart:
Resolves: https://issues.redhat.com/browse/EC-912

### DIFF
--- a/antora/docs/modules/ROOT/pages/task_policy.adoc
+++ b/antora/docs/modules/ROOT/pages/task_policy.adoc
@@ -4,8 +4,27 @@
 
 These rules are applied to Tekton task definitions.
 
+[#step_images_package]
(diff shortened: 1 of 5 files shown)


lcarva (approved):
Let's be sure to add this new policy package [here](https://github.com/konflux-ci/build-definitions/blob/7f1d27464163522ed182e30e9bdd53bd65b101a7/policies/all-tasks.yaml#L26).

zregvart:
> Let's be sure to add this new policy package [here](https://github.com/konflux-ci/build-definitions/blob/7f1d27464163522ed182e30e9bdd53bd65b101a7/policies/all-tasks.yaml#L26).

https://github.com/konflux-ci/build-definitions/pull/1625

PR #1225, opened 2024-11-19
Set dupe entries as warnings in trusted tasks

lcarva:
This commit demotes duplicated entries in the trusted tasks from a violation to warning. The current understanding is that preventing duplicated items is nice to have but does not affect functionality. As such, these should not trigger an EC violation.


joejstuart (approved):

simonbaird (approved):

zregvart (commented):
:+1: thanks!

PR #1226, opened 2024-11-20
Remove the unique items constraint

zregvart:
Turns out we like having duplicate entries, see HACBS-2394[1]

Reference: https://issues.redhat.com/browse/EC-1015

[1] https://issues.redhat.com/browse/HACBS-2394

### DIFF
--- a/policy/lib/tekton/trusted.rego
+++ b/policy/lib/tekton/trusted.rego
@@ -90,7 +90,6 @@ data_errors contains error if {
 					"required": ["effective_on", "ref"],
 					"additionalProperties": false,
 				},
-				"uniqueItems": true,
 				"minItems": 1,
 			}},
 		},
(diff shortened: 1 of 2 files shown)


policy/lib/tekton/trusted.rego
```diff
@@ -90,7 +90,6 @@ data_errors contains error if {
 					"required": ["effective_on", "ref"],
 					"additionalProperties": false,
 				},
-				"uniqueItems": true,
```
lcarva, line 93: Remove the `with_severity_for_pattern` call from below? Or maybe also revert this as part of this PR: https://github.com/enterprise-contract/ec-policies/pull/1225

lcarva: We should do this as a cleanup, but let's go ahead and merge it as is.

lcarva (approved):

PR #1227, opened 2024-11-20
Remove severity reduction to warning

zregvart:
We're no longer requiring unique items so this makes the call to
`with_severity_for_pattern` dead code.

Reference: https://issues.redhat.com/browse/EC-1015

### DIFF
--- a/policy/lib/json/schema.rego
+++ b/policy/lib/json/schema.rego
@@ -59,8 +59,3 @@ _prepare_document(doc) := d if {
 _severity(e) := "warning" if {
 	startswith(e.desc, "Additional property")
 } else := "failure"
-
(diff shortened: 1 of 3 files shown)


policy/lib/tekton/trusted_test.rego
```diff
@@ -86,11 +86,6 @@ test_data_errors if {
 			{"ref": "bad-effective-on", "effective_on": "not-a-date"},
 			{"ref": "bad-effective-on", "effective_on": "2024-01-01T00:00:00Z", "expires_on": "not-a-date"},
 		],
-		# this is allowed
-		"duplicated-entries": [
```
lcarva, line 90: I think we should keep this to make sure we don't accidentally break this use case in the future.

lcarva (approved):

PR #1228, opened 2024-11-21
Don't warn of newer tasks if in grace period

zregvart:
Adds new rule data `task_expiry_warning_days` (default 0) if given and greater than 0 then any Task that will expire in `task_expiry_warning_days` days will emit a warning.

Also adds effective_on of the newest Task to the message of the policy rules asserting that the Tasks used are up to date.

Reference: https://issues.redhat.com/browse/EC-944

### DIFF
--- a/policy/lib/tekton/trusted.rego
+++ b/policy/lib/tekton/trusted.rego
@@ -21,17 +21,24 @@ missing_trusted_tasks_data if {
 	count(_trusted_tasks) == 0
 }
 
-# Returns an object containing the task and effective_on of a newer task for a
-# set of Tasks, value is returned for a task only if newer Task, matched by Task
-# reference, exist.
-newer_tasks_of(tasks) := {t |
-	some task in tasks
-	is_trusted_task(task)
-	newest := _newest_record_for(task)
-	t := {
-		"task": task,
-		"newer_effective_on": newest.effective_on,
-	}
+default task_expiry_warnings_after := 0
+
+task_expiry_warnings_after := grace if {
+	grace_period_days := lib_rule_data("task_expiry_warning_days")
+	grace_period_days > 0
+	grace := time.add_date(
+		time_lib.effective_current_time_ns, 0, 0,
+		grace_period_days,
+	)
+}
+
+# Returns the epoch time in nanoseconds of the time when the Task expires, or
+# nothing if Task is not set to expire currently.
+expiry_of(task) := expires if {
+	expires := _task_expires_on(task)
+
+	# only report if the task is expiring within task_expiry_warning_days days
+	expires > task_expiry_warnings_after
 }
 
 # Returns a subset of tasks that do not use a trusted Task reference.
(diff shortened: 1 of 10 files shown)


zregvart:
I'm wondering if we want the grace period, there is a scenario when a newer task is available and within a grace period long enough or often enough that the Task that is used is no longer trusted. So by that point no warning would be seen until a blocking violation. Perhaps I misunderstood what EC-944 want's to achieve.

lcarva:
> I'm wondering if we want the grace period, there is a scenario when a newer task is available and within a grace period long enough or often enough that the Task that is used is no longer trusted. So by that point no warning would be seen until a blocking violation. Perhaps I misunderstood what [EC-944](https://issues.redhat.com/browse/EC-944) want's to achieve.

Yeah, it's a risk. I think we just have to make sure there is enough time between the grace period and the time it becomes a violation. We could make sure that that period is still 30 days. WDYT?

zregvart:
I'm thinking 8d4943c351fcc0f0655d40c318c8d1f05336c934 should stay and 58da16facb2fd59b1f4c7f3940db272ddafa821e should go

lcarva (approved):

lcarva:
> I'm thinking [8d4943c](https://github.com/enterprise-contract/ec-policies/commit/8d4943c351fcc0f0655d40c318c8d1f05336c934) should stay and [58da16f](https://github.com/enterprise-contract/ec-policies/commit/58da16facb2fd59b1f4c7f3940db272ddafa821e) should go

If we have concerns with EC-944 we should probably discuss in the Konflux Architecture call. The goal of that story was to intentionally hide some warnings on the basis that they are noisy and hide more important warnings.

zregvart:
I'm reworking this based on the `expires_on` from the trusted task record

simonbaird (approved):

zregvart:
Updated description to reflect the latest implementation

policy/lib/tekton/trusted.rego
```diff
@@ -21,11 +21,24 @@ missing_trusted_tasks_data if {
...
 }
 
-# Returns a subset of tasks that use a trusted Task reference, but an updated Task reference exists.
-out_of_date_task_refs(tasks) := {task |
-	some task in tasks
-	is_trusted_task(task)
-	_newer_record_exists(task)
+default task_expiry_grace := 0
+
+task_expiry_grace := grace if {
```
simonbaird, line 26: Might be bike-shedding a little, but how about a different name:

```suggestion
default task_expiry_warnings_after := 0
```

policy/lib/tekton/trusted.rego
```diff
@@ -21,11 +21,24 @@ missing_trusted_tasks_data if {
...
+	grace_days := lib_rule_data("task_expiry_warning_days")
+	grace_days > 0
+	grace := time.add_date(
+		time_lib.effective_current_time_ns, 0, 0,
+		lib_rule_data("task_expiry_warning_days"),
```
simonbaird, line 31:
```suggestion
		grace_days,
```

policy/lib/tekton/trusted.rego
```diff
@@ -21,11 +21,24 @@ missing_trusted_tasks_data if {
...
 
-# Returns a subset of tasks that use a trusted Task reference, but an updated Task reference exists.
-out_of_date_task_refs(tasks) := {task |
-	some task in tasks
-	is_trusted_task(task)
-	_newer_record_exists(task)
+default task_expiry_grace := 0
+
+task_expiry_grace := grace if {
+	grace_days := lib_rule_data("task_expiry_warning_days")
```
simonbaird, line 27: More nitpicks, but I think "grace period" is  little more understandable.

```suggestion
	grace_period_days := lib_rule_data("task_expiry_warning_days")
```

policy/release/trusted_task/trusted_task.rego
```diff
@@ -52,18 +52,21 @@ warn contains result if {
 #   Check if all Tekton Tasks use the latest known Task reference.
 # custom:
 #   short_name: current
-#   failure_msg: Pipeline task %q uses an out of date task reference, %s
+#   failure_msg: >-
+#     Pipeline task %q uses an out of date task reference, %s, new version of the
```
simonbaird, line 56:
```suggestion
#     Pipeline task %q uses an out of date task reference, %s. A newer version of the
```

policy/release/trusted_task/trusted_task.rego
```diff
@@ -52,18 +52,21 @@ warn contains result if {
...
+#     Task must be used before %s
```
simonbaird, line 57: :shrug: 

```suggestion
#     task must be used before %s
```

simonbaird (approved):

zregvart:
Suggestions applied, merging this now

PR #1232, opened 2024-11-22
Add StepAction policy rules

lcarva:
Ref: EC-1010


antora/docs/modules/ROOT/pages/stepaction.adoc
```diff
@@ -0,0 +1 @@
+= StepAction Policy
```
zregvart, line 1: Should this be in here?

zregvart (approved):

PR #1237, opened 2024-11-27
Fix step_images.step_images_accessible check

lcarva:
Although the implementation of ec.oci.image_manifest does return nil if the Image Manifest is not found, rego converts that to no value at all. Thus, asserting that the returned value is nil will never be true causing this policy rule to always pass.

This commit changes the rule evaluation to check for the lack of a value instead.


lcarva:
This fix is correct, but if we merge it, we will break the CI of the build-definitions repo. We need to figure out how to proceed there.

zregvart (approved):

lcarva:
The issue on build-definitions was resolved by making sure the registry auth was provided to EC. Moving it out of draft.

Added an effective_on to the policy rule for `2025-01-10`. 30 days plus enough time past holidays. I'll keep an eye in the logs for build-definitions CI since a warning won't show up in the PRs.

lcarva:
Good call on setting the effective_on annotation. Found two issues: EC-1038 and EC-1039.

PR #1257, opened 2025-01-10
Allow list of RPM repo IDs to be extended

lcarva:
Ref: EC-1059

### DIFF
--- a/policy/release/rpm_repos/rpm_repos.rego
+++ b/policy/release/rpm_repos/rpm_repos.rego
@@ -21,7 +21,9 @@ import data.lib.json as j
 #   failure_msg: "Rule data '%s' has unexpected format: %s"
 #   solution: >-
 #     Include a data source that provides a list of known repository ids under the
-#     'known_rpm_repositories' key under the top level 'rule_data' key.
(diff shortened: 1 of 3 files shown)


policy/release/rpm_repos/rpm_repos.rego
```diff
@@ -181,10 +185,18 @@ _is_rpmish(purl) if {
...
 }
 
-_known_repo_ids := lib.rule_data(_rule_data_key)
+_known_repo_ids := combined if {
+	extra := lib.rule_data(_rule_data_extras_key)
+	known := lib.rule_data(_rule_data_key)
+	combined := array.concat(extra, known)
+} else := known if {
+	known := lib.rule_data(_rule_data_key)
+}
```
simonbaird, line 194: IIRC `lib.rule_data` returns `[]` when the key doesn't exist, so you could skip the else clause here if you wanted to.

simonbaird (approved):

policy/release/rpm_repos/rpm_repos.rego
lcarva: Ah! But if either is not an array (for whatever reason), then `_known_repo_ids` will produce no result having unintended consequences.

PR #1267, opened 2025-01-16
Support base images stored in SPDX format.

joejstuart:
EC adds support for SPDX format along side CycloneDX.

https://issues.redhat.com/browse/EC-882

### DIFF
--- a/policy/release/base_image_registries/base_image_registries.rego
+++ b/policy/release/base_image_registries/base_image_registries.rego
@@ -108,19 +108,36 @@ _image_ref_permitted(image_ref) if {
 	image.parse(image_ref).digest in allowed_digests
 }
 
-_base_images contains base_image if {
+_cyclonedx_base_images := [component.name |
 	some s in sbom.cyclonedx_sboms
 	some formulation in s.formulation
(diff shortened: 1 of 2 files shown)


policy/release/base_image_registries/base_image_registries.rego
```diff
@@ -125,6 +133,15 @@ _is_base_image(component) if {
 	count(base_image_properties) > 0
 }
 
+# spdx format
+_is_base_image(component) if {
```
simonbaird, line 137: Just an idea: There's maybe not much value in having the same rule name as line 128 here. Consider naming it `_is_spdx_base_image` and `_is_cyclonedx_base_image`, then you could remove the two comments.

I guess the same applies for the two `_is_base_image_property` rules.

simonbaird (approved):

policy/release/base_image_registries/base_image_registries.rego
joejstuart: That's a good idea. I'm updating `is_base_image`, the property function is separate so I'll leave that one for now.

PR #1268, opened 2025-01-17
Handle more pipeline_intention values

ralphbean:
The goal here is to make the schedule rules only apply to production releases.

The konflux-ci release pipelines don't yet provide these new pipeline_intention values. I'll follow up in a separate PR with them to provide those.


simonbaird:
If you do `make fmt` and git add the chanes it will get that formatting CI check passing.

joejstuart (approved):

PR #1271, opened 2025-01-22
Add a check for the RPM build pipleline

jorris:
Ensure that all tasks building the RPM come from an approved pipeline.


jorris:
Forgot to sign

PR #1272, opened 2025-01-22
Add a check for the RPM build pipleline

jorris:
Ensure that all tasks building the RPM come from an approved pipeline.

### DIFF
--- a/policy/release/rpm_pipeline/rpm_pipeline.rego
+++ b/policy/release/rpm_pipeline/rpm_pipeline.rego
@@ -1,0 +1,45 @@
+#
+# METADATA
+# title: RPM Pipeline
+# description: >-
+#   This package provides rules for verifying the RPMs are built in an approved pipeline
+#
+package rpm_pipeline
+
+import rego.v1
+
+import data.lib
+import data.lib.tekton
+
+_pipeline_key := "build.appstudio.redhat.com/pipeline"
+
+_rule_data_key := "allowed_rpm_build_pipelines"
+
+# METADATA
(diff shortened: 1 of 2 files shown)


simonbaird:
`make fmt` should fix the formatting CI fail.

policy/release/rpm_pipeline/rpm_pipeline.rego
```diff
@@ -0,0 +1,45 @@
...
+# description: >-
+#   The Tekton Task used specifies an invalid pipeline. The Task is annotated with
+#   `build.appstudio.redhat.com/pipeline` annotation, which must be in the set of
+#   `allowed_rpm_build_pipelines` in the rule data.
+# custom:
+#   short_name: invalid_pipeline
+#   failure_msg: >-
+#     Task %q uses invalid pipleline %s, which is not in the list of valid pipelines: %s
+#   collections:
+#   - redhat
```
simonbaird, line 29: We might need to use a different collection. Adding "redhat" here means this rule is in the redhat collection, and hence it would be required to pass for every container image. I think we want this rule to apply only for rpm artifacts, so either we need some more rego to determine if the att being examined is for an rpm build, or we use a different collection specifically for rpm Konflux builds, perhaps "redhat-rpms" or similar.

jorris:
> `make fmt` should fix the formatting CI fail.

Yeah, I was running it but forgot for the last round.

jorris:
Sigh. Need to set global sign if I'm going to keep forgetting.

PR #1283, opened 2025-02-04
disallowed_package_attributes for spdx sbom

joejstuart:
Disallowed package attribute rule for spdx sboms.

https://issues.redhat.com/browse/EC-1102

### DIFF
--- a/policy/release/sbom_spdx/sbom_spdx.rego
+++ b/policy/release/sbom_spdx/sbom_spdx.rego
@@ -56,7 +56,7 @@ deny contains result if {
 #   short_name: allowed
 #   failure_msg: "Package is not allowed: %s"
 #   solution: >-
-#     Update the image to not use a disallowed package.
+#     Update the image to not use any disallowed package.
 #   collections:
 #   - redhat
 #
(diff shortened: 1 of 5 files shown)


chmeliik (approved):
For what it's worth, LGTM from my side :+1:

policy/release/sbom_spdx/sbom_spdx.rego
```diff
@@ -206,3 +206,46 @@ deny contains result if {
...
+# title: Disallowed package attributes
+# description: >-
+#   Confirm the SPDX SBOM contains only packages without disallowed
+#   attributes. By default all attributes are allowed. Use the
+#   "disallowed_attributes" rule data key to provide a list of key-value pairs
+#   that forbid the use of an attribute set to the given value.
+# custom:
+#   short_name: disallowed_package_attributes
+#   failure_msg: Package %s has the attribute %q set%s
+#   solution: Update the image to not use a disallowed package attributes.
```
simonbaird, line 220:
```suggestion
#   solution: Update the image to not use any disallowed package attributes.
```

policy/release/sbom_spdx/sbom_spdx.rego
```diff
@@ -206,3 +206,46 @@ deny contains result if {
...
+	some pkg in s.packages
+
+	some externalref in pkg.externalRefs
+
+	some annotation in pkg.annotations
+	properties := json.unmarshal(annotation.comment)
+	some disallowed in lib.rule_data(sbom.rule_data_attributes_key)
+	properties.name == disallowed.name
+
+	object.get(properties, "value", "") == object.get(disallowed, "value", "")
```
simonbaird, line 236: If the value key is not present in either then it will produce a violation. Is that intentional? Maybe we're expecting the value key will always be present? (Just trying to understand why it's not just this:)

```suggestion
	properties.value == object.value
```

simonbaird: Never mind, I see based on the existing sbom_cyclonedx/sbom_cyclonedx.rego logic.

simonbaird (approved):

PR #1284, opened 2025-02-05
Strip extra digest from base image name annotation

simonbaird:
The rego code here is not expecting the `org.opencontainers.image.base.name` annotation to include a digest, so we end up with a broken image ref containing two digests, which causes a "a digest must contain exactly one '@' separator" error when ec.oci.image_manifest is called.
    
Fix it by using a regex replace to strip off the digest from the name annotation value before putting it together with the digest from the digest annotation value.
    
See also Luiz' explanation in Jira.
    
Note: The regex replace could have been done inline, but moving it to its own rule makes it easier to write the simple test coverage.

Ref: [EC-1041](https://issues.redhat.com/browse/EC-1041)

### DIFF
--- a/policy/release/labels/labels.rego
+++ b/policy/release/labels/labels.rego
@@ -238,8 +238,13 @@ _image_labels := labels if {
 _parent := {"ref": ref, "manifest": manifest, "config": config} if {
 	image_manifest := ec.oci.image_manifest(input.image.ref)
 
-	name := image_manifest.annotations["org.opencontainers.image.base.name"]
+	raw_name := image_manifest.annotations["org.opencontainers.image.base.name"]
 	digest := image_manifest.annotations["org.opencontainers.image.base.digest"]
+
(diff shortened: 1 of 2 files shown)


simonbaird:
I feel like normally this should have a test, but the mocking of the oci functions in `policy/release/labels/labels_test.rego` makes it tricky. I could maybe split out another helper function and test that, will take a look at that tomorrow.

lcarva:
> I feel like normally this should have a test, but the mocking of the oci functions in `policy/release/labels/labels_test.rego` makes it tricky. I could maybe split out another helper function and test that, will take a look at that tomorrow.

`with ec.oci.image_manifest as _my_mock_function` should be sufficient, no?

simonbaird:
> 
> `with ec.oci.image_manifest as _my_mock_function` should be sufficient, no?

I decided to go for a ~~lazier~~ more lightweight approach :grin: .

robnester-rh (approved):
LGTM

PR #1286, opened 2025-02-07
Fix SBOM count condition for base_image_info_found

chmeliik:
Previously, the check asserted that there was at least 1 CycloneDX SBOM. Use the 'sbom.all_sboms' property instead to also count SPDX SBOMs.

Also reword the description of the check. Make it less specific to avoid having to explain where the base images are in each SBOM format. All the check does is assert that there are some SBOMs, anyway.

### DIFF
--- a/antora/docs/modules/ROOT/pages/release_policy.adoc
+++ b/antora/docs/modules/ROOT/pages/release_policy.adoc
@@ -339,7 +339,7 @@ Verify that the base images used when building a container image come from a kno
 [#base_image_registries__base_image_info_found]
 === link:#base_image_registries__base_image_info_found[Base images provided]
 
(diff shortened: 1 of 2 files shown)


chmeliik:
Tested with:

```bash
make fetch-att IMAGE=quay.io/redhat-user-workloads/rhtap-integration-tenant/yq-container/yq@sha256:dde6e1971d0b0eaf8e9393c0d927591d48cb5d138e7496d8cbfb831289beda78 KEY=konflux-pub.key

# contains the base_image_info_found violation on main, doesn't contain it on this branch
make check-release
```

Should fix https://github.com/konflux-ci/yq-container/pull/107#issuecomment-2642073398

joejstuart (approved):

PR #1291, opened 2025-02-10
Idea for skippable test workaround

simonbaird:
WIP, see https://redhat-internal.slack.com/archives/C031J4KBFME/p1739217953467759 for the context.

### DIFF
--- a/policy/release/test/test.rego
+++ b/policy/release/test/test.rego
@@ -208,8 +208,21 @@ deny contains result if {
 #
 deny contains result if {
 	some test in _resulted_in(lib.rule_data("skipped_tests_results"), "n/a")
+	not _skippable_test(test)
 	result := lib.result_helper_with_term(rego.metadata.chain(), [test], test)
 }
+
+# Special case for the ecosystem-cert-preflight-checks. The task determines the
+# type of the artifact and outputs it in the ARTIFACT_TYPE task result. If it's
+# an operatorbundle then the actual preflight check will be skipped entirely.
+# This is so we don't produce a violation in that scenario.
+_skippable_test(test) if {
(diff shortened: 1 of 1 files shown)


policy/release/test/test.rego
```diff
@@ -208,9 +208,22 @@ deny contains result if {
...
+	not _skippable_test(test)
 	result := lib.result_helper_with_term(rego.metadata.chain(), [test], test)
 }
 
+# Special case for the ecosystem-cert-preflight-checks. The task determines the
+# type of the artifact and outputs it in the ARTIFACT_TYPE task result. If it's
+# an operatorbundle then the actual preflight check will be skipped entirely.
+# This is so we don't produce a violation in that scenario.
+_skippable_test(test) if {
+	test == "ecosystem-cert-preflight-checks"
```
lcarva, line 220: :-1: no hard-coded special cases please.

It sounds like the Task should either not be in the pipeline, or it should be marked as SUCCESS if there is nothing to check. The second is the approach taken by the [deprecated-image-checks](https://github.com/konflux-ci/build-definitions/blob/4d7790091825f42b13d9eadb6931defc2b80a275/task/deprecated-image-check/0.5/deprecated-image-check.yaml#L189-L192) Task, for example. I'd be very cautious about tweaking the policies to accommodate for a lack of understanding of how test Tasks should behave.

lcarva (changes requested):

policy/release/test/test.rego
simonbaird: Yeah, I agree this is not a good solution.

simonbaird: On top of that, it's not secure, since ARTIFACT_TYPE could be set to whatever by whoever..

PR #1292, opened 2025-02-10
Allow informative tests to produce skipped result

simonbaird:
WIP, see https://redhat-internal.slack.com/archives/C031J4KBFME/p1739217953467759 for the context.

### DIFF
--- a/policy/release/test/test.rego
+++ b/policy/release/test/test.rego
@@ -208,6 +208,13 @@ deny contains result if {
 #
 deny contains result if {
 	some test in _resulted_in(lib.rule_data("skipped_tests_results"), "n/a")
+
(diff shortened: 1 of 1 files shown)


lcarva (changes requested):
An informative check is one that must complete successfully but may find issues/violations. If the check is skipped, it did not run successfully.

simonbaird:
> An informative check is one that must complete successfully but may find issues/violations. If the check is skipped, it did not run successfully.

Yeah, you're right. I guess we could invent `lib.rule_data("skippable_tests")`, but even then a simple list is not enough to capture the idea that it's supposed to be skippable only when certain other conditions are met.

PR #1293, opened 2025-02-11
Update ec and examples

lcarva:
* Update ec-cli version
* Simplify refresh-examples script: EC now supports reading literals for policy, key, and images. Let's use that to make the script a little bit easier to follow. Also, use the original image reference so we can access the reports from the original image repo.
* Refresh acceptance test example


simonbaird (approved):
I didn't go over it line by line, but lgtm.

hack/refresh-examples.sh
```diff
@@ -31,53 +35,57 @@ naYJRuLprwIv6FDhZ5yFjYUEtsmoNcW7rx2KM6FOXGsCX3BNc7qhHELT+g==
...
 
-trap 'rm "${TRUSTED_TASKS_FILE}-update"' EXIT
+trap 'rm -f "${TRUSTED_TASKS_FILE}-update"' EXIT
 
-go run -modfile "${ROOT_DIR}/go.mod" github.com/enterprise-contract/ec-cli validate image \
-    --images <(cat << EOF
-{
+DIGEST="$(skopeo inspect --no-tags "docker://${IMAGE}" --format '{{index .Digest}}')"
+ORIGINAL_IMAGE_REF="${ORIGINAL_IMAGE_REPO}@${DIGEST}"
+REVISION="$(skopeo inspect --no-tags "docker://${ORIGINAL_IMAGE_REF}" --format '{{index .Labels "vcs-ref"}}')"
```
simonbaird, line 42: Ooh, we should use this technique in hack/simple-demo.sh instead of trying to get it from api.github.com.

Issue #1182, opened 2024-10-10
Per vulnerability policy rules


zregvart:
From https://github.com/enterprise-contract/ec-policies/pull/1180#pullrequestreview-2357463270, given that we have access to the full Clair vulnerability report, rather than the aggregate, this now enables per-vulnerability rules. For example term-based exclusion on a per-vulnerability basis.

lcarva:
Today, if there's a critical CVE violation, EC will emit something like this: `cve.cve_blockers:critical`. This only allows users to exclude a violation based on the CVE severity not the CVE itself. This has its uses, e.g. you may want a more lenient config to gate promotion to staging vs production.

However, what has become increasingly clear is that excluding all the CVEs for a particular severity is too heavy handed in most cases. Thus this GitHub issue.

Let's tweak the existing policy rules to emit violations like this: `cve.cve_blockers:CVE-123-456`. This will allow users to ignore a certain CVE. However, a violation can only emit a single term (the thing after `:`). Users will no longer be able to use `cve.cve_blockers:critical` to ignore a whole severity of CVEs. To still allow that use case, users can override the rule data instead, e.g. `restrict_cve_security_levels`.

So for example, if this is your current policy config:

```yaml
---
spec:
  sources:
    - data:
        - github.com/release-engineering/rhtap-ec-policy//data
        - oci::quay.io/konflux-ci/tekton-catalog/data-acceptable-bundles:latest
      name: Release Policies
      policy:
        - oci::quay.io/enterprise-contract/ec-release-policy:latest
      config:
        exclude:
          - cve.cve_blockers:critical
        include:
          - "@redhat"
```

If you want to continue excluding all critical CVEs, then:

```yaml
---
spec:
  sources:
    - data:
        - github.com/release-engineering/rhtap-ec-policy//data
        - oci::quay.io/konflux-ci/tekton-catalog/data-acceptable-bundles:latest
      name: Release Policies
      policy:
        - oci::quay.io/enterprise-contract/ec-release-policy:latest
      ruleData:
         # Override default value to exclude "critical"
         restrict_cve_security_levels: ["high"]
         # Optionally, report "critical" as a warning
         warn_cve_security_levels: ["critical"]
      config:
        include:
          - "@redhat"
```

And if you want to only exclude a single CVE:

```yaml
---
spec:
  sources:
    - data:
        - github.com/release-engineering/rhtap-ec-policy//data
[...]

PR #1294, opened 2025-02-11
Rework CVE policies to allow per CVE exceptions

lcarva:
NOTE: This PR contains the changes from https://github.com/enterprise-contract/ec-policies/pull/1293. The last commit here is the difference.

This commit changes the CVE policy rules to inspect the full CVE report available in the OCI registry. By doing so, the policies now have the list of CVEs found, instead of just a summary.

The policy rules have been rewritten to emit the CVE name as the term. This allows users to add exceptions for a particular CVE. Previously, it was only possible to add exceptions for a certain CVE level/severity.

The leeway feature was also tweaked. Now, every violation contains an `effective_on` which is the issue date of the vulnerability plus any leeway days that may have been configured. The policies rely on the EC CLI to downgrade violations to warnings when the effective_on date is in the future.

Fixes https://github.com/enterprise-contract/ec-policies/issues/1182
Ref: [EC-921](https://issues.redhat.com/browse/EC-921)


policy/release/cve/cve.rego
```diff
@@ -256,6 +226,45 @@ deny contains result if {
...
+# 	},
+# 	"restrict_unpatched_cve_security_levels": {
+# 		"critical": {vuln3},
+# 		"high": {vuln4},
+# 	}
+# }
+_grouped_vulns[key][level] contains vuln if {
+	some key, patched in _levels_keys_patched
+	levels := lib.rule_data(key)
+	some vuln in _clair_report.vulnerabilities
```
joejstuart, line 243: Is this making the same network call for each entry in `_levels_keys_patched`? If so, I'm wondering if it can be made once to reduce chance for errors and the time it takes?

lcarva: No, not every time, but it's happening twice which I don't know why.

I built a custom version of the EC binary that just emits a loud message when a manifest and blob are fetched. I tried different approaches and can't seem to make it just a single call. I also noticed that this happens to multiple other refs. We should probably do something about that.

That is unrelated to this PR though.

lcarva: Rego does some caching when it comes to rule resolution. My guess is that there are a couple of queries that run in parallel before the caching kicks in.

simonbaird:
Lgtm, but I don't think I understood it deeply.

simonbaird (approved):

ralphbean:
I think this also fixes https://issues.redhat.com/browse/KONFLUX-6843.

lcarva:
Let's merge it before it gets stale.

simonbaird:
> Lgtm, but I don't think I understood it deeply.

Looked a bit closer today (even though it's merged already). The logic feels more understandable now, so nice improvement I think.

PR #1308, opened 2025-03-03
Use gitCommit as intended

### LINKED ISSUE #1037: Revisit implementation of gitCommit
The slsa_source_correlated package supports digest sets of type [gitCommit](https://github.com/enterprise-contract/ec-policies/blob/44ad2c438d8c7692a26159e77f0313f2a792dcd2/policy/release/slsa_source_correlated.rego#L133). Allowing users to specify the following:

```json
"resolvedDependencies": [
  {
    "digest": {
      "gitCommit": "c24001db921c8c2bbb12aeb1c33ae1d910f20504"
    },
    "uri": "git+https://gitlab.com/lucarval/sign-attest-poc"
  }
]
```

The [spec](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md) of the `digest` attribute is responsible for defining how the value for [gitCommit](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#gitcommit-gittree-gitblob-gittag) is computed:

> This hash is computed over `<type> SP <size> NUL <content>`, where:
> 
> -   `<type>` is one of `commit`, `tree`, `blob`, `tag`
> -   `SP` is the ASCII space character, 0x20
> -   `<size>` is the number of bytes in `<content>`, represented as a decimal
>     ASCII number with no leading zeros
> -   `NUL` is the ASCII NUL character, 0x00
> -   `<content>` is git representation of the object:
>     -   For `commit`, the raw commit object ([more info][so-commit][^git-docs])
>     -   For `tree`,  the raw tree object, which is a series of
>         `<unix-octal-mode> <name> NUL <binary-digest>` entries, sorted by
>         `<name>` in the C locale ([more info][so-tree][^git-docs])
>     -   For `blob`, the raw file contents
>     -   For `tag`, the raw tag object

IIUC, in order to compute `gitCommit` we need the raw commit object. This is not available in the SLSA Provenance attestation. More so, our current implementation is basically performing the suggested computed for a `gitBlob` where the blob is the commit ID. This feels not right.

I _think_ the intention of `gitCommit` is to actually just be the git commit ID as is. The description from the spec is just expanding on how git itself computes this information.

Let's use this issue to have another look at our implementation.

NOTE: There's a related feature request in Chains: https://github.com/tektoncd/chains/issues/914

lcarva:
Instead of using `sha1` as a digest algorithm, the in-toto spec allows using `gitCommit` to give additional meaning to the digest: https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#gitcommit-gittree-gitblob-gittag

The spec docs are a somewhat confusing on this which lead it to be interpreted as:

```
gitCommit="$(printf 'commit 3\0'`git rev-parse HEAD` | sha1sum)"
```

However, the intended use of `gitCommit` is to capture a git commit ID as is:

```
gitCommit="$(git rev-parse HEAD)"
```

Fixes #1037

### DIFF
--- a/policy/release/slsa_source_correlated/slsa_source_correlated.rego
+++ b/policy/release/slsa_source_correlated/slsa_source_correlated.rego
@@ -17,10 +17,6 @@ import rego.v1
 import data.lib
 import data.lib.json as j
 
-# opa fmt will transform "\u0000" into "\x00" which subsequently can't be parsed
-# by OPA, see https://github.com/open-policy-agent/opa/issues/6220
-nul := base64.decode("AA==")
-
 # METADATA
 # title: Source code reference provided
 # description: >-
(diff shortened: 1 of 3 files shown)


lcarva:
I'm having a conversation with the in-toto folks [in slack](https://cloud-native.slack.com/archives/C06HBJUEJBT/p1741024208020949) to get further clarification on this.

policy/release/slsa_source_correlated/slsa_source_correlated.rego
```diff
@@ -123,40 +119,25 @@ deny contains result if {
...
-	]),
-	# tolerate slash at the end of the vcs URI and append .git
-	sprintf("%s.git@gitCommit:%s", [
+		# Tolerate missing .git suffix
+		sprintf("%s.git", [expected_vcs_uri]),
+		# Tolerate trailing slash
 		trim_suffix(expected_vcs_uri, "/"),
-		crypto.sha1(sprintf("commit %d%s%s", [count(expected_revision), nul, expected_revision])),
-	]),
+		# Tolerate trailing slack with missing .git suffix
```
simonbaird, line 132:
```suggestion
		# Tolerate trailing slash with missing .git suffix
```

simonbaird (approved):
Lgtm.

policy/release/slsa_source_correlated/slsa_source_correlated.rego
lcarva: :facepalm:

PR #1310, opened 2025-03-06
Verify a set of pullspecs related to an image

joejstuart:
This was originally rolled back because of an issue, but we're unable to reproduce it. So, removing the `Redhat` collection to give room for testing.


policy/release/olm/olm.rego
```diff
@@ -319,6 +288,36 @@ _name(o) := n if {
...
+		r := {
+			"path": "relatedImage",
+			"ref": image.parse(related_ref),
+		}
+	]]
+	some i in imgs
+
+	e := {"ref": i.ref, "path": i.path}
+]
+
```
simonbaird, line 320: This is a large diff, did you mean to move this code from line 215?

policy/release/olm/olm.rego
```diff
@@ -187,8 +187,7 @@ deny contains result if {
 #     Ensure all related images are available. The related images are defined by
 #     an file containing a json array attached to the validated image. The digest
 #     of the attached file is pulled from the RELATED_IMAGES_DIGEST result.
-#   collections:
-#   - redhat
+#
```
simonbaird, line 190: Consider adding a comment here somehow (not in the metadata) to explain why it's no in the collection.

simonbaird (approved):

policy/release/olm/olm.rego
joejstuart: I did. The Regal linter was complaining.

arewm: I thought that I had already moved this. Maybe it was in a different commit from what you took.

PR #1323, opened 2025-03-20
Move SBOM rpm processing to lib

lcarva:
This change moves `all_rpm_entities` and `rpms_from_sbom` to a lib module so it can be used by other policy packages. This is in preparation for EC-1142.


joejstuart (approved):

simonbaird (approved):
Nice.

PR #1345, opened 2025-03-31
feat(KONFLUX-6986): OLM bundle images should not be multi-arch

arewm:
If OLM bundle images are multi-arch then the image pulled by the client will be interpolated to match the client's platform. This can result in inconsitent behaviors between architectures. Instead, a single image manifest should be created since there is no binary and therefore no architecture dependencies.

resolves: KONFLUX-6986

### DIFF
--- a/policy/release/olm/olm.rego
+++ b/policy/release/olm/olm.rego
@@ -192,7 +192,6 @@ deny contains result if {
 #   effective_on: 2025-03-10T00:00:00Z
 #
 deny contains result if {
-	# This rule was removed from the "redhat" collection for testing
 	_release_restrictions_apply
 
 	snapshot_components := input.snapshot.components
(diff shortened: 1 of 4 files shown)


policy/release/olm/olm.rego
```diff
@@ -318,6 +317,29 @@ deny contains result if {
...
+#   short_name: olm_bundle_multi_arch
+#   failure_msg: The %q bundle image is a multi-arch reference.
+#   solution: >-
+#     Rebuild your bundle image without creating an image index.
+#   collections:
+#   - redhat
+#   effective_on: 2025-5-01T00:00:00Z
+deny contains result if {
+	# Parse manifests from snapshot
+	some csv_manifest in _csv_manifests
```
simonbaird, line 335: I think you need this, since otherwise there's no other test making sure csv_manifest is not empty.

```suggestion
	count(csv_manifest) > 0
```

simonbaird: Actually that's wrong. Disregard!

simonbaird: The some implies it can't be empty.

simonbaird (approved):

nmars (approved):
LGTM

PR #1347, opened 2025-03-31
Make image.parse only validates valid hostnames

joejstuart:
image.parse was allowing invalid hostnames.
Adding a regex to strengthen the validation.

https://issues.redhat.com/browse/EC-1153

### DIFF
--- a/policy/lib/image/image.rego
+++ b/policy/lib/image/image.rego
@@ -5,9 +5,18 @@ import rego.v1
 # parse returns a data structure representing the different portions
 # of the OCI image reference.
 parse(ref) := d if {
-	digest_parts := split(trim_space(ref), "@")
+	trimmed_ref := trim_space(ref)
 
-	contains(digest_parts[0], "/")
+	# Note: This regex is simplified and does not cover all valid hostname cases.
+	# It only matches hostnames in the form of registry.local' or 'Registry1.io'.
+	# It does not include all subdomains and does not support Unicode.
+	regex.match(`^(?:[a-zA-Z0-9-])+\.[a-zA-Z]+`, trimmed_ref)
+
+	# a valid repo will contain a /
+	contains(trimmed_ref, "/")
+
+	digest_parts := split(trimmed_ref, "@")
+
 	repo_parts := split(digest_parts[0], "/")
 
 	tag_parts := split(repo_parts[count(repo_parts) - 1], ":")
@@ -21,8 +30,6 @@ parse(ref) := d if {
 			[tag_parts[0]],
 		),
 	)
-
-	not contains(repo, "://")
 
 	digest := _get(digest_parts, 1, "")
 
--- a/policy/lib/image/image_test.rego
+++ b/policy/lib/image/image_test.rego
@@ -9,7 +9,6 @@ import data.lib.image
 test_parse if {
 	repository := "registry.com/re/po"
 	repository_with_port := "registry.com:8443/re/po"
-	bad_repository := "http://not-a-registry.com"
 	tag := "latest"
 	digest := "sha256:01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b"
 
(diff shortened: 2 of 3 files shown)


policy/lib/image/image.rego
```diff
@@ -7,9 +7,11 @@ import rego.v1
 parse(ref) := d if {
 	digest_parts := split(trim_space(ref), "@")
 
-	contains(digest_parts[0], "/")
 	repo_parts := split(digest_parts[0], "/")
 
+	# match hostname. i.e registry.local
+	regex.match("^(?:[a-zA-Z0-9-])+\\.[a-zA-Z]+", repo_parts[0])
```
lcarva, line 13: Unicode characters, although unusual, are valid host names.

simonbaird:
I don't understand how this relates to the Jira. Am I missing something?

policy/lib/image/image.rego
simonbaird: If you think it's too hard to properly match all valid host names, maybe it's acceptable with a comment, e.g.:
```
  # This incorrectly disallows host names with unicode chars
```

But, (as per my other comment), can you explain why are we're doing this?

lcarva: My guess is that EC is incorrectly assuming a certain string is a hostname when it's something else completely.

This is mentioned in the docs earlier in this file, but just in case it was missed, this check was re-implemented here from OSBS, see their [docs](https://osbs.readthedocs.io/en/latest/users.html#pullspec-locations) which also link to the implementation: https://github.com/containerbuildsystem/atomic-reactor/blob/14b9ad84b281029391f1bb83dc5dee6c573eeb4b/atomic_reactor/utils/operator.py#L316

I believe `alnum` will actually handle unicode characters as well.

joejstuart: The Jira didn't describe it well. `image.parse` was parsing `'{"kind":"Namespace","apiVersion":"v1","metadata":{"name":"openshift-workload-availability","annotations":{"openshift.io/node-selector":""}}}'` and returning `"annotations":{"openshift.io/node-selector":""}` as a repo.

Do we care about unicode? I think adding the comment that it disallows it would be enough?

joejstuart:
> I don't understand how this relates to the Jira. Am I missing something?

https://github.com/enterprise-contract/ec-policies/pull/1347#discussion_r2021444405

policy/lib/image/image.rego
simonbaird: Ooooh, so the entire string was being treated like it might be an image ref.

joejstuart: yeah. I'm also finding a solid regex for matching hosts is ugly, so I thought I'd start with this.

simonbaird: But why was it returning `"annotations":{"openshift.io/node-selector":""}`. Because of the comma?

simonbaird: Alternative idea: We could flip it and look for "{" or "}" chars, and if they're found we assume it's not a url.

simonbaird: ..with some explanatory comments to explain why.

simonbaird: (Not saying that's better, just something to consider.)

joejstuart: Are there reservations about this change or can it move forward? I added a few more tests if it helps. https://github.com/enterprise-contract/ec-policies/pull/1347/commits/82ab6d5ce93fd6f41d4491723390926249f085c4

policy/lib/image/image_test.rego
```diff
@@ -47,8 +46,15 @@ test_parse if {
...
 		{"repo": repository_with_port, "tag": tag, "digest": ""},
 	)
+}
 
-	not image.parse(concat("", [bad_repository, ":", tag, "@", digest]))
+test_not_parse if {
+	tag := "latest"
+	digest := "sha256:01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b"
+	not image.parse(concat("", ["http://not-a-registry.com", ":", tag, "@", digest]))
+	not image.parse("oci://not-a-registry.com")
```
simonbaird, line 55: It might be unexpected for some that oci://valid.registry.io/org/repo fails, but it's clearly shown by this test, so we can worry about it when/if we need to. Lgtm.

simonbaird (approved):

policy/lib/image/image_test.rego
joejstuart: fwiw, I thought that too, but the original code seemed to block anything with `://`. https://github.com/enterprise-contract/ec-policies/blob/main/policy/lib/image/image.rego#L25

PR #1361, opened 2025-04-10
Support both old and new names for cachi2/hermeto

simonbaird:
The name is changing and so the annotations we need to look for in sbom component/package lists are changing accordingly. Match both the old and new names so we can transition smoothly.

There are two different matches, one for CycloneDX and one for SPDX formatted SBOMs.

Ref: https://issues.redhat.com/browse/EC-1176

### DIFF
--- a/policy/lib/sbom/rpm.rego
+++ b/policy/lib/sbom/rpm.rego
@@ -44,21 +44,29 @@ _is_rpmish(purl) if {
 	startswith(purl, "pkg:rpmmod/")
 }
 
+# CycloneDX style
 _component_found_by_cachi2(component) if {
 	some property in component.properties
-	property == cachi2_found_by_property
+	some cachi2_name in _cachi2_names
+	property == _cachi2_found_by_property(cachi2_name)
 } else := false
 
-# This is what cachi2 produces in the component property list
-cachi2_found_by_property := {
-	"name": "cachi2:found_by",
-	"value": "cachi2",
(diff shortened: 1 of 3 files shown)


policy/lib/sbom/rpm.rego
```diff
@@ -44,21 +44,29 @@ _is_rpmish(purl) if {
...
+# Expecting this to be called with one of _cachi2_names
+_cachi2_found_by_property(cachi2_name) := {
+	"name": sprintf("%s:found_by", [cachi2_name]),
+	"value": cachi2_name,
 }
 
+# SPDX style
 _package_found_by_cachi2(pkg) if {
 	some annotation in pkg.annotations
-	regex.match(`.*cachi2.*`, annotation.annotator)
```
simonbaird, line 60: @joejstuart it's unrelated to this PR, but I wondering about the logic here. What if cachi2 decided to annotate with a found_by:some_other_thing? Maybe it's not likely to happen in real life, but it seems we're actually ignoring the "found by" information in this annotation.

joejstuart (approved):

policy/lib/sbom/rpm.rego
joejstuart: I'm not following. Do you think the regex should be more strict?

simonbaird: Perhaps I'll make a PR to show what I mean.

PR #1364, opened 2025-04-15
Mention Trusted Artifacts in Conforma docs

dheerajodha:
* along with the Trusted Task List, which is a
  technique used to validate tasks for pipeline
  security.
* this technique is also used by Trusted
  Artifacts but not for ALL the tasks, but only
  for those ones that directly impact the
  pipelines output.

resolves: EC-948


antora/docs/modules/ROOT/pages/trusted_tasks.adoc
```diff
@@ -66,3 +71,44 @@ compatibility. Will be removed entirely in future.
...
+
+* Workspaces are allowed for non-sharing purposes, like mounting Secrets.
+
+=== Migration and Task Variants To aid transition:
+
+* New variants of standard tasks exist, using the suffix -oci-ta, e.g. git-clone-oci-ta.
+
+* New tasks with fresh functionality that use Trusted Artifacts need not follow the suffix naming convention.
+
+
```
simonbaird, line 109: This is pretty good. I think there's a little more to explain, something like this:

```
When using Trusted Artifacts, only the tasks that directly impact the pipeline's output must be in the trusted task list. Typically this is the git clone task, the prefetch dependencies task, and the build tasks.

In contrast, when using the non-trusted artifacts pipeline, **every** task in the pipeline must be in the trusted task list.
```

I don't think this is perfect explanation, but that's the main thing we want people to understand.

antora/docs/modules/ROOT/pages/trusted_tasks.adoc
```diff
@@ -1,8 +1,13 @@
-= Trusted Tasks
+= Trusted Tasks and Trusted Artifacts
+
+Konflux supports two mechanisms for ensuring secure and compliant task execution in build pipelines:
+
+1. Using a Trusted Task List.
+2. Sharing files securely between tasks via Trusted Artifacts.
```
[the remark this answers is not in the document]
dheerajodha: Thank you, this made things clearer.

This is just for my understanding: we use trusted tasks while using trusted artifacts as visible from the fact that there's an entry for the git-clone-oci-ta task within the `trusted_tekton_tasks.yml` file right [here](https://github.com/enterprise-contract/ec-policies/blob/7c178676023ed070332674dca334d9b2c0a37fb8/example/data/trusted_tekton_tasks.yml#L966). This means we are using a trusted task (git-clone-oci-ta) while using the trusted artifacts (since we're using the created trusted artifact [right here](https://github.com/konflux-ci/build-definitions/blob/14ac82ab4ee9d48a67429cebb2eaa7ba14eb2708/task/git-clone-oci-ta/0.1/git-clone-oci-ta.yaml#L355) within the same git-clone-oci-ta task)

antora/docs/modules/ROOT/pages/trusted_tasks.adoc
dheerajodha: What if I'm using the "Trusted Artifacts" (TA) version of the build pipeline and have included a custom task as well?

Now Conforma would be okay with my custom task because the build-influencing tasks (clone, prefetch, and build) use TA. But is it still possible to include a malicious build-influencing logic in my custom task that could compromise the security of my build? If not, why not?

simonbaird: The idea is that the malicious task could not influence the output of the trusted build task, because the build task is only accessing data from the trusted artifact produced by (one or more) trusted tasks.

I'm not saying this is bulletproof, but it should prevent a range of code tampering attack vectors, such as "git-clone" -> "modify code, add malware" -> "buildah build". The shared workspace used in non-TA pipelines makes that kind of attack quite doable if there are non-trusted, hence potentially malware, tasks allowed in the pipeline.

I guess also we are counting on (or hoping that) Tekton and Kubernetes to provide enough isolation so one task can't reach out and do bad things to another task or pod while it runs.

antora/docs/modules/ROOT/pages/trusted_tasks.adoc
```diff
@@ -36,7 +42,7 @@ would result in such a warning.
 The process of adding Tasks to the list of trusted Tasks is described
 xref:trusting_tasks.adoc[here].
 
-== Trusted Task list
+=== Example Structure
```
simonbaird, line 45: FYI, as mentioned on line 61, the `task-bundles` key and its contents is deprecated. It says we're removing it. This might be a good time. Wdyt @lcarva ?

Not a blocker for merging, but you could consider doing it while you're looking at this, maybe in a commit in this PR, or in another PR.

simonbaird (approved):

antora/docs/modules/ROOT/pages/trusted_tasks.adoc
dheerajodha: This helps a lot, thanks!

antora/docs/modules/ROOT/pages/trusted_tasks.adoc
lcarva: No concerns. Nothing should be using it anymore. I do see [some reference](https://github.com/search?q=repo%3Aenterprise-contract%2Fec-policies%20task-bundles&type=code) but they seem to be leftover tests. We should probably remove those too at some point.

dheerajodha: Thank you for confirming!
I tried deleting the leftover references in [this commit](https://github.com/enterprise-contract/ec-policies/pull/1364/commits/0a2530394e545b7b9ef9579c7dd674b1ba277be0).

dheerajodha: The CI check is complaining about the reduced test coverage ([here](https://github.com/enterprise-contract/ec-policies/actions/runs/14739175539/job/41375202403?pr=1364#step:6:9)), which is the result of the removal of the test-bundle-related tests. Should we ignore this CI check failure?

dheerajodha: I agree with you, let's wind this PR back to a docs change only. Thanks for the suggestions!!

simonbaird (approved):

PR #1371, opened 2025-04-22
Include snapshot digests in trusted build digests

simonbaird:
Adjust the trusted parameters rule so that the digests of images included the snapshot are allowed in parameters.

PR includes two additional minor fix-up commits.

Ref: [EC-1086](https://issues.redhat.com/browse/EC-1086)


simonbaird:
Fixed some Styra gripes and also added some defensive edge case checks in case the snapshot components have no digest.

simonbaird:
For reviewers: Top commit has the significant change so maybe look at that separately.

policy/release/trusted_task/trusted_task.rego
```diff
@@ -307,6 +308,23 @@ _trusted_build_digests contains digest if {
 	some digest in _digests_from_values(lib.result_values(result))
 }
 
+# If an image is part of the snapshot we assume that was built in Konflux and
+# therefore it is considered trustworthy. IIUC the use case is something to do
+# with building an image in one component, and being able to use it while
+# building another component in the same application.
+_trusted_build_digests contains digest if {
+	some component in input.snapshot.components
```
dheerajodha, line 316: How does this `input` variable look? trying to understand its content.

dheerajodha (commented):
Small question, the rest of the logic lgtm!

policy/release/trusted_task/trusted_task.rego
simonbaird: Input is the big global input with everything in it. You can see what it looks like using `--output policy-input=input.json` with ec validate image (or hack/simple-demo.sh). See also <https://conforma.dev/docs/ec-cli/policy_input.html> .

joejstuart (approved):

st3penta (approved):

dheerajodha (approved):

simonbaird:
> LGTM.
> 
> Imho the check about the digest being a string could be omitted because it comes from functions that only accept strings ([here](https://github.com/enterprise-contract/ec-policies/blob/main/policy/lib/image/image.rego#L16-L18)). But if you think it's worth a double check then ok.

Yeah, I was on the fence about whether it was a good idea or not. My thinking is it could possibly be useful if someday/somehow image.parse got re-implemented and it instead return a null. It's probably unlikely, but figured the cost of the is_string check it is low enough to keep it.

simonbaird:
Thanks very much for all the reviews. I'm gonna merge it.

PR #1373, opened 2025-04-24
Emit warning for bundle references without tags

st3penta:
Extend the trusted tasks lookup logic to also support bundle references without tag information, as long as a match for the digest is found. Now the presence of an untagged task reference throws a warning instead of a violation.

Ref: https://issues.redhat.com/browse/EC-712

### DIFF
--- a/policy/lib/tekton/trusted.rego
+++ b/policy/lib/tekton/trusted.rego
@@ -8,6 +8,14 @@ import data.lib.time as time_lib
 # regal ignore:prefer-package-imports
 import data.lib.rule_data as lib_rule_data
 
+# Returns a subset of tasks that use untagged bundle Task references.
+untagged_task_references(tasks) := {task |
+	some task in tasks
+	ref := task_ref(task)
+	ref.bundle
+	not ref.tagged
+}
+
 # Returns a subset of tasks that use unpinned Task references.
 unpinned_task_references(tasks) := {task |
 	some task in tasks
@@ -50,14 +58,30 @@ untrusted_task_refs(tasks) := {task |
 # Returns true if the task uses a trusted Task reference.
 is_trusted_task(task) if {
 	ref := task_ref(task)
-	records := _trusted_tasks[ref.key]
 
-	some record in records
+	some record in trusted_task_records(ref.key)
 
 	# A trusted task reference is one that is recorded in the trusted tasks data, this is done by
 	# matching its pinned reference; note no care is given to the expiry or freshness since expired
 	# records have already been filtered out.
 	record.ref == ref.pinned_ref
+}
+
+trusted_task_records(ref_key) := records if {
+	# the reference key matches exactly the key in the trusted tasks set
+	records := _trusted_tasks[ref_key]
+	count(records) > 0
+} else := records if {
+	startswith(ref_key, "oci://") # only for oci refs
+	records := [match |
+		some key, matches in _trusted_tasks
+		short_key := regex.replace(key, `:[0-9.]+$`, "")
+		ref_key == short_key
+		some match in matches
+	]
+} else := records if {
+	# If the key is not found, default to an empty list
+	records := []
 }
 
 latest_trusted_ref(task) := trusted_task_ref if {
--- a/policy/lib/tekton/trusted_test.rego
+++ b/policy/lib/tekton/trusted_test.rego
@@ -135,6 +135,21 @@ test_is_trusted_task if {
 	tekton.is_trusted_task(newest_trusted_git_task) with data.trusted_tasks as future_trusted_tasks
 }
 
+test_trusted_task_records if {
+	task_ref_expected_matches := {
+		"oci://registry.local/trusty:1.0": 3,
+		"oci://registry.local/trusty": 3,
+		"git+git.local/repo.git//tasks/honest-abe.yaml": 2,
+		"git+git.local/repo.git//tasks/untrusted.yaml": 0,
+		"oci://reg": 0,
+	}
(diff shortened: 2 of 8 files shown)


policy/lib/tekton/trusted.rego
```diff
@@ -50,7 +58,9 @@ untrusted_task_refs(tasks) := {task |
 # Returns true if the task uses a trusted Task reference.
 is_trusted_task(task) if {
 	ref := task_ref(task)
-	records := _trusted_tasks[ref.key]
+
+	some key, records in _trusted_tasks
+	contains(key, ref.key)
```
st3penta, line 63: I'm not sure this approach is the best one...depending on the average cardinality of the `_trusted_tasks` object, I'm concerned this could affect performance, since it would become O(n) instead of O(1).

WDYT?

simonbaird: It could be significant, especially because we have many hundreds of keys. Even so, my hunch is that the Rego evaluation is so fast in comparison to the time spent fetching the data, that we can get away with it.

simonbaird: But how about this, since we expect it's either `$repo:$tag` or just `$repo`

```suggestion
	some key, records in _trusted_tasks
	startswith(key, ref.key)
```

It also might be a little faster.

simonbaird: Something to consider:

We want this new matching behavior to work for the oci refs only, so we might want tighten it up for the git+https keys.

One possible way to do that:
```suggestion
	some key, records in _trusted_tasks
	_key_matches_ref(key, ref.key)
```

Then later:

```
_key_matches_ref(key, ref_key) := true if {
  startswith(key, 'oci://') # only for oci refs
  startswith(key, ref_key) # match with or without the tag
} else := true if {
  key == ref_key
}
```

Another approach:

```
_key_matches_ref(key, ref_key) := true if {
  # Some special handling only for oci refs
  startswith(key, 'oci://')
  # Try matching with the tag (e.g. `:0.2`) removed
  short_key := regex.replace(key, ":[0-9.]+$", "") `:0.2` 
  short_key == ref_key
} else := true if {
  # Otherwise it should be the exact key
  key == ref_key
}
```

(I'm just sharing some ideas, not saying it must be this way.)

st3penta: I like the ideas of using `startswith` instead of `contains`, and also of doing the less performant check only for `oci` refs. Also: the startswith logic should run only if there is no exact match for the key.
I like less the regex replace, I find the startswith  more readable (and probably comparable in performance).

I [refactored the function](https://github.com/enterprise-contract/ec-policies/pull/1373/files#diff-17d26aa4245574b4f639b10f7c6ad71340f7569fdb6c80534e861d145368d9e5R62-R85) to include the feedbacks, let me know what you think!

simonbaird: Even better! Very nice.

st3penta: Thinking more about it, I probably found a potential security issue: the logic with the `startswith` could easily match malicious bundle refs such as `oci://a`. 
Probably the safer alternative would be the regex approach, stripping the tag ref from the key and looking for an exact match.

st3penta: The regex you provided prevents the above edge case, so in the end I used it instead of the `startswith`

policy/release/trusted_task/trusted_task_test.rego
```diff
@@ -81,6 +100,7 @@ test_trusted_violation if {
 		"buildType": lib.tekton_pipeline_run,
 		"buildConfig": {"tasks": [
 			trusted_bundle_pipeline_task,
+			untagged_bundle_pipeline_task,
```
simonbaird, line 103: I was wondering about the test coverage. IIUC this here is enough to get it to 100%, right? ..Since line 535 is a key that is matched by the new matcher.

(Even so, you could consider adding little extra test for the new `trusted_task_records` function.)

simonbaird: ...I guess it would belong in policy/lib/tekton/trusted_test.rego not here.

simonbaird (approved):

policy/release/trusted_task/trusted_task_test.rego
st3penta: [test added](https://github.com/enterprise-contract/ec-policies/pull/1373/files#diff-93154068b9b2123e6a1de36c8c345d6b822cb09358d9e947ddebb81a41fb0af4R138-R152)

simonbaird: Awesome, thanks.

policy/lib/tekton/trusted_test.rego
```diff
@@ -135,6 +135,21 @@ test_is_trusted_task if {
...
+test_trusted_task_records if {
+	task_ref_expected_matches := {
+		"oci://registry.local/trusty:1.0": 3,
+		"oci://registry.local/trusty": 3,
+		"git+git.local/repo.git//tasks/honest-abe.yaml": 2,
+		"git+git.local/repo.git//tasks/untrusted.yaml": 0,
+		"oci://reg": 0,
+	}
+
+	every ref, expected in task_ref_expected_matches {
```
simonbaird, line 147: This looks like a really nice tidy way to test a list of cases. Very nice! :monocle_face:

st3penta: thanks! :blush:

PR #1377, opened 2025-04-28
Push effective date on rpm_packages.unique_version

simonbaird:
This check went active today, and it sems to be causing problems. While we figure out the details, let's disable it. Rather than revert, I've pushed it the effective date back two months. (Of course we can bring if forward or push it back again if needed.)

Ref: https://issues.redhat.com/browse/EC-1142

Original story:
Ref: https://issues.redhat.com/browse/EC-1232


robnester-rh (approved):
LGTM

PR #1378, opened 2025-04-30
Use binaries for OPA IDE integration

lcarva:
The OPA VSCode plugin relies on the OPA and Regal binaries to be available on the system. The exact version of those matter. In fact, the OPA binary used has to be the one wrapped by the EC CLI, i.e. `ec opa`. To make this easier, we track those dependencies in go.mod and use bash scripts to call `go run ...` on the corresponding modules. This works ok, but it's kind of slow, and it's likely the reason the VSCode plugin crashes all the time.

This commit changes the bash script to no longer use `go run`. Instead, it introduces Make targets to create the required binaries. The bash scripts are modified to automatically call the Make targets, which should be a no-op if the binaries don't need to be created, and then invoking the binary directly.

This should give us a snappier development experience.

### DIFF
--- a/hack/ec-opa.sh
+++ b/hack/ec-opa.sh
@@ -19,4 +19,5 @@
 # by EC, e.g. ec.oci.image_files, a custom version of OPA is required.
 set -euo pipefail
 cd "$(dirname "$0")/.."
-exec go run github.com/enterprise-contract/ec-cli opa "$@"
+make --silent bin/ec
+exec bin/ec opa "$@"
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules/
 .cache/
 *.swp
 coverage.json
+bin/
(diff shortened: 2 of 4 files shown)


hack/ec-opa.sh
```diff
@@ -19,4 +19,5 @@
 # by EC, e.g. ec.oci.image_files, a custom version of OPA is required.
 set -euo pipefail
 cd "$(dirname "$0")/.."
-exec go run github.com/enterprise-contract/ec-cli opa "$@"
+make ec-bin
```
simonbaird, line 22: I think if the target is phony, it will want to build it every time, which is defeating the purpose. I think you need a non-phony target that is the filename, or do a check here in bash to see if the bin/ec exists already, something like  `[ -x bin/ec ] || make ec-bin`

lcarva: Thank you! Updated.

simonbaird (approved):

PR #1394, opened 2025-05-20
Improve 'Untrusted Task' error for outdated tasks

st3penta:
The error message for the trusted_task.trusted violation is misleading when a task is flagged as untrusted because its digest is out of date.

This commit improves the message by pointing out the trusted digest for a given untrusted task.

Ref: https://issues.redhat.com/browse/EC-1269


simonbaird (approved):
Nice.

policy/release/trusted_task/trusted_task.rego
```diff
@@ -354,3 +346,29 @@ _digests_from_values(values) := {digest |
 	some pattern in _digest_patterns
 	some digest in regex.find_n(pattern, value, -1)
 }
+
+_format_trust_error(task, dependency_chain) := error if {
+	latest_trusted_ref := tekton.latest_trusted_ref(task)
```
joejstuart, line 351: I think this selects tasks that are "active" and not necessarily the latest task ref. The function that eventually gets called and makes me think this. https://github.com/enterprise-contract/ec-policies/blob/main/policy/lib/time/time.rego#L88

st3penta: So you're saying that the actual latest version of the task could be filtered out, if its `effective_on` is still in the future, right?
You're suggesting to change the name of this variable to something like `latest_active_trusted_ref`?

st3penta: Yes the threshold computed in `_active_threshold` is the most recent date that is not in the future, but then it is used [here](https://github.com/enterprise-contract/ec-policies/blob/main/policy/lib/time/time.rego#L72) to retrieve all the valid tasks, including the future ones, no?

simonbaird: I do agree that it would ideally show the newest possible bundle ref.

I think EC-424 is going to change this, since IIUC effective_on is being thrown out as soon as possible. In that scenario, I guess the ref without any "expires_on" field is the newest and best choice.

simonbaird: ..with a fallback to one with the latest "expires_on", just in case there are none without an "expires_on".

st3penta: Yes this seems the most reasonable approach for dealing with 'recency' once we remove the effective_on dependency, although technically it might not be as accurate as the effective_on logic.

Anyway i think this discussion is more related to [EC-424](https://issues.redhat.com/browse/EC-424) than to this PR: here we only rely on `latest_trusted_ref`, we don't care about the implementation as long as the function's purpose doesn't change

st3penta:
On a related note, I can't find any kind of sorting of the `_trusted_tasks` variable: it seems that `latest_trusted_ref` only relies on the fact that the latest trusted task is the first one in the list.
Are we ok with that, or should we add a sorting instruction in the  `latest_trusted_ref` function?

simonbaird:
> On a related note, I can't find any kind of sorting of the `_trusted_tasks` variable: it seems that `latest_trusted_ref` only relies on the fact that the latest trusted task is the first one in the list. Are we ok with that, or should we add a sorting instruction in the `latest_trusted_ref` function?

Up to you I think. But it seems like it would be not too hard to sort [this](https://github.com/enterprise-contract/ec-policies/pull/1395/files#diff-17d26aa4245574b4f639b10f7c6ad71340f7569fdb6c80534e861d145368d9e5R110-R114) and make it more robust.

(FWIW, I have assumed elsewhere that the first one is the newest, see [here](https://github.com/simonbaird/konflux-pipeline-patcher/blob/main/pipeline-patcher#L56-L57). I think it's a reasonably safe assumption for now, but I do think it's better to be defensive and explicitly sort.

simonbaird (approved):

simonbaird:
Lgtm.

PR #1395, opened 2025-05-22
Remove effective_on logic from trusted tasks rules

st3penta:
The 'effective_on' field was introduced to deal with sets of required tasks that would become effective some time in the future, so it was not meant to handle the validity of the trusted tasks.

For this reason it has been removed from the trusted tasks, and the rules that relied on that were migrated to the 'expires_on' field.

Ref: https://issues.redhat.com/browse/EC-424


policy/lib/tekton/trusted.rego
```diff
@@ -101,11 +101,26 @@ _task_expires_on(task) := expires if {
 	expires = time.parse_rfc3339_ns(record.expires_on)
 }
 
+_unexpired_records(records) := all_unexpired if {
+	never_expires := [record |
+		some record in records
+		not record.expires_on
```
simonbaird, line 107: I think it works the same (unless maybe the value of expires_on is `false`), but consider doing it like this for readability:

```suggestion
		"expires_on" in object.keys(obj)
```

policy/lib/tekton/trusted.rego
```diff
@@ -101,11 +101,26 @@ _task_expires_on(task) := expires if {
 	expires = time.parse_rfc3339_ns(record.expires_on)
 }
 
+_unexpired_records(records) := all_unexpired if {
```
simonbaird, line 104: Perhaps it's worth adding a small simple test to exercise/demonstrate _unexpired_records behavior, since that's a new rule being added here.

policy/lib/tekton/trusted.rego
```diff
@@ -101,11 +102,27 @@ _task_expires_on(task) := expires if {
...
+		some record in records
+		not "expires_on" in object.keys(record)
+	]
+
+	future_expires := [record |
+		some record in records
+		expires := time.parse_rfc3339_ns(record.expires_on)
+		expires > time_lib.effective_current_time_ns
+	]
+	future_expires_sorted := array.reverse(arrays.sort_by("expires_on", future_expires))
```
st3penta, line 116: This sorting was added to address [this comment](https://github.com/enterprise-contract/ec-policies/pull/1394#issuecomment-2901645717)

PR #1397, opened 2025-05-26
feat: Bump spdx schema to 2.3.1

Allda:
In the SPDX 2.3 schema there have been couple of breaking changes introduced that made the schema not compatible with older version. We found out that a 2.3 schema refuses a valid documents generated by the spdx-tools libraries due to this incompatibility.

This commit introduces a schema 2.3.1 which is backward compatible and makes some enums compatible with multiple supported values.

One of the examples fixed in this schema is:

"PACKAGE-MANAGER" vs "PACKAGE_MANAGER" for referenceCategory field.

Related Github issue: https://github.com/spdx/spdx-spec/issues/792

Jira Issue: https://issues.redhat.com/browse/EC-1292

### DIFF
--- a/policy/release/sbom_spdx/sbom_spdx_test.rego
+++ b/policy/release/sbom_spdx/sbom_spdx_test.rego
@@ -339,6 +339,7 @@ _sbom_attestation := {"statement": {
 	"predicateType": "https://spdx.dev/Document",
 	"predicate": {
 		"spdxVersion": "SPDX-2.3",
+		"documentNamespace": "https://example.dev/spdxdocs/example-310683af-e9a0-4f66-a6a4-119352915b51",
 		"dataLicense": "CC0-1.0",
 		"SPDXID": "SPDXRef-DOCUMENT",
 		"name": "registry.local/bacon@sha256:123",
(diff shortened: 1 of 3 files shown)


policy/release/sbom_spdx/sbom_spdx_test.rego
```diff
@@ -337,6 +337,7 @@ test_attributes_multiple_external_refs if {
 
 _sbom_attestation := {"statement": {
 	"predicateType": "https://spdx.dev/Document",
+	"documentNamespace": "https://example.dev/spdxdocs/example-310683af-e9a0-4f66-a6a4-119352915b51",
```
dheerajodha, line 340: Shouldn't the `documentNamespace` field be within the `predicate` field?
Because the [tests are failing](https://github.com/enterprise-contract/ec-policies/actions/runs/15255739775/job/42957338652?pr=1397) with:
```
"msg": "SPDX SBOM at index 0 is not valid: (Root): documentNamespace is required"}
```

simonbaird:
> Would it be better if we support both schemas for a while?

Reading it closer, I can see it's supposed to be backwards compatible, so no need to support the old schema IIUC. Let me see if I can get tests passing.

robnester-rh (approved):
LGTM

PR #1406, opened 2025-06-03
Correctly read base image digests from spdx sboms

simonbaird:
We need to look inside an externalRef to extract a purl, then read the digest from that, because the name field doesn't contain it.

Ref: https://issues.redhat.com/browse/EC-1298


joejstuart (approved):

policy/release/base_image_registries/base_image_registries.rego
```diff
@@ -134,28 +134,60 @@ _is_cyclonedx_base_image(component) if {
...
+	qualifier.key == "repository_url"
+
+	# Example repo_url: "someregistry.io/someorg/someapp"
+	# It's probably the same as pkg.name, but let's use the value from the purl
+	repo_url := qualifier.value
+
+	# Put them together to make a pinned image_ref
+	image_ref := sprintf("%s@%s", [repo_url, image_digest])
+}
+
```
simonbaird, line 190: Did I re-implement `sbom.image_ref_from_purl` here???

PR #1422, opened 2025-06-13
Enforce all pre-build script images to be in SBOM

st3penta:
This ensures that all the images used during the build process (or in this case during the pre-build phase) are included in the SBOM.

I also created an additional rule that enforces that all the pre-build script tasks must have an output named SCRIPT_RUNNER_IMAGE_REFERENCE. This is to prevent the case in which a task doesn't expose the above output, resulting in a missing image in the SBOM.

Ref: https://issues.redhat.com/browse/EC-1242


policy/release/pre_build_script_task/pre_build_script_task.rego
```diff
@@ -0,0 +1,88 @@
...
+
+import data.lib
+import data.lib.sbom
+import data.lib.tekton
+
+# METADATA
+# title: Script runner image is listed in the task results
+# description: >-
+#   Verify that the image used to run the pre-build script task is
+#   listed in the task output SCRIPT_RUNNER_IMAGE_REFERENCE
```
simonbaird, line 20: It's a tekton task result so this seems like a better description:

```suggestion
#   listed in the task result SCRIPT_RUNNER_IMAGE_REFERENCE
```

policy/release/pre_build_script_task/pre_build_script_task.rego
```diff
@@ -0,0 +1,88 @@
...
+#   solution: >-
+#     Make sure the image used to run the pre-build task is referenced
+#     in the 'SCRIPT_RUNNER_IMAGE_REFERENCE' task result.
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	some attestation in lib.pipelinerun_attestations
+	some task in tekton.pre_build_tasks(attestation)
+	not tekton.task_result(task, _pre_build_run_script_runner_image_result)
```
simonbaird, line 35: Not a big deal, and I'm not suggesting to change it, but I notice the actual check is doing a bit less than what the description in the metadata says. I.e. this would still pass if any old garbage was in the task result value, right?

policy/release/pre_build_script_task/pre_build_script_task_test.rego
```diff
@@ -0,0 +1,106 @@
...
+		},
+	]}]},
+}}
+
+_cyclonedx_sbom_attestation := {"statement": {
+	"predicateType": "https://cyclonedx.org/bom",
+	"predicate": {"components": [
+		{"purl": "oci://example.com/org/spam:v0.2"},
+		{"purl": "registry.redhat.io/ubi7:latest@sha256:abc"},
+		{"purl": "quay.io/konflux-ci/bazel6-ubi9:latest@sha256:bcd"},
```
simonbaird, line 104: Yeah, these don't look like real purls. 

Some real examples:

```
cosign download sbom quay.io/redhat-user-workloads/rhtap-contract-tenant/ec-v06/cli-v06@sha256:1e88beb02d865293a67c8efea27402d69a3b3c722835e74d84574930307f5663|grep 'pkg:oci'
```

policy/lib/sbom/sbom_test.rego
```diff
@@ -136,6 +136,15 @@ test_ignore_unrelated_sboms if {
...
 
+test_image_ref_from_purl if {
+	# regal ignore:line-length
+	purl := "pkg:oci/ubi-minimal@sha256:92b1d5747a93608b6adb64dfd54515c3c5a360802db4706765ff3d8470df6290?repository_url=registry.access.redhat.com/ubi9/ubi-minimal"
+
+	# regal ignore:line-length
+	image_ref := "registry.access.redhat.com/ubi9/ubi-minimal@sha256:92b1d5747a93608b6adb64dfd54515c3c5a360802db4706765ff3d8470df6290"
+	lib.assert_equal(sbom.image_ref_from_purl(purl), image_ref)
+}
+
```
simonbaird, line 147: Nice.

PR #1431, opened 2025-06-27
Ignore tag on pre-build sbom ref comparison

simonbaird:
We're expecting the value of SCRIPT_RUNNER_IMAGE_REFERENCE to be an image ref, the policy check is to make sure that image ref appears in the SBOM. It was doing a direct string comparison, and because the sbom purl doesn't include the tag, the comparison was failing.

For example, "quay.io/foo/bar:latest@sha256:abc" compared to "quay.io/foo/bar@sha256:abc" in the SBOM.

After this change, those will be considered equal, and the pre_build_script_task.pre_build_script_task_runner_image_in_sbom violation will not be produced.

Note: It would be reasonable to not do this and suggest that the tag should be removed from the image ref before it gets added to the SCRIPT_RUNNER_IMAGE_REFERENCE, but I figured let's adhere to the princple of least surprise and make Conforma handle it the way someone might expect it to.

See also: https://issues.redhat.com/browse/EC-1242

Ref: https://issues.redhat.com/browse/KFLUXSPRT-3853

### DIFF
--- a/policy/release/pre_build_script_task/pre_build_script_task.rego
+++ b/policy/release/pre_build_script_task/pre_build_script_task.rego
@@ -130,10 +130,11 @@ _script_runner_image_refs := [image_ref |
 
 _pre_build_run_script_runner_image_result := "SCRIPT_RUNNER_IMAGE_REFERENCE"
 
-_is_image_in_sbom(image) if {
+_is_image_in_sbom(image_ref) if {
 	some s in sbom.all_sboms
 	some purl in _purls_from_sbom(s)
-	sbom.image_ref_from_purl(purl) == image
+	image_ref_from_purl := sbom.image_ref_from_purl(purl)
+	image.equal_ref(image_ref_from_purl, image_ref)
 }
 
 _purls_from_sbom(s) := purls if {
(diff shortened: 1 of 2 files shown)


policy/release/pre_build_script_task/pre_build_script_task.rego
```diff
@@ -130,10 +130,11 @@ _script_runner_image_refs := [image_ref |
 
 _pre_build_run_script_runner_image_result := "SCRIPT_RUNNER_IMAGE_REFERENCE"
 
-_is_image_in_sbom(image) if {
+_is_image_in_sbom(image_ref) if {
 	some s in sbom.all_sboms
 	some purl in _purls_from_sbom(s)
-	sbom.image_ref_from_purl(purl) == image
+	image_ref_from_purl := sbom.image_ref_from_purl(purl)
+	image.equal_ref(image_ref_from_purl, image_ref)
```
simonbaird, line 137: See https://github.com/conforma/policy/blob/536da0820f3262b348b3e23905d71b346d8a0d01/policy/lib/image/image.rego#L59-L83 .

dheerajodha (approved):

st3penta (approved):

PR #1437, opened 2025-07-02
feat(ci): support pushing bundles to multiple organizations

robnester-rh:
Update GitHub workflow and bundle push script to support pushing the same bundles to both enterprise-contract and conforma organizations. This enables a smooth transition between organizations while ensuring identical bundle content across both.

Changes:
- Modify workflow to pass repository prefixes as script arguments
- Update script to accept multiple repository prefixes via arguments
- Add support for organization-specific credentials
- Change default organization from enterprise-contract to conforma

Assisted-by: ChatGPT
Ref: EC-1184, EC-1185, EC-1187, EC-1189

### DIFF
--- a/hack/update-bundles.sh
+++ b/hack/update-bundles.sh
@@ -18,11 +18,32 @@
 # Pushes policy bundles to quay.io, but only if anything changed since
 # the last bundle was pushed.
 #
+# Usage: ./update-bundles.sh [repo-prefix1] [repo-prefix2] ...
+# Example: ./update-bundles.sh "quay.io/enterprise-contract/" "quay.io/conforma/"
+#
+# If no arguments provided, uses REPO_PREFIXES environment variable
+# If no arguments and no REPO_PREFIXES environment variable, defaults to "quay.io/conforma/"
+# Secrets must be provided via environment variables for security
+#
 set -o errexit
 set -o pipefail
 set -o nounset
 
-REPO_PREFIX="${REPO_PREFIX-quay.io/enterprise-contract/}"
+# Get repository prefixes from arguments or environment variable
+if [[ $# -gt 0 ]]; then
+  REPO_PREFIXES="$*"
+  echo "Using repository prefixes from arguments: $REPO_PREFIXES"
+else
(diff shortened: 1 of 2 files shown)


hack/update-bundles.sh
```diff
@@ -55,36 +91,53 @@ function cleanup() {
...
-    echo "Policy bundle $push_repo:$tag exists already, no push needed"
+  # Check all repositories to see if any need the bundle
+  repos_needing_push=()
+  all_repos_have_bundle=true
+
+  for repo_prefix in $REPO_PREFIXES; do
+    push_repo="${repo_prefix}$repo"
+    
+    # Login with correct credentials for tag checking
+    if [[ "$push_repo" == *"enterprise-contract"* ]]; then
```
joejstuart, line 110: Should the tag checking and pushing be added here also? Seems like it could be done all at once to avoid the duplicate login on push.

simonbaird (approved):
Some suggestions, but no blockers. Lgtm.

hack/update-bundles.sh
robnester-rh: So the tag checking determines if a given bundle needs to be built, that way it doesn't need to build all bundles every time.

hack/update-bundles.sh
```diff
@@ -111,12 +164,27 @@ for b in $BUNDLES; do
...
+    # Push to all repositories that need it
+    for push_repo in "${repos_needing_push[@]}"; do
+      echo "Pushing to $push_repo:$tag"
+      
+      # Login with the correct credentials for this repository
+      if [[ "$push_repo" == *"enterprise-contract"* ]]; then
+        echo "$EC_REGISTRY_PASS" | docker login quay.io --username "$EC_REGISTRY_USER" --password-stdin
+      elif [[ "$push_repo" == *"conforma"* ]]; then
+        echo "$CONFORMA_REGISTRY_PASS" | docker login quay.io --username "$CONFORMA_REGISTRY_USER" --password-stdin
+      fi
```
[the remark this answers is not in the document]
robnester-rh: ack. Updated.

PR #1441, opened 2025-07-02
Exclude test rego from bundles

simonbaird:
Benefits:
* Smaller bundles will be faster to download.
* Less work for the evaluator since it won't have to parse and compile all these files.
* Less work for the cli since it won't have to opa inspect all these test files when looking for `deny` and `warn` metadata.

Ref: https://issues.redhat.com/browse/EC-1350

### DIFF
--- a/hack/update-bundles.sh
+++ b/hack/update-bundles.sh
@@ -79,6 +79,7 @@ function bundle_subdir() {
 
 function exclusions() {
   echo "artifacthub-pkg.yml"
+  echo "*_test.rego"
 }
 
 function repo_name() {


simonbaird:
Draft because I think I should push one of these and make sure it works (rather than test it in production).

hack/update-bundles.sh
```diff
@@ -79,6 +79,7 @@ function bundle_subdir() {
 
 function exclusions() {
   echo "artifacthub-pkg.yml"
+  echo "*_test.rego"
```
simonbaird, line 82: See line 170.

simonbaird:
I pushed this to experiment with:

https://quay.io/repository/sbaird/release-policy?tab=tags

st3penta (approved):
Great idea!

PR #1447, opened 2025-07-08
Allow pre-build script from more places

simonbaird:


simonbaird:
See also #1448 . This is an alternative option, but I think I like #1448 better now.

PR #1449, opened 2025-07-10
Extend 'outdated task' message to non-ta pipelines

st3penta:
The violation description for the rule 'trusted_task.trusted' informs the user when the violation can be resolved by updating the task digest to a newer version.

Anyway, this helper description was only implemented for PipelineRuns that use Trusted Artifacts.

This commit extends it also to PipelineRuns that don't use Trusted Artifacts.

Ref: https://github.com/conforma/policy/pull/1394
Ref: https://issues.redhat.com/browse/EC-1269


simonbaird (approved):
Lgtm.

PR #1450, opened 2025-07-16
Use purl to determine base image registry

Allda:
A SBOM generator fixed some minor issues with the SBOM document that affect a base images. This was not compatible with conforma rule. The new implementation of the rule now depends on the based image purl and gets the image from there. The spdx type was correctly handled. The cyclonedx type was fixed in this commit.

JIRA: ISV-5866

This PR is related to changes in https://github.com/konflux-ci/build-definitions/pull/2532


simonbaird:
Nice, thank you! I think it's quite similar to #1406 which is the similar fix for SPDX format.

simonbaird:
Lgtm , but consider fixing my one nitpick.

simonbaird (approved):

simonbaird (approved):

Allda:
@simonbaird I don't have permission to merge it. Can you do it, please?

PR #1460, opened 2025-08-06
Make quay_expiration.expires_label rule stricter

st3penta:
This change completely removes the possibility of having the 'quay.expires-after' label in the released image, since it might cause unexpected behaviors when an image is pushed to a repository.

It also makes the error message clearer.

Ref: https://issues.redhat.com/browse/KONFLUX-9448


simonbaird (approved):
Nice.

PR #1461, opened 2025-08-06
Handle matrix tasks result digest lists correctly

simonbaird:
There's a check that requires that all test tasks verify that they did test the image being validated. To do this the tasks produce a task result that must contain the digest of the image.

Some tests tasks are now moving to run as a Tekton matrix task, so they split up and run as separate tasks for each arch being tested.

This means we end up with more than one task visible in the attestation, and hence one particular task result might not include the digest of specific image being tested, because it was actually tested in one of the other similar matrix tasks.

Work around the problem by doing some massaging of the task result list to combine the digest lists in the task results from similar matrix tasks into one combined digest list, and then referencing that in the policy check instead of the raw list.

Ref: https://issues.redhat.com/browse/KONFLUX-9576


simonbaird:
Draft because I think it needs some extra tests.

simonbaird:
Test coverage added (thanks Claude). Moving out of draft.

policy/release/test/test.rego
```diff
@@ -257,7 +257,8 @@ deny contains result if {
 	img := image.parse(input.image.ref)
 	img_digest := img.digest
 
-	some task in lib.images_processed_results_from_tests
```
arewm, line 260: Do we use this library anywhere else or can we remove it wholesale?

policy/release/lib/attestations.rego
```diff
@@ -124,6 +124,29 @@ results_from_tests := results_named(task_test_result_name)
 
 images_processed_results_from_tests := results_named(task_test_image_result_name)
```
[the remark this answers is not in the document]
simonbaird: Done.

policy/release/test/test.rego
simonbaird: I moved it inside the other rule, and also to the policy file.

robnester-rh (approved):
LGTM

PR #1506, opened 2025-08-28
feat(CLOUDDST-29067): Add OLM allowed resource kinds policy

asergienk:


policy/lib/rule_data.rego
```diff
@@ -89,6 +89,29 @@ rule_data_defaults := {
 		"registry.access.redhat.com/",
 		"registry.redhat.io/",
 	],
+	# Used in release/olm.rego
+	"allowed_olm_resource_kinds": [
```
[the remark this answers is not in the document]
yashvardhannanavati: It might be nice to link the source in the comments if we indeed have a single source of truth

policy/release/olm/olm.rego
```diff
@@ -358,6 +358,37 @@ deny contains result if {
...
+
+	result := lib.result_helper_with_term(rego.metadata.chain(), [manifest.kind], manifest.kind)
+}
+
+# Returns true if the kind is allowed
+allowed_kind_found(kind) := result if {
+	some allowed_kind in lib.rule_data("allowed_olm_resource_kinds")
+	allowed_kind == kind
+	result := true
+}
```
st3penta, line 390: you can simplify the syntax here and get rid of the result variable:
```suggestion
allowed_kind_found(kind) if {
    some allowed_kind in lib.rule_data("allowed_olm_resource_kinds")
	allowed_kind == kind
}
```

simonbaird: +1 to this suggestion

simonbaird: This might work also:

```suggestion

allowed_kind_found(kind) if {
	kind in lib.rule_data("allowed_olm_resource_kinds")
}
```

asergienk: Thanks for both of your suggestions! I was able to remove the helper function completely and replace it with a one-liner in the policy itself.

policy/lib/rule_data.rego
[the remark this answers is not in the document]
asergienk: There isn’t a single source of truth for this list right now. 

Allowed resources list was taken from [here](https://github.com/operator-framework/operator-registry/blob/3212635c5e3847e5e9f4b5db3f87b147405274fa/pkg/lib/bundle/supported_resources.go#L4-L23) but there is a NetworkPolicy still defined...so it is not fully correct.

There is also this [list](https://github.com/operator-framework/operator-lifecycle-manager/blob/9958a87b38ca5ce0269c0cbec0636ec4764a61dc/pkg/controller/operators/catalog/supportedresources.go#L4-L13) but it is not full.

The JIRA ticket exists to centralize it from multiple sources (tracked in [OPRUN-4067](https://issues.redhat.com/browse/OPRUN-4067)). Once that’s complete, we can link the single source of truth here.

asergienk:
@simonbaird could you please merge this PR?

PR #1704, opened 2026-04-03
Use image manifest to resolve tag-based SBOM refs

st3penta:
Tag-based references from ec.oci.image_tag_refs (e.g. repo:sha256-abc.sbom)
cannot be passed directly to ec.oci.blob, which only accepts digest refs.
This adds a lib.oci.blob_from_image helper that fetches the image manifest
and extracts blob content from its first layer, then updates
_sboms_from_tag_refs to use it.

This follows the existing pattern in cve.rego and aligns with reviewer
feedback on https://github.com/conforma/cli/pull/3214 (@lcarva, @simonbaird) to keep ec.oci.blob digest-only and
compose at the Rego level instead.

Ref: https://issues.redhat.com/browse/EC-1748

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/policy/lib/oci/oci.rego
+++ b/policy/lib/oci/oci.rego
@@ -1,0 +1,16 @@
+package lib.oci
+
+import data.lib.image
+import rego.v1
+
+# blob_from_image fetches the blob content of the first layer from an OCI
+# image manifest identified by ref. This is useful when ref is a tag-based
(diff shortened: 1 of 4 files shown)


simonbaird (approved):
Lgtm.

PR #1706, opened 2026-04-03
feat(EC-1722): add policy rule to verify hermetic builds enable Sonatype proxy

joejstuart:
## Summary
  - Adds `hermeto_proxy_enabled` rule to the `hermetic_task` package that denies when a hermetic build task does not have `enable-hermeto-proxy` set to `true`

  ## Jira
  https://redhat.atlassian.net/browse/EC-1722

### DIFF
--- a/policy/release/hermetic_task/hermetic_task.rego
+++ b/policy/release/hermetic_task/hermetic_task.rego
@@ -4,7 +4,8 @@
 # description: >-
 #   This package verifies that all the tasks in the attestation that
 #   are required to be hermetic were invoked with the proper
-#   parameters to perform a hermetic execution.
+#   parameters to perform a hermetic execution, including enabling
+#   the Sonatype proxy when required.
 #
 package hermetic_task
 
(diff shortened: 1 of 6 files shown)


robnester-rh:
@coderabbitai review

policy/release/hermetic_task/hermetic_task.rego
```diff
@@ -55,20 +56,58 @@ deny contains result if {
...
+	not _task_is_hermetic(task)
+}
+
+_hermetic_tasks_without_proxy contains task if {
+	some task in _required_hermetic_build_tasks
+	_task_is_hermetic(task)
+	not _task_has_proxy_enabled(task)
+}
+
+_required_hermetic_build_tasks contains task if {
```
simonbaird: Given the discussion with code rabbit, seems this would be better if it didn't have "build" in its name..?


```suggestion
_required_hermetic_tasks contains task if {
```

policy/release/hermetic_task/hermetic_task.rego
```diff
@@ -55,20 +56,58 @@ deny contains result if {
...
 }
 
 _task_is_hermetic(task) if {
 	tekton.task_param(task, "HERMETIC")
 	tekton.task_param(task, "HERMETIC") == "true"
 }
 
+_task_has_proxy_enabled(task) if {
+	tekton.task_param(task, "enable-hermeto-proxy")
+	tekton.task_param(task, "enable-hermeto-proxy") == "true"
```
simonbaird: I don't mind keeping it for readability if you prefer, but I think it works the same without line 107.


```suggestion
	tekton.task_param(task, "enable-hermeto-proxy") == "true"
```

simonbaird (dismissed):
Lgtm but consider my suggestion.

joejstuart:
@simonbaird @robnester-rh I should have all comments resolved. PTAL again.

robnester-rh (dismissed):
Looks good. Clean refactoring of `_required_hermetic_tasks` as a shared comprehension, tests are thorough (both SLSA v0.2 and v1.0, proxy present/missing/false, non-hermetic tasks correctly excluded from proxy check), and the `_task_has_proxy_enabled` is now consistent with `_task_is_hermetic` style after Simon's feedback.

One minor question (non-blocking): the rule applies to all `required_hermetic_tasks` unconditionally. If there are ever hermetic tasks that don't need the Sonatype proxy (e.g., non-language builds), this would need scoping via separate rule data. But for the `redhat` collection today this seems fine.

robnester-rh (approved):
LGTM

simonbaird (approved):

PR #1708, opened 2026-04-03
feat(EC-1725): Validate proxy URLs against approved patterns in SBOMs

joejstuart:
## Summary
  - Adds `allowed_proxy_urls` rule to `sbom_cyclonedx` — checks `externalReferences` URLs where type is `distribution` against `allowed_proxy_url_patterns` for
  PURL types listed in `proxy_enabled_purl_types`
  - Adds `allowed_proxy_urls` rule to `sbom_spdx` — checks `downloadLocation` against the same patterns (skips `NOASSERTION`)
  - Only PURL types in `proxy_enabled_purl_types` are validated; all others are ignored

  **Note:** This branch includes merge commits from EC-1722 and EC-1724 as it depends on both.

  ## Jira
  https://redhat.atlassian.net/browse/EC-1725


simonbaird:
I think the commit titled `feat(EC-1724): add rule_data configuration for Sonatype proxy enforcement` got shrunk down to one docs line number change. So maybe clean up the history so it does what it says, or squash it away entirely maybe.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -266,6 +266,47 @@ deny contains result if {
...
+#   failure_msg: >-
+#     Package %s has proxy URL %q which does not match any allowed pattern for PURL type %q
+#   solution: >-
+#     Ensure the proxy URL matches one of the patterns defined in the
+#     allowed_proxy_url_patterns rule data for the given PURL type.
+#   collections:
+#   - redhat
+#
+deny contains result if {
+	proxy_enabled := {t | some t in rule_data.get("proxy_enabled_purl_types")}
```
simonbaird, line 296: 1. Is there a special reason to convert this to a set, but not the other list from rule data on the next line?
2. Maybe `proxy_enabled_purl_types` or `enabled_purl_types` is easier to read and follow. Something that makes it more clear that it's a list (or set).

policy/release/sbom_spdx/sbom_spdx.rego
```diff
@@ -252,6 +252,48 @@ deny contains result if {
...
 
+# METADATA
+# title: Allowed proxy URLs
+# description: >-
+#   For packages with a PURL type listed in proxy_enabled_purl_types, verify the
+#   downloadLocation matches at least one pattern from allowed_proxy_url_patterns.
+# custom:
+#   short_name: allowed_proxy_urls
+#   failure_msg: >-
+#     Package %s has proxy URL %q which does not match any allowed pattern for PURL type %q
```
simonbaird, line 270: It seems weird to create the same check twice, once for cyclonedx and once for spdx. Could we perhaps refactor so there's a package called sbom_purl_proxies (or something like that) that defines the "deny" just once in a way that it can sees purls from both kinds of sbom formats?

simonbaird (commented):
See my comment about not having (essentially) the same logic defined twice, once for each format of sbom.

policy/release/sbom_spdx/sbom_spdx.rego
joejstuart: It's the same rule name/check, but different logic. Both are specific to the kind of `sbom`. It's the same pattern as `disallowed_package_external_references` and `allowed_package_sources`.

antora/docs/modules/ROOT/pages/packages/release_hermetic_task.adoc
```diff
@@ -1,13 +1,25 @@
...
+=== link:#hermetic_task__hermeto_proxy_enabled[Hermetic build task has Sonatype proxy enabled]
+
+Verify that hermetic build tasks have the enable-hermeto-proxy parameter set to true. This ensures that hermetic builds use the Sonatype proxy for dependency resolution.
+
+*Solution*: Make sure the task has the input parameter 'enable-hermeto-proxy' set to 'true'.
+
+* Rule type: [rule-type-indicator failure]#FAILURE#
+* FAILURE message: `Task '%s' is hermetic but does not have the enable-hermeto-proxy parameter set to true`
+* Code: `hermetic_task.hermeto_proxy_enabled`
+* https://github.com/conforma/policy/blob/{page-origin-refhash}/policy/release/hermetic_task/hermetic_task.rego#L42[Source, window="_blank"]
```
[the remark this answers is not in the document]
joejstuart: Not applicable — the source links in these docs are auto-generated by `make generate-docs` and consistently point to the `# METADATA` annotation block, not the `deny` rule body. This is the convention across all generated doc pages (e.g., `release_sbom_spdx.adoc` links to L52, L75, etc., which are all `# METADATA` lines). Manually changing the anchor would be overwritten on the next doc regeneration.

antora/docs/modules/ROOT/pages/packages/release_hermetic_task.adoc
```diff
@@ -18,7 +30,7 @@ Verify the task in the PipelineRun attestation was invoked with the proper param
 * Rule type: [rule-type-indicator failure]#FAILURE#
 * FAILURE message: `Task '%s' was not invoked with the hermetic parameter set`
 * Code: `hermetic_task.hermetic`
-* https://github.com/conforma/policy/blob/{page-origin-refhash}/policy/release/hermetic_task/hermetic_task.rego#L19[Source, window="_blank"]
+* https://github.com/conforma/policy/blob/{page-origin-refhash}/policy/release/hermetic_task/hermetic_task.rego#L20[Source, window="_blank"]
```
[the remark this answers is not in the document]
joejstuart: Same as above — these source links are auto-generated and consistently point to `# METADATA` blocks across all doc pages. This is by design, not a bug.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
joejstuart: `proxy_enabled` is converted to a set because it's used with the in operator (parsed_purl.type in proxy_enabled), which is O(1) lookup on a set vs O(n) on an array.

policy/release/hermetic_task/hermetic_task.rego
```diff
@@ -38,6 +39,29 @@ deny contains result if {
...
+#   the Sonatype proxy for dependency resolution.
+# custom:
+#   short_name: hermeto_proxy_enabled
+#   failure_msg: >-
+#     Task '%s' is hermetic but does not have the enable-hermeto-proxy parameter set to true
+#   solution: >-
+#     Make sure the task has the input parameter 'enable-hermeto-proxy'
+#     set to 'true'.
+#   collections:
+#   - redhat
```
st3penta, line 55: **issue:** Missing `effective_on` grace period

The new `hermeto_proxy_enabled` rule has no `effective_on` date, unlike comparable rules in the codebase. For reference, `allowed_package_sources` has `effective_on: 2024-12-15` (CycloneDX) and `effective_on: 2025-02-17` (SPDX).

This rule fires immediately for all hermetic tasks, with no gate on whether proxy patterns are even configured. Existing builds that predate the Sonatype proxy feature will start failing as soon as this merges.

This creates pressure for teams to either: (a) add `enable-hermeto-proxy=true` to tasks that may not support it yet, (b) remove tasks from `required_hermetic_tasks`, or (c) exclude the entire `hermetic_task` package — which would also disable the existing `hermetic` check, weakening the overall security posture.

The SBOM-level `allowed_proxy_urls` rules are partially mitigated by `proxy_enabled_purl_types` defaulting to empty (rule is a no-op until configured), but adding `effective_on` would be consistent with the existing pattern.

**Recommendation**: Add an `effective_on` date (30–60 days out) to at least this rule, consistent with `allowed_package_sources` precedent.

policy/release/hermetic_task/hermetic_task.rego
```diff
@@ -38,6 +39,29 @@ deny contains result if {
...
+#   Verify that hermetic build tasks have the enable-hermeto-proxy
+#   parameter set to true. This ensures that hermetic builds use
+#   the Sonatype proxy for dependency resolution.
+# custom:
+#   short_name: hermeto_proxy_enabled
+#   failure_msg: >-
+#     Task '%s' is hermetic but does not have the enable-hermeto-proxy parameter set to true
+#   solution: >-
+#     Make sure the task has the input parameter 'enable-hermeto-proxy'
+#     set to 'true'.
```
st3penta, line 53: **suggestion:** Document the dependency on SBOM-level configuration

The task-level `hermeto_proxy_enabled` check and the SBOM-level `allowed_proxy_urls` check are independently configured. An operator who enables the task-level proxy flag but forgets to configure `proxy_enabled_purl_types` gets a false sense of security — the task says "proxy is enabled" but SBOM URLs are never actually validated because the SBOM-level rule data was never populated.

Consider either:
- Adding a note in the solution text mentioning this dependency (e.g., "Also ensure `proxy_enabled_purl_types` and `allowed_proxy_url_patterns` are configured in rule data to enable SBOM-level proxy URL validation.")
- Emitting a warning-severity result when `enable-hermeto-proxy=true` is present on a task but `proxy_enabled_purl_types` is empty.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -266,6 +266,50 @@ deny contains result if {
...
+#   failure_msg: >-
+#     Package %s has proxy URL %q which does not match any allowed pattern for PURL type %q
+#   solution: >-
+#     Ensure the proxy URL matches one of the patterns defined in the
+#     allowed_proxy_url_patterns rule data for the given PURL type.
+#   collections:
+#   - redhat
+#   - redhat_rpms
+#   - policy_data
+#
```
st3penta, line 294: **suggestion:** Three METADATA improvements for this rule

**1. Scope clarification**: Unlike `allowed_package_sources` (which filters by `hermeto:found_by` / `cachi2:found_by`), this rule intentionally checks ALL components with proxy-enabled PURL types. This broader scope is the right design (see separate praise comment), but it will be a recurring reviewer question. Consider extending the description:
```
#   Unlike allowed_package_sources, this rule intentionally checks all components
#   regardless of fetcher (hermeto/cachi2) to ensure all distribution URLs for
#   proxy-enabled types route through the approved proxy.
```

**2. Failure message terminology** (L278): The `failure_msg` says "proxy URL" but the field being checked is actually a distribution URL (`externalReferences[type=distribution].url`) that *should* be a proxy URL. An operator seeing a non-proxy URL in the output may be confused by it being called a "proxy URL". Consider: `has distribution URL %q which does not match any allowed proxy pattern for PURL type %q`.

**3. Cross-reference with `allowed_package_sources`** (solution text): For CycloneDX, a Hermeto-fetched component with a proxy-enabled PURL type has its distribution URL checked by both `allowed_package_sources` and this rule independently. An operator who configures proxy patterns correctly but forgets to also allow the proxy URL in `allowed_package_sources` gets a confusing `sbom_cyclonedx.allowed_package_sources` failure. Consider adding to the solution text:
```
#     Note: for Hermeto-fetched components, the same distribution URL is also
#     checked by allowed_package_sources — ensure both pattern sets are
#     configured consistently.
```
And add a matching note in the existing `allowed_package_sources` solution text pointing the other direction.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -266,6 +266,50 @@ deny contains result if {
...
+deny contains result if {
+	proxy_enabled := {t | some t in rule_data.get("proxy_enabled_purl_types")}
+	allowed_patterns := rule_data.get("allowed_proxy_url_patterns")
```
st3penta, line 297: **suggestion:** Two validation gaps for the rule data keys consumed here

**1. Independent validation**: `proxy_enabled_purl_types` and `allowed_proxy_url_patterns` are consumed here without validation — schema validation for these keys lives only in `hermetic_task.rego`'s `_rule_data_errors`. If `hermetic_task` is ever excluded from evaluation, these SBOM rules operate on unvalidated configuration. Consider adding rule data validation for these keys to the sbom lib (`policy/lib/sbom/sbom.rego`), or document the dependency explicitly.

**2. Missing patterns cross-validation**: When a PURL type is in `proxy_enabled_purl_types` but has no entry in `allowed_proxy_url_patterns`, `object.get` at L303 defaults to `[]`, causing every component of that type to be blanket-denied. The violation message says the URL doesn't match any pattern, but the real problem is no patterns were configured — the operator gets no signal of the root cause. Consider adding a cross-validation rule in `_rule_data_errors` that warns when an enabled PURL type has zero patterns:

```rego
_rule_data_errors contains error if {
    proxy_types := rule_data.get("proxy_enabled_purl_types")
    patterns := rule_data.get("allowed_proxy_url_patterns")
    some purl_type in proxy_types
    type_patterns := object.get(patterns, purl_type, [])
    count(type_patterns) == 0
    error := sprintf("PURL type %q is in proxy_enabled_purl_types but has no patterns in allowed_proxy_url_patterns", [purl_type])
}
```

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -266,6 +266,50 @@ deny contains result if {
...
+	reference.type == "distribution"
+
+	purl := component.purl
+	parsed_purl := ec.purl.parse(purl)
+	parsed_purl.type in proxy_enabled
+
+	distribution_url := object.get(reference, "url", "")
+	distribution_url != "NOASSERTION"
+	patterns := object.get(allowed_patterns, parsed_purl.type, [])
+	not sbom.url_matches_any_pattern(distribution_url, patterns)
```
st3penta, line 312: **suggestion:** Three edge cases to consider hardening

**1. Invalid/missing PURL** (L298): `ec.purl.parse` on a malformed PURL returns undefined, silently skipping the component — fail-open on a security path. PURL validity is not enforced elsewhere for CycloneDX components; there are no deny rules flagging missing or malformed PURLs. This is a pre-existing codebase gap affecting all PURL-dependent rules, not just this PR. Worth considering a follow-up issue to add a deny rule flagging SBOM components with missing/invalid PURLs.

**2. Empty distribution URL** (L301): If a distribution reference has no `url` field, the fallback to `""` triggers a violation with `proxy URL ""`. If there's no URL, there's nothing to validate. Consider adding an empty-string guard:
```rego
distribution_url := object.get(reference, "url", "")
distribution_url != ""
distribution_url != "NOASSERTION"
```

**3. Regex anchoring** (L304): OPA's `regex.match` does substring matching by default. An unanchored pattern like `https://proxy\.example\.com/maven/.*` could theoretically be bypassed by embedding the expected hostname in a malicious URL (e.g., `https://evil.com/https://proxy.example.com/maven/exploit.jar`). The test data correctly uses `^` anchoring, but the policy doesn't enforce it. Consider either programmatically ensuring patterns are anchored in `url_matches_any_pattern`, adding validation that warns on unanchored patterns, or at minimum documenting the anchoring requirement in the rule's solution text.

These observations apply equally to the SPDX `allowed_proxy_urls` rule.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -266,6 +266,50 @@ deny contains result if {
...
+#     Package %s has proxy URL %q which does not match any allowed pattern for PURL type %q
+#   solution: >-
+#     Ensure the proxy URL matches one of the patterns defined in the
+#     allowed_proxy_url_patterns rule data for the given PURL type.
+#   collections:
+#   - redhat
+#   - redhat_rpms
+#   - policy_data
+#
+deny contains result if {
```
st3penta, line 295: **praise:** The decision to check ALL components with proxy-enabled PURL types — not just Hermeto-fetched ones — is the right call for a proxy enforcement rule. It provides stronger supply chain guarantees by preventing any component from claiming a non-proxy distribution URL, regardless of which fetcher was used.

policy/release/sbom_spdx/sbom_spdx.rego
```diff
@@ -252,6 +252,51 @@ deny contains result if {
...
+	some pkg in s.packages
+
+	some externalref in pkg.externalRefs
+	externalref.referenceType == "purl"
+
+	purl := externalref.referenceLocator
+	parsed_purl := ec.purl.parse(purl)
+	parsed_purl.type in proxy_enabled
+
+	download_location := object.get(pkg, "downloadLocation", "")
```
st3penta, line 294: **suggestion:** Document the data-source choice

This rule checks `downloadLocation` (SPDX package field), while the adjacent `allowed_package_sources` checks the PURL `download_url` qualifier. These are different data sources for semantically similar concepts. A one-line inline comment would prevent maintainer confusion:

```rego
# Check downloadLocation (SPDX package field), not the PURL download_url qualifier.
# The proxy URL is the declared SPDX download location, not the Hermeto-resolved source.
download_location := object.get(pkg, "downloadLocation", "")
```

policy/release/sbom_cyclonedx/sbom_cyclonedx_test.rego
st3penta, line 1: **suggestion:** Three missing test equivalence classes for proxy URL checking

1. **Multiple distribution references**: No test exercises a component with two `externalReferences` of type `distribution` where one URL passes and one fails. The rule iterates `some reference in component.externalReferences`, so this is a real edge case — add a test confirming exactly one violation is produced.

2. **Empty `proxy_enabled_purl_types`**: The rollout strategy depends on the rule being a no-op when `proxy_enabled_purl_types` is empty, yet every proxy test provides `["maven", "npm"]`. Add a test with `proxy_enabled_purl_types: []` and a component with a non-matching URL, asserting no violations.

3. **PURL type enabled but no patterns configured**: When a PURL type is in `proxy_enabled_purl_types` but has no entry in `allowed_proxy_url_patterns`, `object.get` defaults to `[]` and ALL components of that type get blanket-denied. This plausible misconfiguration should be documented via a test.

policy/release/sbom_spdx/sbom_spdx_test.rego
st3penta, line 1: **suggestion:** Two missing test equivalence classes for SPDX proxy URL checking

1. **Non-proxy PURL type is ignored**: CycloneDX has `test_proxy_url_cyclonedx_non_proxy_purl_type` but SPDX has no equivalent. The `parsed_purl.type in proxy_enabled` branch is uncovered in the SPDX code path. Add a `test_proxy_url_spdx_non_proxy_purl_type` with a `pkg:golang/...` package and a non-matching URL, asserting no violations.

2. **Missing `downloadLocation`**: If `downloadLocation` is absent, `object.get` defaults to `""`, which isn't `"NOASSERTION"`, so the rule fires a violation. This edge case is worth documenting with a test.

policy/release/hermetic_task/hermetic_task_test.rego
st3penta, line 1: **nitpick:** Consider adding a test for case-variant `enable-hermeto-proxy` values

The `_task_has_proxy_enabled` helper checks `== "true"` (lowercase). A test case with `"TRUE"` would document that the check is case-sensitive, consistent with the existing `_task_is_hermetic` testing pattern in `test_task_is_hermetic`.

antora/docs/modules/ROOT/pages/packages/release_sbom_cyclonedx.adoc
```diff
@@ -45,6 +45,18 @@ For each of the components fetched by Hermeto which define externalReferences of
 * Effective from: `2024-12-15T00:00:00Z`
 * https://github.com/conforma/policy/blob/{page-origin-refhash}/policy/release/sbom_cyclonedx/sbom_cyclonedx.rego#L225[Source, window="_blank"]
 
+[#sbom_cyclonedx__allowed_proxy_urls]
+=== link:#sbom_cyclonedx__allowed_proxy_urls[Allowed proxy URLs]
+
+For components with externalReferences of type distribution, verify proxy URLs match at least one pattern from allowed_proxy_url_patterns for the component's PURL type. Only PURL types listed in proxy_enabled_purl_types are checked.
+
+*Solution*: Ensure the proxy URL matches one of the patterns defined in the allowed_proxy_url_patterns rule data for the given PURL type.
```
st3penta, line 53: **suggestion:** Document the expected format of `proxy_enabled_purl_types` and `allowed_proxy_url_patterns`

The `allowed_proxy_urls` rule consumes two rule data keys, but the documentation here doesn't describe their expected format. An operator reading this page has no way to know that `proxy_enabled_purl_types` is an array of strings and `allowed_proxy_url_patterns` is an object mapping PURL types to arrays of regex patterns. The validation logic lives only in `hermetic_task.proxy_rule_data_format`.

Consider adding a brief note describing the expected format, or cross-referencing the `hermetic_task.proxy_rule_data_format` rule. Same applies to the SPDX equivalent page.

acceptance/samples/policy-input-golden-container.json
```diff
@@ -476,6 +476,7 @@
                     "HERMETIC": "true",
                     "IMAGE": "quay.io/redhat-user-workloads/rhtap-contract-tenant/golden-container/golden-container:d11c0ada39f18f631ff4f54beafa72a9b62dcd7c-amd64",
                     "IMAGE_EXPIRES_AFTER": "",
+                    "enable-hermeto-proxy": "true",
```
st3penta, line 479: **nitpick:** Consider adding acceptance test assertions for the new rules

The golden container sample was updated with `enable-hermeto-proxy: true` to prevent regressions, which is good. Consider adding acceptance test assertions that exercise the new `hermeto_proxy_enabled` and `allowed_proxy_urls` rules against this sample to ensure they work end-to-end.

st3penta (changes requested):
## Summary

This PR adds three new Rego policy rules to enforce Sonatype proxy usage in hermetic builds, covering both the task-level (`hermeto_proxy_enabled`) and SBOM-level checks (CycloneDX and SPDX `allowed_proxy_urls`). The architecture is sound: the safe-default behavior (rules are no-op when `proxy_enabled_purl_types` is empty) enables incremental rollout, and the decision to check all components with proxy-enabled PURL types — not just Hermeto-fetched ones — provides stronger supply chain guarantees.

The refactoring of `_not_hermetic_tasks` into `_required_hermetic_build_tasks` plus derived sets is clean and well-structured. Test coverage for the new rules is solid, and the Antora documentation was properly updated.

## Verdict

**Request Changes** — The main concern is the missing `effective_on` grace period on the `hermeto_proxy_enabled` rule, which could cause immediate failures for existing hermetic builds that predate the Sonatype proxy feature. This is a deviation from the established convention used by comparable rules like `allowed_package_sources`. The SBOM-level rules are partially mitigated by `proxy_enabled_purl_types` defaulting to empty, but adding `effective_on` to those as well would be consistent with precedent.

Additionally, please ensure these three new enforcement rules are called out in the next release notes — operators need to know about new rules that will affect their builds when enabled.

## Review Breakdown

- **1 issue** (blocking): missing `effective_on` on `hermeto_proxy_enabled`
- **8 suggestions**: cross-validation dependency, METADATA improvements, validation gaps, edge case hardening, data-source documentation, test coverage gaps, Antora docs
- **2 nitpicks**: case-variant test, acceptance test assertions
- **1 praise**: defense-in-depth scope on SBOM proxy rules

policy/release/sbom_spdx/sbom_spdx.rego
st3penta: +1 on the refactoring suggestion. This PR's own history demonstrates the duplication risk: the CycloneDX NOASSERTION guard was initially missed while SPDX had it from the start (fixed in `b4bc54d`). The `allowed_package_sources` precedent cuts both ways — it established the pattern, but it also established the duplication risk. The thread should be resolved with an explicit decision, even if the decision is to defer the refactoring.

joejstuart: The pattern is the rules pertaining to `spdx` go in `sbom_spdx.rego` and the `cyclonedx` rules go in `sbom_cyclonedx.rego`. Two rules may be checking for the same thing, but they have to go about it differently for each SBOM format.

policy/release/hermetic_task/hermetic_task.rego
joejstuart: I'm not sure this is relevant. I don't see the dependency. You can care if the proxy is enabled and also care if hermeto used the right urls independently.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
joejstuart: This was a good catch.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
joejstuart: This one makes my head hurt and doesn't seem meaningful.

policy/release/sbom_spdx/sbom_spdx.rego
joejstuart: This doesn't seem very helpful.

policy/release/sbom_cyclonedx/sbom_cyclonedx_test.rego
joejstuart: These are good

policy/release/sbom_spdx/sbom_spdx_test.rego
joejstuart: These are good

policy/release/hermetic_task/hermetic_task_test.rego
joejstuart: I guess, sure.

antora/docs/modules/ROOT/pages/packages/release_sbom_cyclonedx.adoc
joejstuart: this is good

acceptance/samples/policy-input-golden-container.json
joejstuart: It should already be covered in `Scenario: Red Hat collection`

policy/release/hermetic_task/hermetic_task.rego
joejstuart: This isn't set, because I thought it was supposed to go out right away.

policy/release/sbom_cyclonedx/sbom_cyclonedx_test.rego
```diff
@@ -658,3 +658,169 @@ _sbom_1_6_attestation := json.patch(_sbom_1_5_attestation, [
...
+			"NOASSERTION",
+		),
+	}])
+
+	assertions.assert_empty(sbom_cyclonedx.deny) with input.attestations as [att]
+		with input.image.ref as "registry.local/spam@sha256:1230000000000000000000000000000000000000000000000000000000000123"
+		with ec.oci.image_referrers as []
+		with ec.oci.image_tag_refs as []
+		with data.rule_data as _proxy_rule_data
+}
```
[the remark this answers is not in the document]
joejstuart: The NOASSERTION skip is intentional and correct for CycloneDX. A few points:

1. **It's documented in the rule's own METADATA** (line 279): *"Components with a URL of 'NOASSERTION' are skipped."* This was a deliberate design choice, not an oversight.

2. **NOASSERTION is a standard SBOM sentinel value** meaning "no information available" — it's not a real URL. Denying on it would produce false positives for components that legitimately have no distribution URL.

3. **Both formats handle it identically** — the SPDX rule has the same `!= "NOASSERTION"` guard (line 293). The claim that this skip is "called out for SPDX, not CycloneDX" is incorrect; the CycloneDX rule description explicitly documents it.

4. **It's not a bypass** — the rule is a `deny` rule, so the guard simply means components with no real URL don't produce a denial. An attacker can't exploit this because `NOASSERTION` isn't a valid proxy URL that would actually fetch dependencies.

simonbaird (dismissed):
Lgtm, but I didn't follow some of the discussion.

policy/release/sbom_spdx/sbom_spdx.rego
```diff
@@ -252,6 +252,56 @@ deny contains result if {
...
+#   packages of that type are denied.
+# custom:
+#   short_name: allowed_proxy_urls
+#   failure_msg: >-
+#     Package %s has proxy URL %q which does not match any allowed pattern for PURL type %q
+#   solution: >-
+#     Ensure the proxy URL matches one of the patterns defined in the
+#     allowed_proxy_url_patterns rule data for the given PURL type.
+#   collections:
+#   - redhat
```
st3penta, line 275: I think you lost some changes here (you had this in this commit: https://github.com/conforma/policy/commit/aec79a103ece1038128c5d7ef463f8a6ff6f6b88)

```suggestion
#   collections:
#   - redhat
#   - redhat_rpms
#   - policy_data
```

st3penta (commented):
Looks good now, just one nitpick about rule collections

policy/release/sbom_spdx/sbom_spdx.rego
joejstuart: Oh good catch

st3penta (dismissed):

st3penta (approved):

PR #1710, opened 2026-04-08
Add backwards compatibility shim for lib.result_helper functions

dheerajodha:
Re-export result_helper functions at their original data.lib.* paths to maintain compatibility with external demos, documentation, and integrations that reference the old locations after reorganization to data.lib.metadata.*.

Provides deprecated aliases for:
- data.lib.result_helper
- data.lib.result_helper_with_term
- data.lib.result_helper_with_severity

New code should import from data.lib.metadata.* directly.


simonbaird (approved):
I like it.

PR #1714, opened 2026-04-20
Extend CVE policy to accept TPA scan reports

Acepresso:
The TPA CVE scan report format is similar to the Clair report, so I just added the TPA media type (application/vnd.redhat.tpa-report+json) alongside the Clair media type.

Ref: https://redhat.atlassian.net/browse/EC-1770


gbenhaim:
I've tried to use this PR and now it fails on another issue `CVE scan results were not found`

simonbaird (approved):

joejstuart (approved):

PR #1715, opened 2026-04-20
Trusted task refactor

joejstuart:
When multiple data sources each define trusted_task_rules entries, OPA fails to merge them because arrays at the same path produce a merge error at data load time, before Rego evaluation begins. This change switches allow and deny from arrays to objects keyed by a unique name, allowing OPA to deep-merge entries from multiple data sources without conflict.

https://redhat.atlassian.net/browse/EC-1782

### DIFF
--- a/policy/lib/tekton/trusted.rego
+++ b/policy/lib/tekton/trusted.rego
@@ -142,34 +142,86 @@ _trusted_task_rules_data := {
 	),
 }
 
-# Safely extract allow from data.trusted_task_rules
+# Safely extract allow from data.trusted_task_rules.
+# Supports both array and object formats.
 default _data_allow_array := []
 
 _data_allow_array := data.trusted_task_rules.allow if {
 	is_object(data.trusted_task_rules)
+	is_array(data.trusted_task_rules.allow)
 }
 
-# Safely extract deny from data.trusted_task_rules
+_data_allow_array := [entry |
+	is_object(data.trusted_task_rules)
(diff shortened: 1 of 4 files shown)


joejstuart:
Depends on this - https://github.com/konflux-ci/architecture/pull/336

policy/lib/tekton/trusted.rego
```diff
@@ -142,34 +142,60 @@ _trusted_task_rules_data := {
...
 
-_data_allow_array := data.trusted_task_rules.allow if {
+_data_allow_array := [rule |
 	is_object(data.trusted_task_rules)
+	is_object(data.trusted_task_rules.allow)
+	some _, rules in data.trusted_task_rules.allow # regal ignore:in-wildcard-key
+	some rule in rules
+] if {
+	is_object(data.trusted_task_rules)
+	is_object(data.trusted_task_rules.allow)
```
simonbaird, line 154: With these two lines here, are the line 150 and 151 still needed? Seems like it could be repeating the logic needlessly? (Same thing for the deny array).

simonbaird (dismissed):
Lgtm, but see my question about possibly duplicated logic.

simonbaird:
Actually Coderabbit mentioned the same thing and also another nitpick that might be worth fixing.

policy/lib/tekton/trusted.rego
joejstuart: Removed

simonbaird (approved):

PR #1718, opened 2026-04-24
fix(olm): use ec.oci.descriptor to check unmapped image accessibility

joejstuart:
`ec.oci.image_manifest` resolves multi-arch indexes to linux/amd64, failing for indexes that only contain other platforms. Use `ec.oci.descriptor` instead, which performs a HEAD request to verify the image exists without platform resolution.

https://redhat.atlassian.net/browse/EC-1781


robnester-rh (approved):
LGTM

simonbaird (approved):

PR #1719, opened 2026-04-30
feat: verify operator bundle operands use NetworkPolicies for new releases

yashvardhannanavati:
Network Policies have been supported for Operand images since v4.6. However, there’s no check anywhere to enforce it.

The ability of an operand to manage network policies is identified by the presence of necessary RBAC permissions in the operator bundle CSV. A warning for missing RBAC will be issued initially, with a plan to transition this to a release-blocking error later.

What it checks:
The rule inspects the operator bundle's ClusterServiceVersion (CSV) for the presence of explicit RBAC permissions targeting networking.k8s.io/networkpolicies with at least one of the verbs create, update, patch, or delete — in either clusterPermissions or permissions.

Exception handling:
Before enforcing the check, the rule looks up the operator's package name and major.minor version (e.g., "3scale-operator" / "0.10") in the operator_network_policy_rbac_exceptions rule_data map. If the version is found there, the check is skipped entirely for that bundle.

 

Refers to CLOUDDST-32507

Assisted-By: Claude

### DIFF
--- a/antora/docs/modules/ROOT/pages/packages/release_olm.adoc
+++ b/antora/docs/modules/ROOT/pages/packages/release_olm.adoc
@@ -83,6 +83,19 @@ Each image indicated as a related image should match an entry in the list of pre
 * Effective from: `2025-04-15T00:00:00Z`
 * https://github.com/conforma/policy/blob/{page-origin-refhash}/policy/release/olm/olm.rego#L232[Source, window="_blank"]
 
+[#olm__required_network_policy_rbac_for_operands]
+=== link:#olm__required_network_policy_rbac_for_operands[Required NetworkPolicy RBAC missing from OLM bundle]
+
+Operators are required to manage the network policies of their operands. This rule verifies that operator bundles request sufficient RBAC permissions to manage NetworkPolicy lifecycle (create, delete, and update/patch) for networking.k8s.io/networkpolicies in their ClusterServiceVersion. Bundles whose operator name and major.minor version are listed in the `operator_network_policy_rbac_exceptions` rule data key are exempt from this requirement.
+
+*Solution*: Add a rule granting create, delete, and update/patch on networking.k8s.io/networkpolicies to the ClusterServiceVersion clusterPermissions or permissions, or add the operator name and its major.minor version to the operator_network_policy_rbac_exceptions rule data key.
+
+* Rule type: [rule-type-indicator failure]#FAILURE#
+* FAILURE message: `Operator %q version %q is missing required NetworkPolicy RBAC (networking.k8s.io/networkpolicies with create, delete, and update/patch)`
+* Code: `olm.required_network_policy_rbac_for_operands`
+* Effective from: `2026-08-07T00:00:00Z`
+* https://github.com/conforma/policy/blob/{page-origin-refhash}/policy/release/olm/olm.rego#L387[Source, window="_blank"]
+
 [#olm__required_olm_features_annotations_provided]
 === link:#olm__required_olm_features_annotations_provided[Required OLM feature annotations list provided]
 
--- a/policy/lib/rule_data/rule_data.rego
+++ b/policy/lib/rule_data/rule_data.rego
@@ -117,6 +117,9 @@ defaults := {
 		"ConsoleLink",
 		"ConsolePlugin",
 	],
+	# Used in release/olm.rego
+	# Operators excepted from NetworkPolicy RBAC requirement by name and major.minor version
+	"operator_network_policy_rbac_exceptions": {},
 	#
 	# Used in release/hermetic_task/hermetic_task.rego
 	"required_hermetic_tasks": [
--- a/policy/release/olm/olm.rego
+++ b/policy/release/olm/olm.rego
@@ -384,6 +384,85 @@ deny contains result if {
 	result := metadata.result_helper_with_term(rego.metadata.chain(), [manifest.kind], manifest.kind)
 }
 
+# METADATA
+# title: Required NetworkPolicy RBAC missing from OLM bundle
+# description: >-
+#   Operators are required to manage the network policies of their operands.
+#   This rule verifies that operator bundles request sufficient RBAC permissions
+#   to manage NetworkPolicy lifecycle (create, delete, and update/patch) for
+#   networking.k8s.io/networkpolicies in their ClusterServiceVersion.
+#   Bundles whose operator name and major.minor version are listed in the
+#   `operator_network_policy_rbac_exceptions` rule data key are exempt from this
+#   requirement.
+# custom:
+#   short_name: required_network_policy_rbac_for_operands
+#   failure_msg: >-
+#     Operator %q version %q is missing required NetworkPolicy RBAC
+#     (networking.k8s.io/networkpolicies with create, delete, and update/patch)
+#   solution: >-
+#     Add a rule granting create, delete, and update/patch on networking.k8s.io/networkpolicies
+#     to the ClusterServiceVersion clusterPermissions or permissions, or add the operator name
+#     and its major.minor version to the operator_network_policy_rbac_exceptions rule data key.
+#   collections:
+#   - redhat
+#   rule_data_fields:
+#   - operator_network_policy_rbac_exceptions
+#   effective_on: 2026-08-07T00:00:00Z
+deny contains result if {
+	some manifest in _csv_manifests
+	not _network_policy_rbac_excepted(manifest)
+	not _has_network_policy_rbac(manifest)
+	result := metadata.result_helper(rego.metadata.chain(), [_operator_package_name, manifest.spec.version])
+}
+
+_operator_package_name := input.image.config.Labels["operators.operatorframework.io.bundle.package.v1"]
+
+_major_minor_version(ver) := mm if {
+	parts := split(ver, ".")
+	count(parts) >= 2
+	mm := concat(".", [parts[0], parts[1]])
+}
+
+_network_policy_rbac_excepted(manifest) if {
+	exceptions := rule_data.get("operator_network_policy_rbac_exceptions")
+	version := _major_minor_version(manifest.spec.version)
+	version in exceptions[_operator_package_name]
+}
+
+_has_network_policy_rbac(manifest) if {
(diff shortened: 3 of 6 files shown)


policy/release/olm/olm.rego
```diff
@@ -382,6 +382,72 @@ deny contains result if {
...
+    some rule in perm.rules
+    _is_network_policy_rule(rule)
+}
+
+_is_network_policy_rule(rule) if {
+    "networking.k8s.io" in rule.apiGroups
+    "networkpolicies" in rule.resources
+    some verb in rule.verbs
+    verb in {"create", "update", "patch", "delete"}
+}
```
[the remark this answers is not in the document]
yashvardhannanavati: the goal of the check is to indeed pass with one of the verbs

policy/release/olm/olm.rego
```diff
@@ -382,6 +382,73 @@ deny contains result if {
...
+#     Operator %q version %q is missing required NetworkPolicy RBAC
+#     (networking.k8s.io/networkpolicies with create/update/patch/delete)
+#   solution: >-
+#     Add a rule granting create, update, patch, and delete on networking.k8s.io/networkpolicies
+#     to the ClusterServiceVersion clusterPermissions or permissions, or add the operator name
+#     and its major.minor version to the operator_network_policy_rbac_exceptions rule data key.
+#   collections:
+#   - redhat
+#   rule_data_fields:
+#   - operator_network_policy_rbac_exceptions
```
simonbaird, line 409: Hmm, documenting the rule data fields use in the metadata is a nice idea. I think this is the first time I've seen it though. Claude being creative?

policy/release/olm/olm.rego
```diff
@@ -382,6 +382,73 @@ deny contains result if {
...
+	some manifest in _csv_manifests
+	not _network_policy_rbac_excepted(manifest)
+	not _has_network_policy_rbac(manifest)
+	result := metadata.result_helper(rego.metadata.chain(), [_operator_package_name(manifest), manifest.spec.version])
+}
+
+_operator_package_name(manifest) := name if {
+	name := manifest.metadata.annotations["operators.operatorframework.io/bundle.package.v1"]
+} else := name if {
+	name := regex.replace(manifest.metadata.name, `\.v?\d+\.\d+.*$`, "")
```
simonbaird: A comment would be useful here to explain why there's a fallback, and what the regex replace is intending to do.

simonbaird (dismissed):
Lgtm, but consider my two comments/suggestions.

Nice PR.

policy/release/olm/olm.rego
yashvardhannanavati: yes, it was indeed Claude

policy/release/olm/olm.rego
yashvardhannanavati: good idea! done

simonbaird (dismissed):

policy/release/olm/olm.rego
```diff
@@ -384,6 +382,75 @@ deny contains result if {
...
+#   effective_on: 2026-07-07T00:00:00Z
+deny contains result if {
+	some manifest in _csv_manifests
+	not _network_policy_rbac_excepted(manifest)
+	not _has_network_policy_rbac(manifest)
+	result := metadata.result_helper(rego.metadata.chain(), [_operator_package_name(manifest), manifest.spec.version])
+}
+
+_operator_package_name(manifest) := name if {
+	name := manifest.metadata.annotations["operators.operatorframework.io/bundle.package.v1"]
```
joelanford: The definitive place to find the package name in a bundle is in `metadata/annotations.yaml` in the annotation `operators.operatorframework.io.bundle.package.v1`. Do we have access to that information in this rule somehow?

I also don't think I recall `csv.Metadata.Annotations["operators.operatorframework.io/bundle.package.v1"]` being a thing. I'm not sure we should invent a new place for the package name to need to be.

I also don't think it is safe to assume the CSV name prefix is the package name. There are multiple examples where that is not the case.

policy/release/olm/olm.rego
```diff
@@ -384,6 +382,75 @@ deny contains result if {
...
+	some perm in manifest.spec.install.spec.permissions
+	some rule in perm.rules
+	_is_network_policy_rule(rule)
+}
+
+_is_network_policy_rule(rule) if {
+	"networking.k8s.io" in rule.apiGroups
+	"networkpolicies" in rule.resources
+	some verb in rule.verbs
+	verb in {"create", "update", "patch", "delete"}
```
joelanford: It would also be acceptable to have the following for verbs, I think:
- `["*"]`
- `["create", "update", "delete"]`
- `["create", "patch", "delete"]`

policy/release/olm/olm.rego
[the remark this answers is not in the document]
joelanford: Actually, I think we're looking for specific sets of verbs. For example `delete`, by itself is definitely not indicative of the operand NP being managed by the operator.

Essentially what we want to see is the the operator has permission to handle the lifecycle of NP:
- creation
- changes
- deletion


So the rule should only pass if we see combinations of verbs that cover all three of those concepts.

policy/release/olm/olm.rego
joelanford: And these should be set-based checks. I.e.:
- does the verbs list contain all of `["*"]`?
- does the verbs list contain all of `["create", "update", "delete"]`?
- does the verbs list contain all of `["create", "patch", "delete"]`?

If any of the answers to those questions is "yes", check should pass.

joelanford: Another way to think about it:

```ruby
anyOf(
	"*",
	allOf(
		"create",
		anyOf(
			"update", "patch"
		),
		"delete"
	)
)
```

antora/docs/modules/ROOT/pages/packages/release_olm.adoc
```diff
@@ -83,6 +83,19 @@ Each image indicated as a related image should match an entry in the list of pre
 * Effective from: `2025-04-15T00:00:00Z`
 * https://github.com/conforma/policy/blob/{page-origin-refhash}/policy/release/olm/olm.rego#L232[Source, window="_blank"]
 
+[#olm__required_network_policy_rbac_for_operands]
+=== link:#olm__required_network_policy_rbac_for_operands[Required NetworkPolicy RBAC missing from OLM bundle]
+
+Operator bundles must request permissions to create, update, patch, and delete NetworkPolicies (networking.k8s.io/networkpolicies) in their ClusterServiceVersion. Bundles whose operator name and major.minor version are listed in the `operator_network_policy_rbac_exceptions` rule data key are exempt from this requirement.
```
joelanford: I think I would include something here that explains why it is required to request these permissions. Something like:

```
Operators are required to manage the network policies of their operands. 
We detected that your operator does not include RBAC rules that would
grant it permission to manage operand network policy.

...
```

policy/release/olm/olm.rego
yashvardhannanavati: aha, I thought it was grabbing the label from annotations. the forward slash is a typo. I'll fetch it directly from Labels then.

policy/release/olm/olm.rego
```diff
@@ -382,6 +382,85 @@ deny contains result if {
...
+
+_is_network_policy_rule(rule) if {
+	"networking.k8s.io" in rule.apiGroups
+	"networkpoliciesss" in rule.resources
+	_has_lifecycle_verbs(rule.verbs)
+}
+
+_has_lifecycle_verbs(verbs) if {
+	"*" in verbs
+}
```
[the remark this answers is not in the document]
yashvardhannanavati: fixed PR description

antora/docs/modules/ROOT/pages/packages/release_olm.adoc
yashvardhannanavati: done

asergienk (approved):
lgtm

policy/lib/rule_data/rule_data.rego
```diff
@@ -117,6 +117,9 @@ defaults := {
 		"ConsoleLink",
 		"ConsolePlugin",
 	],
+	# Used in release/olm.rego
+	# Operators excepted from NetworkPolicy RBAC requirement by name and major.minor version
+	"operator_network_policy_rbac_exceptions": {},
```
joelanford, line 122: When do we populate this?

joelanford:
Looks great!

policy/lib/rule_data/rule_data.rego
yashvardhannanavati: I have a PR up for that: https://github.com/release-engineering/rhtap-ec-policy/pull/215

policy/release/olm/olm.rego
```diff
@@ -382,6 +382,85 @@ deny contains result if {
...
+#   rule_data_fields:
+#   - operator_network_policy_rbac_exceptions
+deny contains result if {
+	some manifest in _csv_manifests
+	not _network_policy_rbac_excepted(manifest)
+	not _has_network_policy_rbac(manifest)
+	result := metadata.result_helper(rego.metadata.chain(), [_operator_package_name, manifest.spec.version])
+}
+
+_operator_package_name := input.image.config.Labels["operators.operatorframework.io.bundle.package.v1"]
```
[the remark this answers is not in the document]
yashvardhannanavati: for non-bundle images, _csv_manifests will be empty so `deny` body will never be executed.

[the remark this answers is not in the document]
yashvardhannanavati: A bundle without that label is an invalid bundle. So the scenario is unlikely

simonbaird (approved):
Lgtm. Beware, the way updates are happening nowdays, I can't predict when this will hit production RH Konflux. Better ask @joejstuart if you have a specific timeline in mind.

simonbaird:
I'll hit merge if you're happy to merge it.

yashvardhannanavati:
>  Beware, the way updates are happening nowdays, I can't predict when this will hit production RH Konflux. Better ask @joejstuart if you have a specific timeline in mind.

sounds good. thanks!

> I'll hit merge if you're happy to merge it.

@simonbaird yes, please. thank you!

PR #1722, opened 2026-05-07
feat: add in-toto statement discovery library

robnester-rh:
## Summary

- New `policy/lib/intoto/` library for discovering unsigned in-toto statements attached to images as OCI referrers
- Follows the established SBOM referrer discovery pattern (`policy/lib/sbom/sbom.rego`)
- Filters referrers by `application/vnd.in-toto+json` artifact type
- Provides `statements_by_predicate()` helper for filtering by predicate type (test-result, vuln-scan)
- Statements are unsigned — trust verification is handled separately in EC-1774

Part of EC-1773. Foundation for EC-1774 (chain of trust) and EC-1775 (test content verification).

## Test plan

- [x] 831/831 tests pass
- [x] 100% line coverage on new files
- [x] New test: statement discovery from raw in-toto referrer
- [x] New test: unrelated referrer types filtered out
- [x] New test: empty referrers returns empty set
- [x] New test: non-in-toto JSON blobs skipped
- [x] New test: predicate type filtering
- [x] New test: predicate constants
- [x] `make quiet-test opa-check fmt-check` all pass

### DIFF
--- a/policy/lib/intoto/intoto.rego
+++ b/policy/lib/intoto/intoto.rego
@@ -18,26 +18,18 @@ package lib.intoto
 
 import rego.v1
 
-# Artifact types used to discover in-toto attestations attached as OCI referrers.
 # TODO(EC-1773): Confirm the artifact type once the test task implementation is finalized.
-# Sigstore bundle format used by cosign attest.
-_sigstore_bundle_type := "application/vnd.dev.sigstore.bundle.v0.3+json"
+_artifact_type := "application/vnd.in-toto+json"
 
-# Raw in-toto statement format.
-_intoto_statement_type := "application/vnd.in-toto+json"
-
-_artifact_types := {
-	_sigstore_bundle_type,
-	_intoto_statement_type,
-}
-
-# statements returns the set of in-toto statements attached to the image as OCI referrers.
-# Supports both raw in-toto JSON and Sigstore bundle (DSSE envelope) formats.
+# statements returns the set of unsigned in-toto statements attached to the
+# image as OCI referrers. Trust is established via Chains provenance (EC-1774),
+# not via signatures on the statements themselves.
 statements contains statement if {
 	some referrer in ec.oci.image_referrers(input.image.ref)
-	referrer.artifactType in _artifact_types
+	referrer.artifactType == _artifact_type
 	blob := ec.oci.blob(referrer.ref)
-	statement := _parse_statement(blob)
+	statement := json.unmarshal(blob)
+	statement._type == "https://in-toto.io/Statement/v1"
 }
 
 # Filter statements by predicate type.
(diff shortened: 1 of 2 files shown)


policy/lib/intoto/intoto.rego
```diff
@@ -0,0 +1,42 @@
...
+
+# statements returns the set of unsigned in-toto statements attached to the
+# image as OCI referrers. Trust is established via Chains provenance (EC-1774),
+# not via signatures on the statements themselves.
+statements contains statement if {
+	some referrer in ec.oci.image_referrers(input.image.ref)
+	referrer.artifactType == _artifact_type
+	blob := ec.oci.blob(referrer.ref)
+	statement := json.unmarshal(blob)
+	statement._type == "https://in-toto.io/Statement/v1"
```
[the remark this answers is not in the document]
robnester-rh: Addressed in 441e9633. We cannot use bracket notation here because `opa fmt` reverts it back to dot notation. Instead, added a `# regal ignore:leaked-internal-reference` directive — same pattern used in `policy/release/attestation_type/attestation_type.rego:38` for the same `_type` field.

robnester-rh:
Responding to the Qodo findings:

**1. Missing DSSE unwrap (Requirement gap)** — This is intentional. Joe confirmed on EC-1773 that statements will be unsigned, with trust established via Chains provenance (EC-1774). Supporting signed bundles would introduce signing key concerns we do not want right now.

**2. Unsafe JSON unmarshal (Bug)** — In OPA, `json.unmarshal` on invalid input causes the rule body to fail silently (the statement is simply not produced). This is safe behavior and matches the established pattern in `policy/lib/sbom/sbom.rego:91` which also does direct `json.unmarshal(blob)` without a `json.is_valid` guard.

st3penta (approved):
lgtm

policy/lib/intoto/intoto.rego
```diff
@@ -0,0 +1,44 @@
...
+# image as OCI referrers. Trust is established via Chains provenance (EC-1774),
+# not via signatures on the statements themselves.
+statements contains statement if {
+	some referrer in ec.oci.image_referrers(input.image.ref)
+	referrer.artifactType == _artifact_type
+	blob := ec.oci.blob(referrer.ref)
+	statement := json.unmarshal(blob)
+
+	# regal ignore:leaked-internal-reference
+	statement._type == "https://in-toto.io/Statement/v1"
```
simonbaird, line 33: It might be reasonable to include intoto v0.1 here also. In `policy/release/attestation_type/attestation_type.rego` it uses rule_data.get("known_attestation_types"). So you could use that too, something like this:


```suggestion
	statement._type in rule_data.get("known_attestation_types")
```

simonbaird: Or you could define it here locally - not sure there's a lot of value in using the rule data.

robnester-rh: Good catch on v0.1 support — went with defining it locally rather than pulling from rule_data. The library's job is discovery ("can I parse this?") while `attestation_type.rego` handles enforcement ("does policy accept this?"). If we shared the rule_data key, a user removing v0.1 from `known_attestation_types` to enforce a no-v0.1 policy would get silent omission instead of a violation — the worst failure mode for a policy engine. Fixup coming in a follow-up PR on branch EC-1773_fixup.

PR #1725, opened 2026-05-11
feat(EC-1774): verify chain of trust for in-toto statements

robnester-rh:
## Summary
- Adds `verified_statements` and `verified_statements_by_predicate` to `lib.intoto`
- Two-level OCI referrer discovery → sigstore verification → task trust checking
- Vacuous truth guards prevent security bypass on empty sets
- 15 test cases covering all verification paths with 100% coverage

## Chain of Trust Model
```
Image (input.image.ref)
  └─ OCI referrer: unsigned in-toto statement
       └─ OCI referrer: signed SLSA provenance from Chains
            → ec.sigstore.verify_attestation (signature check)
            → tekton.untrusted_task_refs (task trust check)
```

Trust model: "one valid provenance chain suffices." A statement is verified if ANY provenance referrer attached to the statement provides valid Sigstore-attested SLSA provenance with all tasks trusted.

## Key Design Decisions
- **Fail-closed**: blob/parse/type failures silently exclude statements; consumer deny rules surface the absence
- **Vacuous truth guards**: `count > 0` on both attestations and tasks prevents empty-set bypass
- **No deny rules in library**: `trust.rego` provides data, not policy decisions
- **OPA v1.12.1 workaround**: `object.get` used instead of dot notation due to type-checker panic

## Files
- `policy/lib/intoto/trust.rego` — 98 lines, chain-of-trust verification
- `policy/lib/intoto/trust_test.rego` — 381 lines, 15 tests
- `policy/lib/intoto/intoto.rego` — minor comment update (cross-reference to trust.rego)

## Depends On
- #1722 (EC-1773: in-toto statement discovery) — merged
- #1723 (EC-1773 follow-up: v0.1 statement support) — merged

## Test plan
- [x] All 15 new test cases pass
- [x] 100% coverage on trust.rego
- [x] 882/882 full suite passes
- [x] Lint and OPA strict checks pass
- [x] No modifications to existing intoto.rego logic (comment-only change)

Ref: EC-1774

### DIFF
--- a/policy/lib/intoto/intoto.rego
+++ b/policy/lib/intoto/intoto.rego
@@ -27,8 +27,8 @@ _artifact_type := "application/vnd.in-toto+json"
 _known_types := {"https://in-toto.io/Statement/v0.1", "https://in-toto.io/Statement/v1"}
 
 # statements returns the set of unsigned in-toto statements attached to the
-# image as OCI referrers. Trust is established via Chains provenance (EC-1774),
-# not via signatures on the statements themselves.
+# image as OCI referrers. Trust is established via Chains provenance
+# verification; see verified_statements in trust.rego.
 statements contains statement if {
 	some referrer in ec.oci.image_referrers(input.image.ref)
 	referrer.artifactType == _artifact_type
--- a/policy/lib/intoto/trust.rego
+++ b/policy/lib/intoto/trust.rego
@@ -64,7 +64,7 @@ verified_statements_by_predicate(predicate_type) := {statement |
 
 _has_trusted_provenance(referrer) if {
 	some provenance_referrer in ec.oci.image_referrers(referrer.ref)
-	_verify_provenance(provenance_referrer)
+	_verify_provenance(provenance_referrer, referrer.digest)
 }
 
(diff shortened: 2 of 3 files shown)


robnester-rh:
Re: CodeRabbit's suggestion to add an `artifactType` filter on provenance referrers —

We deliberately chose **not** to filter second-level referrers by artifact type. The Chains-generated SLSA provenance is itself an in-toto statement (`predicateType: https://slsa.dev/provenance/v1`), so it may use the same `application/vnd.in-toto+json` artifact type as the test statement. Adding a filter like `artifactType == "application/vnd.dsse.envelope.v1+json"` could **break verification** if Chains uses a different type.

Instead, `ec.sigstore.verify_attestation` acts as a natural filter — non-provenance referrers fail signature verification and are silently skipped (fail-closed). This is documented in the spec and was explicitly debated across multiple committee review rounds.

If we later confirm what artifact type Chains uses for provenance referrers, we can add the filter as a performance optimization.

joejstuart:
LGTM.  I think maybe just checkout this comment - https://github.com/conforma/policy/pull/1725#discussion_r3222062918

joejstuart (dismissed):

joejstuart (approved):

PR #1728, opened 2026-05-15
feat(EC-1775): verify test attestation content

robnester-rh:
## Summary

- Add new policy package `policy.release.test_attestation` that verifies in-toto test-result attestation content
- 4 rules: `no_failed_tests` (deny), `no_test_warnings` (warn), `test_result_known` (deny), `test_data_found` (deny)
- 2 helpers: `_test_name` (extracts configuration name), `_test_list` (safely formats test name arrays)
- Consumes `intoto.verified_statements_by_predicate` from EC-1774's trust layer — only chain-of-trust-verified attestations are evaluated
- Complete no-op when no test-result attestations are present
- 13 test cases covering happy path, failure modes, edge cases (WARNED+FAILED coexistence, multiple failures, non-string result type safety)

## Design Decisions

- **Verify what exists** — checks attestations present, does not enforce a required tests list (deferred per Joe Stuart)
- **Decision on overall result, detail from arrays** — deny/warn driven by top-level `result` enum; optional `failedTests`/`warnedTests` arrays enrich error messages
- **Rule 3/4 mutual exclusion** — `test_result_known` guards with `statement.predicate.result` before checking set membership, preventing overlap with `test_data_found`
- **Test identity from predicate** — uses `configuration[0].name` rather than OCI annotations

## Test plan

- [x] 13 test cases all passing with msg content assertions
- [x] 913/913 total tests passing (baseline was 900)
- [x] OPA format clean
- [x] 100% coverage on new files
- [x] Committee-reviewed design spec and testing strategy
- [x] Two rounds of local code review with all findings addressed

Jira: EC-1775


robnester-rh:
As a note, we should probably ensure that https://github.com/conforma/cli/pull/3311 is merged before we merge this as it's predicated on some of the changes introduced in the CLI there.

antora/docs/modules/ROOT/pages/packages/release_test_attestation.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+* Rule type: [rule-type-indicator failure]#FAILURE#
+* FAILURE message: `Test attestation %q has a failed result, failed tests %s`
+* Code: `test_attestation.no_failed_tests`
+* https://github.com/conforma/policy/blob/{page-origin-refhash}/policy/release/test_attestation/test_attestation.rego#L51[Source, window="_blank"]
+
+[#test_attestation__no_test_warnings]
+=== link:#test_attestation__no_test_warnings[No test attestation warnings]
+
+Produce a warning if any test result attestation has a result of "WARNED". Warned test names from the attestation predicate are included in the message when available.
```
simonbaird, line 26: Was expecting this:


```suggestion
Produce a warning if any test result attestation has a result of "WARNING". Warned test names from the attestation predicate are included in the message when available.
```

simonbaird: Oh, is is because "WARNED" is as per the in-toto standard spec?

simonbaird: ..which is different to the app-studio TEST_DATA conventions.

robnester-rh: Correct — `WARNED` is from the [in-toto test-result predicate spec](https://github.com/in-toto/attestation/blob/main/spec/predicates/test-result.md) which uses `PASSED`/`FAILED`/`WARNED`. The existing TEST_OUTPUT convention uses `SUCCESS`/`FAILURE`/`WARNING`/`ERROR`/`SKIPPED`. The vocabularies are intentionally different since they come from different specs.

policy/release/test_attestation/test_attestation.rego
```diff
@@ -0,0 +1,158 @@
...
+#   attestations are present.
+#
+package test_attestation
+
+import rego.v1
+
+import data.lib.intoto
+import data.lib.metadata
+
+_test_attestations := intoto.verified_statements_by_predicate(intoto.predicate_test_result)
```
st3penta, line 34: The package is a deliberate no-op when no test-result attestations are attached, but Jira AC-1 says "Policy denies if no test-result attestations are attached to the image." PR description says this was deferred per Joe -- is that recorded anywhere besides this PR? The ticket still lists it as an acceptance criterion. Worth updating EC-1775 to reflect the scope reduction so the decision is traceable.

policy/release/test_attestation/test_attestation.rego
```diff
@@ -0,0 +1,158 @@
...
+_test_name(statement) := name if {
+	predicate := object.get(statement, "predicate", {})
+	config := object.get(predicate, "configuration", [])
+	count(config) > 0
+	name := config[0].name
+} else := "unknown test"
+
+_test_list(predicate, key) := result if {
+	value := object.get(predicate, key, [])
+	is_array(value)
```
st3penta, line 45: The `is_array` guard is untested. No test case provides a non-array value (string, object, number) for `failedTests`/`warnedTests`. That's a distinct equivalence class from "key missing" and "key present with valid array."

policy/release/test_attestation/test_attestation.rego
```diff
@@ -0,0 +1,158 @@
...
+	items := [x | some x in value; is_string(x)]
+	count(items) > 0
```
st3penta, line 47: This boundary is also untested. An empty `failedTests: []` takes a different path than a missing key, but both should produce "(none listed)."

policy/release/test_attestation/test_attestation.rego
```diff
@@ -0,0 +1,158 @@
...
+# custom:
+#   short_name: no_failed_tests
+#   failure_msg: Test attestation %q has a failed result, failed tests %s
+#   solution: >-
+#     Ensure all test attestations have a passing result. Review the
+#     failed tests listed in the attestation predicate.
+#   collections:
+#   - redhat
+#   depends_on:
+#   - attestation_type.known_attestation_type
```
st3penta, line 66: Worth noting a latent coupling here. `attestation_type.known_attestation_type` checks `lib.pipelinerun_attestations` -- attestations embedded in SLSA provenance via `input.attestations`. This package consumes a completely different trust chain: OCI referrer attestations verified independently through `intoto.verified_statements_by_predicate`.

Today both attestation pipelines are always present together, so `depends_on` works fine. But the semantic mismatch means if the project ever moves toward OCI referrer-only attestations (no pipelinerun attestation), `known_attestation_type` would deny and `depends_on` would suppress this package's findings -- even though they're independently valid from a separate trust chain.

This is the first consumer of `verified_statements_by_predicate` outside the intoto library, so there's no established `depends_on` pattern for referrer-based policies yet. Not actionable for this PR, but worth keeping in mind when a second referrer-based package appears -- that's when you'd want a new base rule analogous to `known_attestation_type` for the referrer trust chain.

policy/release/test_attestation/test_attestation.rego
```diff
@@ -0,0 +1,158 @@
...
+#
+deny contains result if {
+	some statement in _test_attestations
+	statement.predicate.result == "FAILED"
+	failed := _test_list(statement.predicate, "failedTests")
+	result := metadata.result_helper_with_term(
+		rego.metadata.chain(),
+		[_test_name(statement), failed],
+		_test_name(statement),
```
st3penta, line 75: The `term` resolves to `configuration[0].name`. If two attestations share the same config name (e.g. two "clair-scan" runs at different pipeline stages), their violations get identical terms. A policy exclusion targeting one silently suppresses the other.

The existing `test` package doesn't have this problem because its term comes from pipeline task names, which are unique within a pipeline. Test attestation config names have no such guarantee.

Is `configuration[0].name` expected to be unique across test-result attestations for a given image? If not, would incorporating additional identity (e.g. the statement subject digest) into the term produce more precise exclusion targeting?

policy/release/test_attestation/test_attestation.rego
```diff
@@ -0,0 +1,158 @@
...
+#     The test result attestation predicate must include a "result" field
+#     with a value of PASSED, WARNED, or FAILED.
+#   collections:
+#   - redhat
+#   depends_on:
+#   - attestation_type.known_attestation_type
+#
+deny contains result if {
+	some statement in _test_attestations
+	not statement.predicate.result
```
st3penta, line 125: `not statement.predicate.result` in Rego is true for both absent fields and falsy values (`false`, `null`, `0`, `""`). The failure message says "missing the required result field" but `result: false` would trigger this too. Intentional, or should this check for key absence specifically?

st3penta (commented):
Very good overall, but i have a few questions and test gaps

policy/release/test_attestation/test_attestation.rego
robnester-rh: Good catch on traceability. The deferral was per conversation with Joe — he wanted the initial version to focus on content verification of attestations that *are* present, with the "require attestations exist" gate as a follow-up. I'll update EC-1775 to formally record the scope reduction so it's not just buried in this PR description.

policy/release/test_attestation/test_attestation.rego
robnester-rh: Fixed — added `test_non_array_failed_tests` and `test_non_array_warned_tests` covering string and object values respectively. Both correctly produce "(none listed)" via the `is_array` guard fallback.

policy/release/test_attestation/test_attestation.rego
robnester-rh: Fixed — added `test_empty_failed_tests_array` and `test_empty_warned_tests_array`. Both confirm that `failedTests: []` produces "(none listed)" — same user-visible behavior as a missing key, but exercises the `count(items) > 0` branch rather than the `is_array` fallback.

policy/release/test_attestation/test_attestation.rego
robnester-rh: Great observation. You're right that `depends_on: attestation_type.known_attestation_type` creates a semantic coupling between two independent trust chains — pipelinerun attestations (via `input.attestations`) and OCI referrer attestations (via `verified_statements_by_predicate`).

As you noted, this works today because both pipelines are always present together. When a second referrer-based package appears, that's the right time to introduce a `depends_on` target analogous to `known_attestation_type` for the referrer trust chain. I'll add a note to track this.

policy/release/test_attestation/test_attestation.rego
robnester-rh: Good question. In practice, `configuration[0].name` is expected to be unique per test-result attestation because each attestation represents a distinct test execution (e.g. one clair-scan, one sanity-check). The in-toto test-result predicate spec models each attestation as a single test run.

However, you're right that there's no *structural* guarantee of uniqueness. If the same test name appears across multiple attestation stages, exclusions would collide. Incorporating the subject digest would add precision but also complexity — the term becomes harder to configure in policy exclusion rules.

For now I think the pragmatic approach is to keep `configuration[0].name` as the term (matches user expectations for exclusion targeting) and document the uniqueness assumption. If we encounter real-world collisions, we can extend the term at that point. Does that seem reasonable?

policy/release/test_attestation/test_attestation.rego
robnester-rh: Great question — I verified the actual Rego semantics here and the behavior is more nuanced than typical language falsy rules:

- **`false`**: `not statement.predicate.result` evaluates to `true` (Rego treats `not false` as true), so this is caught by `test_data_found`. This is the only "falsy" value that triggers this rule.
- **`null`, `0`, `""`**: These are *defined* values in Rego, so `not X` evaluates to `false`. They pass through to `test_result_known` where they're correctly caught as unsupported values since they're not in `{"PASSED", "FAILED", "WARNED"}`.
- **Missing key**: `not statement.predicate.result` is true (undefined), caught by `test_data_found`. ✓

So `result: false` *does* trigger "missing the required result field" rather than "unsupported result value" — but since `false` is equally invalid per the in-toto spec, the practical outcome (deny) is the same regardless of which rule catches it. I've added explicit tests for `false`, `null`, and `""` to document this behavior.

joejstuart (approved):

PR #1736, opened 2026-06-02
feat: deny releases using experimental Hermeto backends

jsmid1:
Builds whose SBOM contains components fetched by an experimental Hermeto backend (hermeto:backend:experimental:*) now violate the release policy. Covers both SPDX and CycloneDX SBOM formats.

Issue: [EC-1799](https://redhat.atlassian.net/issues?filter=-1&selectedIssue=EC-1799)

Co-Authored-By: Claude


policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -359,6 +359,40 @@ deny contains result if {
...
+#   solution: >-
+#     Use a supported, non-experimental package manager backend in your build
+#     process, or request a policy exception.
+#   collections:
+#   - minimal
+#   - redhat
+#   - redhat_rpms
+deny contains result if {
+	some s in sbom.cyclonedx_sboms
+	some annotation in s.annotations
```
simonbaird, line 385: This indicates that the annotation key is at the top level of the sbom data. Is that right?

simonbaird: I look at the requirements in the Jira and it seems like that's right.

simonbaird (dismissed):
I think it's good. Can you think of a way to test it on a real image with a real sbom that has an experimental Hermeto? The reporter might be able to share an image like that.

policy/release/sbom_spdx/sbom_spdx.rego
```diff
@@ -356,11 +356,50 @@ deny contains result if {
...
+#   - minimal
+#   - redhat
+#   - redhat_rpms
+deny contains result if {
+	some s in sbom.spdx_sboms
+	some pkg in s.packages
+
+	some annotation in pkg.annotations
+	annotation.annotator == "Tool: hermeto:backend"
+	startswith(annotation.comment, "hermeto:backend:experimental:")
```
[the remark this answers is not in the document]
jsmid1: Documented in the description of  https://redhat.atlassian.net/browse/EC-1799.

policy/release/sbom_spdx/sbom_spdx.rego
```diff
@@ -356,11 +356,50 @@ deny contains result if {
...
+
 _package_purl(pkg) := purl if {
 	purls := [ref.referenceLocator | some ref in pkg.externalRefs; ref.referenceType == "purl"]
 	purl := purls[0]
 } else := ""
 
+_spdx_package_id(pkg) := id if {
+	purls := [ref.referenceLocator | some ref in pkg.externalRefs; ref.referenceType == "purl"]
+	id := purls[0]
+} else := pkg.name
```
st3penta: This duplicates the PURL extraction logic from `_package_purl` just above, only differing in the fallback ("" vs `pkg.name`). Could reuse it:

```suggestion
_spdx_package_id(pkg) := purl if {
	purl := _package_purl(pkg)
	purl != ""
} else := pkg.name
```

policy/release/sbom_spdx/sbom_spdx_test.rego
```diff
@@ -1021,3 +1021,91 @@ test_proxy_metadata_required_spdx_vcs_url_passes if {
 		with ec.oci.image_tag_refs as []
 		with data.rule_data as _spdx_proxy_rule_data
 }
+
+# experimental_hermeto_backend tests
+
+test_experimental_hermeto_backend_spdx_denied if {
```
st3penta, line 1027: All tests use a single annotation per package. Might be worth adding a test where the same package carries both an experimental and a non-experimental annotation, to verify the rule fires only for the experimental one. In practice this scenario is unlikely (a package would rarely have two different backend annotations), but it would confirm the per-annotation iteration works as expected.

st3penta (commented):
Beware: these new rules don't have an `effective_on` date, meaning they will take effect immediately on deployment. 
If no production builds currently use experimental Hermeto backends, that's fine. Otherwise, an `effective_on` date,  and maybe a konflux announce would give teams a migration window.

Is the immediate enforcement intentional?

policy/release/sbom_spdx/sbom_spdx.rego
jsmid1: Done, thanks.

policy/release/sbom_spdx/sbom_spdx_test.rego
jsmid1: Added tests for both SPDX and CycloneDX with mixed annotations on the same package.

jsmid1:
> Beware: these new rules don't have an effective_on date, meaning they will take effect immediately on deployment.
If no production builds currently use experimental Hermeto backends, that's fine. Otherwise, an effective_on date, and maybe a konflux announce would give teams a migration window.

@st3penta I set the `effective_on` date to 1.8.2026. Would that be a long enough time window?
I will send the announcement once the PR gets merged.

simonbaird (approved):
Lgtm, nice PR.

simonbaird:
About the effective date, you could ask in Jira what the requestor and/or stakeholders think is the best choice. Perhaps they'd prefer something shorter. (A good technique just in case no-one cares, you can say "I'm planning to have this rule become affective on X, let me know if you prefer another date.)

One more note: FYI, the way we're deploying policy changes now, it takes up to two weeks to go from merged to deployed, which might be a reason to choose a later effective date.

jsmid1:
Tested with SPDX SBOM https://redhat.atlassian.net/browse/EC-1799?focusedCommentId=17374185.

st3penta (approved):

PR #1739, opened 2026-06-04
feat(EC-1881): add except_when support to disallowed_attributes rule data

robnester-rh:
## Summary

- Extends the `disallowed_attributes` rule data schema with an optional `except_when` field that suppresses violations when a package's PURL qualifier matches regex patterns
- Enables teams to allow specific disallowed attributes (e.g., `hermeto:pip:package:binary`) for packages from trusted indexes while still blocking untrusted sources
- Adds fail-closed semantics: if no purl-type ref exists, the qualifier is absent, or no pattern matches, the violation is still produced

## Test plan

- [ ] 941/941 tests pass with 100% coverage
- [ ] Schema validation rejects malformed `except_when` entries and invalid regex patterns
- [ ] SPDX `disallowed_package_attributes` suppresses violations when qualifier matches
- [ ] CycloneDX `disallowed_package_attributes` suppresses violations when qualifier matches
- [ ] Fail-closed: violations produced when no purl ref, missing qualifier, or no pattern match
- [ ] Multiple patterns and multiple `except_when` entries work correctly
- [ ] Entries without `except_when` behave exactly as before (backwards compatible)

### DIFF
--- a/policy/lib/sbom/sbom.rego
+++ b/policy/lib/sbom/sbom.rego
@@ -258,6 +258,21 @@ rule_data_errors contains error if {
 				"name": {"type": "string"},
 				"value": {"type": "string"},
 				"effective_on": {"type": "string", "format": "date-time"},
+				"except_when": {
+					"type": "array",
+					"items": {
+						"type": "object",
+						"properties": {
+							"purl_qualifier": {"type": "string"},
+							"patterns": {
+								"type": "array",
+								"items": {"type": "string", "format": "regex"},
+							},
+						},
+						"required": ["purl_qualifier", "patterns"],
+						"additionalProperties": false,
+					},
+				},
 			},
 			"additionalProperties": false,
 			"required": ["name"],
@@ -267,6 +282,32 @@ rule_data_errors contains error if {
 		"message": sprintf("Rule data %s has unexpected format: %s", [rule_data_attributes_key, e.message]),
 		"severity": e.severity,
 	}
+}
+
+# Verify items in disallowed_attributes except_when have valid regular expressions.
+rule_data_errors contains error if {
+	some attr_index, attr in rule_data.get(rule_data_attributes_key)
+	some ew in object.get(attr, "except_when", [])
+	some pattern in ew.patterns
+	not regex.is_valid(pattern)
+	error := {
+		"message": sprintf(
+			"Item at index %d in %s has an invalid regular expression in except_when: %q",
+			[attr_index, rule_data_attributes_key, pattern],
+		),
+		"severity": "failure",
+	}
+}
+
+# disallowed_attribute_excepted returns true when the package's PURL has a qualifier
+# matching an except_when condition, meaning the violation should be suppressed.
+disallowed_attribute_excepted(disallowed, purl_string) if {
+	purl_string != ""
+	some exception in disallowed.except_when
+	parsed := ec.purl.parse(purl_string)
+	some qualifier in parsed.qualifiers
+	qualifier.key == exception.purl_qualifier
+	url_matches_any_pattern(qualifier.value, exception.patterns)
 }
 
 # Verify allowed_external_references is an array of type/url pairs
--- a/policy/lib/sbom/sbom.rego
+++ b/policy/lib/sbom/sbom.rego
@@ -299,17 +299,6 @@ rule_data_errors contains error if {
 	}
 }
 
-# disallowed_attribute_excepted returns true when the package's PURL has a qualifier
-# matching an except_when condition, meaning the violation should be suppressed.
-disallowed_attribute_excepted(disallowed, purl_string) if {
-	purl_string != ""
-	parsed := ec.purl.parse(purl_string)
-	some exception in disallowed.except_when
-	some qualifier in parsed.qualifiers
-	qualifier.key == exception.purl_qualifier
-	url_matches_any_pattern(qualifier.value, exception.patterns)
-}
-
 # Verify allowed_external_references is an array of type/url pairs
 rule_data_errors contains error if {
 	some e in j.validate_schema(rule_data.get(rule_data_allowed_external_references_key), {
@@ -434,6 +423,17 @@ rule_data_errors contains error if {
 		"message": sprintf("%q is not a valid regular expression for PURL type %q", [pattern, purl_type]),
 		"severity": "failure",
 	}
+}
+
+# disallowed_attribute_excepted returns true when the package's PURL has a qualifier
+# matching an except_when condition, meaning the violation should be suppressed.
+disallowed_attribute_excepted(disallowed, purl_string) if {
+	purl_string != ""
+	parsed := ec.purl.parse(purl_string)
+	some exception in disallowed.except_when
+	some qualifier in parsed.qualifiers
+	qualifier.key == exception.purl_qualifier
+	url_matches_any_pattern(qualifier.value, exception.patterns)
 }
 
 # component_found_by_hermeto checks if a CycloneDX component was fetched by
--- a/policy/lib/sbom/sbom.rego
+++ b/policy/lib/sbom/sbom.rego
@@ -425,7 +425,7 @@ rule_data_errors contains error if {
 	}
 }
 
-# disallowed_attribute_excepted returns true when the package's PURL has a qualifier
+# disallowed_attribute_excepted checks if the package's PURL has a qualifier
 # matching an except_when condition, meaning the violation should be suppressed.
 disallowed_attribute_excepted(disallowed, purl_string) if {
 	purl_string != ""
@@ -434,7 +434,7 @@ disallowed_attribute_excepted(disallowed, purl_string) if {
 	some qualifier in parsed.qualifiers
 	qualifier.key == exception.purl_qualifier
 	url_matches_any_pattern(qualifier.value, exception.patterns)
-}
+} else := false
 
 # component_found_by_hermeto checks if a CycloneDX component was fetched by
 # cachi2 or hermeto, based on the component's properties.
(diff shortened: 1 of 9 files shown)


policy/lib/sbom/sbom.rego
```diff
@@ -269,6 +284,32 @@ rule_data_errors contains error if {
...
+
+# disallowed_attribute_excepted returns true when the package's PURL has a qualifier
+# matching an except_when condition, meaning the violation should be suppressed.
+disallowed_attribute_excepted(disallowed, purl_string) if {
+	purl_string != ""
+	some exception in disallowed.except_when
+	parsed := ec.purl.parse(purl_string)
+	some qualifier in parsed.qualifiers
+	qualifier.key == exception.purl_qualifier
+	url_matches_any_pattern(qualifier.value, exception.patterns)
```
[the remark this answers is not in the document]
robnester-rh: Fixed in 5d1b802 — moved `ec.purl.parse()` before the `except_when` iteration so the PURL is parsed once.

robnester-rh:
Addressed both findings from Fullsend review in 5d1b802:

- **Medium (edge-case)**: `_package_purl(pkg)` now uses array comprehension (`purls[0]`) instead of a complete rule that could conflict when multiple PURL refs exist with different locators.
- **Low (missing-test)**: The anchored regex patterns (`^https://...`) already prevent substring injection. The existing test coverage is sufficient since the patterns in the Jira spec all use anchored regexes.

policy/lib/sbom/sbom.rego
```diff
@@ -395,6 +425,17 @@ rule_data_errors contains error if {
...
 
+# disallowed_attribute_excepted returns true when the package's PURL has a qualifier
+# matching an except_when condition, meaning the violation should be suppressed.
+disallowed_attribute_excepted(disallowed, purl_string) if {
+	purl_string != ""
+	parsed := ec.purl.parse(purl_string)
+	some exception in disallowed.except_when
+	some qualifier in parsed.qualifiers
+	qualifier.key == exception.purl_qualifier
+	url_matches_any_pattern(qualifier.value, exception.patterns)
```
st3penta, line 436: `regex.match` does substring matching, not full-string matching. A pattern like `console\.redhat\.com` also matches `evil-console.redhat.com`. This is inherited from `url_matches_any_pattern`, but it now extends to a security-sensitive bypass path.

Worth either wrapping patterns with `^`/`$` in the matching function, or at minimum documenting the anchoring requirement so operators don't accidentally widen exception scope.

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
```diff
@@ -149,6 +149,8 @@ deny contains result if {
 	property.name == disallowed.name
 	object.get(property, "value", "") == object.get(disallowed, "value", "")
 
+	not sbom.disallowed_attribute_excepted(disallowed, object.get(component, "purl", ""))
```
st3penta, line 154: The METADATA description block for this rule (above) doesn't mention `except_when`. These annotations generate the public docs on conforma.dev and are the primary way operators discover rule data configuration. EC-1881 AC requires documentation updates.

The SPDX counterpart has the same gap. Worth describing the new field, its schema, and matching semantics in both METADATA blocks.

policy/release/sbom_spdx/sbom_spdx_test.rego
```diff
@@ -423,6 +423,182 @@ _sbom_attestation := {"statement": {
...
+		},
+	}])
+
+	results := sbom_spdx.deny with input.attestations as [att]
+		with input.image.ref as "registry.local/spam@sha256:1230000000000000000000000000000000000000000000000000000000000123"
+		with ec.oci.image_referrers as []
+		with ec.oci.image_tag_refs as []
+		with data.rule_data as {sbom.rule_data_attributes_key: disallowed_attributes}
+
+	count({r | some r in results; r.code == "sbom_spdx.disallowed_package_attributes"}) == 0
```
st3penta, line 582: CycloneDX has `test_attributes_except_when_without_except_when_unchanged` covering mixed entries with/without `except_when`. Missing the SPDX equivalent. The two formats have structurally different iteration patterns, so the backward-compat path is worth testing independently.

policy/lib/sbom/sbom.rego
```diff
@@ -395,6 +425,17 @@ rule_data_errors contains error if {
...
+}
```
st3penta: Sibling predicates like `component_found_by_hermeto` and `package_found_by_hermeto` use `else := false`. This one relies on implicit undefined-is-falsy. Adding `else := false` would make the fail-closed intent explicit.

policy/release/sbom_cyclonedx/sbom_cyclonedx_test.rego
```diff
@@ -214,6 +214,188 @@ test_attributes_not_allowed_value_no_purl if {
 		with data.rule_data as {sbom.rule_data_attributes_key: [{"name": "syft:distro:id", "value": "rhel"}]}
 }
 
+test_attributes_except_when_match_suppresses_violation if {
```
st3penta, line 217: SPDX tests extract `_spdx_excepted_package` to reduce boilerplate. These tests inline identical component structures across ~7 tests. A similar helper would cut the duplication.

policy/release/sbom/sbom_test.rego
```diff
@@ -190,6 +192,18 @@ test_rule_data_validation if {
...
+			# regal ignore:line-length
+			"msg": "Rule data disallowed_attributes has unexpected format: 8.except_when.0.patterns.0: Does not match format 'regex'",
+			"severity": "failure",
+		},
+		{
+			"code": "sbom.disallowed_packages_provided",
+			# regal ignore:line-length
+			"msg": "Item at index 8 in disallowed_attributes has an invalid regular expression in except_when: \"[\"",
+			"severity": "failure",
+		},
```
st3penta, line 206: Schema validation tests cover the invalid regex case but don't include a well-formed `except_when` entry among the valid inputs. The positive path is only implicitly tested via integration tests.

lcarva:
I didn't see this and created a similar PR (https://github.com/conforma/policy/pull/1744) 🫣
I'll close that out but mentioning it here in case it's helpful.

Also, I have an image that can be used to verify this, feel free to use it. Here's the snapshot:
```json
{
  "application": "lucarval-test",
  "artifacts": {},
  "componentGroup": "",
  "components": [
    {
      "containerImage": "quay.io/redhat-user-workloads/rhtap-shared-team-tenant/awesome-fermi-c9ff1@sha256:27e73129cb102cff44c47b31cd5beef2b0a8cdd337ba42bac88613cdb3f9a326",
      "name": "awesome-fermi-c9ff1",
      "source": {
        "git": {
          "revision": "d764b86481955715da6a34ec6d15d068e65a25dc",
          "url": "https://github.com/lucarval-rhtap/awesome-fermi"
        }
      },
      "version": ""
    }
  ]
}
```

policy/lib/sbom/sbom.rego
simonbaird: +1 to that suggestion

policy/release/sbom_spdx/sbom_spdx.rego
```diff
@@ -352,6 +354,11 @@ deny contains result if {
 	)
 }
 
+_package_purl(pkg) := purl if {
```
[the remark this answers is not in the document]
simonbaird: meh

policy/lib/sbom/sbom.rego
robnester-rh: Good catch. The substring matching is inherited from `url_matches_any_pattern` which is shared with `allowed_package_sources` and `allowed_proxy_urls`. Changing it would affect all callers. The Jira spec's example patterns use anchored regexes (`^https://...`) and the docs should make this explicit. I've updated the METADATA descriptions to document except_when. For the anchoring concern specifically — should we open a follow-up to audit all pattern-matching call sites, or is documenting the anchoring requirement sufficient?

policy/release/sbom_cyclonedx/sbom_cyclonedx.rego
robnester-rh: Fixed in 7254fac — updated both CycloneDX and SPDX METADATA descriptions to document except_when.

policy/release/sbom_spdx/sbom_spdx_test.rego
robnester-rh: Fixed in 7254fac — added `test_attributes_except_when_without_except_when_unchanged` for SPDX, matching the CycloneDX equivalent.

policy/lib/sbom/sbom.rego
robnester-rh: Fixed in 7254fac — added `else := false` matching the pattern used by `component_found_by_hermeto` and `package_found_by_hermeto`.

policy/release/sbom_cyclonedx/sbom_cyclonedx_test.rego
robnester-rh: Fixed in 7254fac — extracted `_cdx_excepted_component(purl, attr_name, attr_value)` helper, mirroring the SPDX `_spdx_excepted_package` pattern.

policy/release/sbom/sbom_test.rego
robnester-rh: Fixed in 7254fac — added a valid except_when entry to the schema validation test's positive path.

robnester-rh:
Thanks @lcarva — appreciate the heads up on PR #1744, and the test image is very helpful! I'll use it for manual verification.

policy/lib/sbom/sbom.rego
st3penta: I'm ok with tracking this in a folloup jira

st3penta (approved):

policy/lib/sbom/sbom.rego
robnester-rh: Will create a follow-up ticket to audit all `regex.match` pattern-matching call sites for anchoring. Thanks for the review!

joejstuart (approved):
LGTM

PR #1745, opened 2026-06-11
feat: Ignore local go dependencies when using proxy

a-ovchinnikov:
When fetching dependencies for a golang project Hermeto will include local modules shipped with the compiler as dependencies, it will also include VCS dependencies if such are present. Since none of these dependencies could originate from an artifact registry proxy their inclusion would result in a violation when such a proxy is enabled.

This change updates dependency exclusion logic to prevent Conforma from marking local golang dependencies as policy violations.

### DIFF
--- a/policy/lib/sbom/sbom.rego
+++ b/policy/lib/sbom/sbom.rego
@@ -460,6 +460,7 @@ package_found_by_hermeto(pkg) if {
 is_registry_dependency(parsed_purl, dependency) if {
 	_is_registry_purl(parsed_purl)
 	not _is_bundled(dependency)
+	not _is_local_gomod_dep(parsed_purl)
 }
 
 _is_registry_purl(parsed_purl) if {
@@ -481,6 +482,45 @@ _is_bundled(dependency) if {
 	parsed := json.unmarshal(annotation.comment)
 	parsed == _npm_bundled_property
 }
+
+# Hermeto counts standard libraries mentioned in go.mod as dependencies and
+# includes them into SBOM. Since they won't originate from an artifact
+# registry proxy they must be excluded. The same logic applies to packages
+# originating from a VCS.
+_is_local_gomod_dep(parsed_purl) if {
+	parsed_purl.type == "golang"
+	_gomod_locality_rule(parsed_purl)
+}
+
+# A gomod package will have vcs_url qualifier either if it is the main package
+# or a module was obtained from local file system via a local replace.
+# The relevant lines in Hermeto:
+#  https://github.com/hermetoproject/hermeto/blob/db5e5f9d4f3dd2e44316a1cb2c44368c92007d54/
+#	hermeto/core/package_managers/gomod/main.py#L216
+#  https://github.com/hermetoproject/hermeto/blob/db5e5f9d4f3dd2e44316a1cb2c44368c92007d54/
+#	hermeto/core/package_managers/gomod/main.py#L232
+# go mod local replace:
+#  https://go.dev/doc/modules/gomod-ref#replace
+_gomod_locality_rule(parsed_purl) if {
+	qualifiers := object.get(parsed_purl, "qualifiers", [])
+
+	some qualifier in qualifiers
+	qualifier.key == "vcs_url"
+
+	# This is not supposed to happen ever, but in case it happens empty vcs_urls are forbidden.
+	not _is_empty(qualifier.value)
+}
+
+# Missing version indicates that this is either the main module (i.e. the package that is
+# being built which origin we cannot establish) or a standard package shipped with golang.
+# In this case the version will be missing and will not be reported in SBOM:
+#  https://github.com/hermetoproject/hermeto/blob/db5e5f9d4f3dd2e44316a1cb2c44368c92007d54/
+#	hermeto/core/package_managers/gomod/main.py#L244
+# In all other cases the version must be set to something:
+#  https://go.dev/ref/mod#versions
+_gomod_locality_rule(parsed_purl) if _is_empty(parsed_purl.version)
+
+_is_empty(field) if field in {null, ""}
 
 _npm_bundled_property := {"name": "cdx:npm:package:bundled", "value": "true"}
 
--- a/policy/lib/sbom/sbom_test.rego
+++ b/policy/lib/sbom/sbom_test.rego
@@ -488,3 +488,33 @@ test_disallowed_attribute_excepted_multiple_except_when if {
 	purl := "pkg:pypi/some-lib@1.0?index_url=https://internal.example.com/pypi/"
 	sbom.disallowed_attribute_excepted(disallowed, purl)
 }
+
+test_golang_non_registry_empty_version_dependencies_excluded if {
+	parsed_purl := {"type": "golang", "name": "localdep", "version": "", "qualifiers": []}
+	sbom._is_local_gomod_dep(parsed_purl)
+}
+
+test_golang_non_registry_null_version_dependencies_excluded if {
+	parsed_purl := {"type": "golang", "name": "localdep", "version": null, "qualifiers": []}
+	sbom._is_local_gomod_dep(parsed_purl)
+}
+
+test_golang_non_registry_vcs_dependencies_excluded if {
+	parsed_purl := {"type": "golang", "name": "localdep", "qualifiers": [{"key": "vcs_url", "value": "https://github.com/example/lib.git"}]}
+	sbom._is_local_gomod_dep(parsed_purl)
+}
+
+test_golang_empty_present_vcs_url_not_excluded if {
+	parsed_purl := {"type": "golang", "name": "localdep", "qualifiers": [{"key": "vcs_url", "value": ""}]}
+	not sbom._is_local_gomod_dep(parsed_purl)
+}
+
+test_golang_null_present_vcs_url_not_excluded if {
+	parsed_purl := {"type": "golang", "name": "localdep", "qualifiers": [{"key": "vcs_url", "value": null}]}
+	not sbom._is_local_gomod_dep(parsed_purl)
+}
+
+test_golang_regular_package_not_rejected_as_local if {
+	parsed_purl := {"type": "golang", "name": "adep", "version": "v0.0.1", "qualifiers": []}
+	not sbom._is_local_gomod_dep(parsed_purl)
+}


policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,25 @@ _is_bundled(dependency) if {
 	parsed == _npm_bundled_property
 }
 
+# Hermeto counts standard libraries mentioned in go.mod as dependencies and
+# includes them into SBOM. Since they won't originate from an artifact
+# registry proxy they must be excluded. The same logic applies to packages
+# originating from a VCS.
+_is_local_gomod_dep(parsed_purl) if {
```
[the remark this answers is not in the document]
a-ovchinnikov: This looks like a declarative rule set.

policy/lib/sbom/sbom_test.rego
```diff
@@ -415,3 +415,13 @@ test_is_registry_dependency_no_properties_or_annotations if {
 	parsed_purl := {"type": "npm", "name": "lib", "qualifiers": []}
 	sbom.is_registry_dependency(parsed_purl, {})
 }
+
```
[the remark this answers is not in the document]
a-ovchinnikov: This is a declarative system, the new rule is AND-ed to existing ones. Further, since this a variety of Datalog any statements about "return" values are incorrect.

taylormadore:
LGTM

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,30 @@ _is_bundled(dependency) if {
...
+	parsed_purl.type == "golang"
+	_gomod_locality_rule(parsed_purl)
+}
+
+_gomod_locality_rule(parsed_purl) if {
+	qualifiers := {q.key | some q in object.get(parsed_purl, "qualifiers", [])}
+	"vcs_url" in qualifiers
```
simonbaird: Maybe it's an unlikely edge case, but do we need to make sure the value is present just in case the value of the qualifier is a null or an empty string?

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,30 @@ _is_bundled(dependency) if {
...
+}
+
+_gomod_locality_rule(parsed_purl) if {
+	"version" in object.keys(parsed_purl)
+	parsed_purl.version == ""
+}
```
simonbaird: I agree with the robot suggesting some explanatory comments would be good. Are we saying that any purl without a version considered to be local? How come? Are we confident there'll never be a non-local go module without a version key?

policy/lib/sbom/sbom_test.rego
```diff
@@ -415,3 +415,18 @@ test_is_registry_dependency_no_properties_or_annotations if {
...
```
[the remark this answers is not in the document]
simonbaird: +1 it's probably worth adding a little testing so we have that covered.

policy/lib/sbom/sbom.rego
simonbaird: I guess some explanation would be good here also. Why does the presence of this qualifier indicate the go module is local?

a-ovchinnikov: Added explanatory comments and a rule to reject empty vsc_urls (which should not appear, but let's reject them for the sake of completeness).

policy/lib/sbom/sbom.rego
a-ovchinnikov: Added comments explaining why this is a safe assumption.

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,53 @@ _is_bundled(dependency) if {
...
+_gomod_locality_rule(parsed_purl) if {
+	qualifiers := object.get(parsed_purl, "qualifiers", [])
+
+	qualifier_keys := {q.key | some q in qualifiers}
+	"vcs_url" in qualifier_keys
+
+	# This is not supposed to happen ever, but in case it happens empty vcs_urls are forbidden.
+	vcs_url := {h.value | some h in qualifiers; h.key == "vcs_url"}
+	vcs_url != null
+	vcs_url != ""
```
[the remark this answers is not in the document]
a-ovchinnikov: This is a good point, reworked the rule.

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,53 @@ _is_bundled(dependency) if {
...
+# Missing version indicates that this is either the main module (i.e. the package that is
+# being built which origin we cannot establish) or a standard package shipped with golang.
+# In this case the version will be missing and will not be reported in SBOM:
+#  https://github.com/hermetoproject/hermeto/blob/db5e5f9d4f3dd2e44316a1cb2c44368c92007d54/
+#	hermeto/core/package_managers/gomod/main.py#L244
+# In all other cases the version must be set to something:
+#  https://go.dev/ref/mod#versions
+_gomod_locality_rule(parsed_purl) if {
+	"version" in object.keys(parsed_purl)
+	parsed_purl.version == ""
```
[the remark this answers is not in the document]
a-ovchinnikov: > If Hermeto omits the version key entirely from the purl object (which is the natural representation of 'missing version')

No.

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,55 @@ _is_bundled(dependency) if {
 	parsed == _npm_bundled_property
 }
 
+# Hermeto counts standard libraries mentioned in go.mod as dependencies and
+# includes them into SBOM. Since they won't originate from an artifact
+# registry proxy they must be excluded. The same logic applies to packages
+# originating from a VCS.
+_is_local_gomod_dep(parsed_purl) if {
```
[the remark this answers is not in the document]
a-ovchinnikov: >  dead-code

header directly contradicts the

> This branch serves as defense-in-depth.

statement.

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,55 @@ _is_bundled(dependency) if {
...
+	vcs_url != null
+	vcs_url != ""
+}
+
+# Missing version indicates that this is either the main module (i.e. the package that is
+# being built which origin we cannot establish) or a standard package shipped with golang.
+# In this case the version will be missing and will not be reported in SBOM:
+#  https://github.com/hermetoproject/hermeto/blob/db5e5f9d4f3dd2e44316a1cb2c44368c92007d54/
+#	hermeto/core/package_managers/gomod/main.py#L244
+# In all other cases the version must be set to something:
```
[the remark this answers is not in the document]
a-ovchinnikov: This is a request to test behavior external to the code, resolving.

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,55 @@ _is_bundled(dependency) if {
...
+	_gomod_locality_rule(parsed_purl)
+}
+
+# A gomod package will have vcs_url qualifier either if it is the main package
+# or a module was obtained from local file system via a local replace.
+# The relevant lines in Hermeto:
+#  https://github.com/hermetoproject/hermeto/blob/db5e5f9d4f3dd2e44316a1cb2c44368c92007d54/
+#	hermeto/core/package_managers/gomod/main.py#L216
+#  https://github.com/hermetoproject/hermeto/blob/db5e5f9d4f3dd2e44316a1cb2c44368c92007d54/
+#	hermeto/core/package_managers/gomod/main.py#L232
```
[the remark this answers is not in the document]
a-ovchinnikov: >  this correctly implements the documented Hermeto contract

This adds noise, resolving.

policy/lib/sbom/sbom_test.rego
```diff
@@ -415,3 +415,33 @@ test_is_registry_dependency_no_properties_or_annotations if {
...
```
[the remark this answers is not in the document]
a-ovchinnikov: Due to the nature of the language I think it is sufficient to verify every individual conjunction term separately, resolving.

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,55 @@ _is_bundled(dependency) if {
...
+
+	qualifier_keys := {q.key | some q in qualifiers}
+	"vcs_url" in qualifier_keys
+
+	# This is not supposed to happen ever, but in case it happens empty vcs_urls are forbidden.
+	some qualifier in qualifiers
+	qualifier.key == "vcs_url"
+	vcs_url := qualifier.value
+	vcs_url != null
+	vcs_url != ""
```
simonbaird: It feels clunky to reference qualifiers twice. See if something like this can work the same:

```suggestion
_gomod_locality_rule(parsed_purl) if {
	qualifiers := object.get(parsed_purl, "qualifiers", [])

	some q in qualifiers
	q.key == "vcs_url"
	
	# Should never happen in reality, but don't match an an empty vcs_url
	q.value != ""
```

Also, from [here](https://github.com/conforma/cli/blob/8e0e5b548d2ef865e4c6695fd26d4ee115a3b11b/internal/rego/purl/purl.go#L147-L153) I can tell it's always a string, so we don't need to check for null.

simonbaird: If you prefer to make it a little shorter, this should work the same:

```
    some q in object.get(parsed_purl, "qualifiers", [])
```

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,55 @@ _is_bundled(dependency) if {
...
+# Missing version indicates that this is either the main module (i.e. the package that is
+# being built which origin we cannot establish) or a standard package shipped with golang.
+# In this case the version will be missing and will not be reported in SBOM:
+#  https://github.com/hermetoproject/hermeto/blob/db5e5f9d4f3dd2e44316a1cb2c44368c92007d54/
+#	hermeto/core/package_managers/gomod/main.py#L244
+# In all other cases the version must be set to something:
+#  https://go.dev/ref/mod#versions
+_gomod_locality_rule(parsed_purl) if {
+	"version" in object.keys(parsed_purl)
+	parsed_purl.version == ""
```
simonbaird:
```suggestion
	parsed_purl.version == ""
```

I'm pretty sure the object.keys check doesn't do anything useful and can be skipped here. The `parsed_purl.version` line will already quit out of the rule if the key is not there.

policy/lib/sbom/sbom.rego
```diff
@@ -441,6 +442,55 @@ _is_bundled(dependency) if {
...
+}
+
+_gomod_locality_rule(parsed_purl) if {
+	"version" in object.keys(parsed_purl)
+	parsed_purl.version == null
+}
```
simonbaird: Could remove this entirely I think, see [here](https://github.com/conforma/cli/blob/main/internal/rego/purl/purl.go#L158), but if you want to explicitly check for the null, I think you could combine this and the above rule like this:
```
    parsed_purl.version in {null, ""}
```

simonbaird (commented):
Thanks for the explanations. Just a few minor nitpicks about redundant rego, but otherwise lgtm.

policy/lib/sbom/sbom.rego
simonbaird: Similarly if you want to be defensive about null values in the other rule about vcs_url, I think this would work:

```
  not q.value in {null, ""}
```

There's also an is_string function that could also eliminate nulls, e.g.:

```
   is_string(q.value)
   q.value != ""
```

policy/lib/sbom/sbom.rego
a-ovchinnikov: Oh, yes, this looks much better, thank you!

policy/lib/sbom/sbom.rego
a-ovchinnikov: Done.

policy/lib/sbom/sbom.rego
a-ovchinnikov: Done.

policy/lib/sbom/sbom.rego
```diff
@@ -482,6 +483,45 @@ _is_bundled(dependency) if {
...
+# Missing version indicates that this is either the main module (i.e. the package that is
+# being built which origin we cannot establish) or a standard package shipped with golang.
+# In this case the version will be missing and will not be reported in SBOM:
+#  https://github.com/hermetoproject/hermeto/blob/db5e5f9d4f3dd2e44316a1cb2c44368c92007d54/
+#	hermeto/core/package_managers/gomod/main.py#L244
+# In all other cases the version must be set to something:
+#  https://go.dev/ref/mod#versions
+_gomod_locality_rule(parsed_purl) if _is_empty(parsed_purl.version)
+
+_is_empty(field) if field in {null, ""}
```
simonbaird, line 523: Nice.

simonbaird (approved):

simonbaird:
If you're happy with this, I'm happy to merge it.

a-ovchinnikov:
>  I'm happy to merge it.

Yes, sure, let's do it! Thank you!

PR #1766, opened 2026-07-03
Add redhat_security policy collection

st3penta:
Add a `redhat_security` collection containing the 124 release rules that map to
a ProdSec requirement. The mapping comes from the conforma-prodsec internal docs.

The existing `redhat` collection is unchanged. This is purely additive.

- New collection definition at `policy/release/collection/redhat_security/`
- 124 rules across 33 rego files get `redhat_security` in their `collections:` annotation
- 5 rules that previously had no `collections:` field now have one
- Updated `maven_repos_test.rego` assertions to match

20 non-release rules (pipeline, task, build_task, stepaction) and 2 builtin rules
are in the mapping but not tagged, since collections only apply to release rules.

Ref: EC-1967

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


simonbaird (dismissed):
Lgtm.

simonbaird:
Will the new collection appear right away in https://conforma.dev/docs/policy/release_policy.html#_available_rule_collections ?

st3penta:
@simonbaird yes it will. i didn't see it as a problem, since we already have the `redhat` collection, but we can discuss about possible alternatives if you don't like that

st3penta:
/fs-fix rebase on top of main and resolve the conflicts

st3penta:
oh fullsend doesn't run on forks...makes sense. I'll do it myself like in 2025

Acepresso (commented):
Do we have a test that covers the use of the new collection?

st3penta:
@Acepresso AFAIK we don't have tests that cover the specific collections.
We have a validation that ensures that rules tagged with `depends_on` and their dependencies are both included in a collection, but that's it.
No new rules were added in this case, it's just a "metadata" change.

Were you thinking about something specifically?

simonbaird (approved):

Acepresso:
> @Acepresso AFAIK we don't have tests that cover the specific collections. We have a validation that ensures that rules tagged with `depends_on` and their dependencies are both included in a collection, but that's it. No new rules were added in this case, it's just a "metadata" change.
> 
> Were you thinking about something specifically?

I was thinking of the most basic use case - when a policy includes the new collection, the rules are included and the exit code is 0.

st3penta:
> when a policy includes the new collection, the rules are included and the exit code is 0.

@Acepresso the CLI repo has unit tests in `internal/evaluator/filters_test.go` that cover the collection filtering mechanism. They verify that when a config includes a collection, all the rules tagged with that collection are evaluated.
I think this covers what you're asking, wdyt?

Acepresso:
> > when a policy includes the new collection, the rules are included and the exit code is 0.
> 
> @Acepresso the CLI repo has unit tests in `internal/evaluator/filters_test.go` that cover the collection filtering mechanism. They verify that when a config includes a collection, all the rules tagged with that collection are evaluated. I think this covers what you're asking, wdyt?

Indeed. It doesn't check all collections, though it checks the mechanism which I think is sufficient.
Thanks for verifying that.

Acepresso (approved):

PR #1806, opened 2026-08-12
feat(EC-2031): remove data.trusted_task_rules in favor of rule_data only

BohdanMar:
## Summary

- Remove the unvalidated `data.trusted_task_rules` input path — all trusted task rules now flow exclusively through `lib_rule_data("trusted_task_rules")` which receives JSON schema validation
- Simplify `_trusted_task_rules_data` by removing the `array.concat` merge logic
- Delete the dedicated `data_errors` validator for the old path
- Remove `.regal/config.yaml` exemptions for `data.trusted_task_rules`
- Update all tests to supply data via `data.rule_data.trusted_task_rules`

Resolves [EC-2031](https://redhat.atlassian.net/browse/EC-2031)

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -95,8 +95,6 @@ rules:
         - data.rule_data_custom
         - data.trusted_tasks
         - data.trusted_tasks.*
-        - data.trusted_task_rules
-        - data.trusted_task_rules.*
     # To fix this rule, in the policy/lib folder we need to remove the imports
     # from data.lib and use the fully qualified name when using function calls.
     # E.g.: from `lib.assert_equal` to `data.lib.assert_equal`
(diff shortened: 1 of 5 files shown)


jsmid1 (commented):
Could we add a test that exercises both allow and deny rules together through data.rule_data.trusted_task_rules?
Otherwise lgtm

BohdanMar:
@jsmid1 I believe that test_rule_data_trusted_task_rules_extraction (line 488) does this — it defines rules with both allow and deny keys via data.rule_data.trusted_task_rules and tests allow, deny (including deny-takes-precedence), and empty fallback. It was the existing counterpart to the deleted test_data_trusted_task_rules_extraction. But if you had a specific scenario in mind - please let me know.

jsmid1 (approved):

PR #1813, opened 2026-08-13
Fix test mock _type to match stepaction output

joejstuart:
## Summary
- The `stepaction-attest-test-result` produces in-toto statements with `_type: "https://in-toto.io/Statement/v0.1"`, but the test mocks in `test_attestation_test.rego` were using `v1`
- Both values are accepted by the intoto library's `_known_types`, so tests passed either way, but the mocks should reflect what the real producer emits
- Updated all 5 occurrences in mock statement builders

## Test plan
- [x] `make fmt` clean
- [x] `make test` passes (1070/1070)


robnester-rh (approved):
LGTM

PR #1842, opened 2026-09-08
docs: require direct unit tests for new helper functions

dheerajodha:
#### What:

Document that new public helper functions in policy/lib/ must have direct unit tests, not just indirect coverage through consuming rules. This reflects the existing convention enforced during code review but was not previously documented in AGENTS.md or .claude/rules/policy-rules.md.

#### Why:

This helps us to ensure accurate test coverage of our helper functions.

#### Tickets:

Resolves: EC-2098


simonbaird (approved):
Okay I guess.


Ps, I though we were supposed to be instruct Fullsend to make these PRs automatically. Not sure exactly how, but I think you type `/fs-fix` in the upstream issue, or something like that.

cuipinghuo (approved):

dheerajodha:
> Okay I guess.
> 
> Ps, I though we were supposed to be instruct Fullsend to make these PRs automatically. Not sure exactly how, but I think you type `/fs-fix` in the upstream issue, or something like that.

Ooo I missed that.

PR #1847, opened 2026-09-13
docs(EC-2103): fix stale trusted_task_rules architecture in THREAT_MODEL.md

cuipinghuo:
#### What:
Updates `THREAT_MODEL.md` to match the current `trusted_task_rules` architecture. After PR #1806 removed the `data.trusted_task_rules` input path and the dual-source merge from `policy/lib/tekton/trusted.rego`, the threat model still described the old additive-merge design. This rewrites the affected sections:

- **Section 3.3**: describes `trusted_task_rules` as single-source through `lib_rule_data("trusted_task_rules")`, while noting the legacy `trusted_tasks` allowlist still merges `data.trusted_tasks` via `object.union`.
- **Threat DP-2**: injection into `trusted_task_rules` is now only via ruleData (no system-vs-ruleData merge to exploit); legacy `trusted_tasks` merge behavior called out separately.
- **Threat LE-4**: reframed around flattening of named allow/deny groups within a single ruleData object; deny-before-allow precedence retained.
- **Open question 3**: the removed system-vs-ruleData merge precedence question is now moot; narrowed to overlapping patterns across groups.
- **Recommendation 6**: updated to reference the single ruleData-sourced set.

`grep -c 'data.trusted_task_rules' THREAT_MODEL.md` now returns 0.

#### Why:
The document described a trust boundary that no longer exists in code. A reviewer relying on it would incorrectly believe `trusted_task_rules` can flow from two independent data sources with additive merge semantics, when there is now a single validated path through `lib_rule_data`.

#### Tickets:
https://redhat.atlassian.net/browse/EC-2103 (upstream https://github.com/conforma/policy/issues/1807)


st3penta (approved):

jsmid1:
Seems like another PR exists for the same issue: https://github.com/conforma/policy/pull/1810.

jsmid1 (approved):
lgtm

cuipinghuo:
> Seems like another PR exists for the same issue: #1810.

I'll close #1810,thx!

PR #1856, opened 2026-09-16
Mock ec.oci.parsed_blob

joejstuart:
We recently switched to using parsed_blob.which broke this test.

### Explanation
The test passed CI because the PR’s required check ran on September 4 against an older main, before the ec.oci.parsed_blob change landed. main advanced afterward, but branch protection had strict checks disabled, so the stale successful check was still accepted and the PR merged. Post-merge CI then caught the failure.


robnester-rh (approved):
LGTM

Acepresso (approved):

st3penta (approved):

